PR #12, opened 2021-07-16
Apply some bug fixes / fix translations

charleskawczynski:
- Adds type annotations to method arguments
 - properly defines and uses `argwhere`
 - Fixes a scope bug in `buoyancy`


charleskawczynski:
Amazing that this had _0_ impact on the regression tests!

PR #16, opened 2021-07-19
LES driven SCM case

yairchn:
This PR adds a general case for driving the SCM with LES data. 
The case loads relevant initial conditions, large scale forcing tendencies and radiation from LES and forces a SCM simulation.


yairchn:
7/19 - This PR is not complete yet, I am making it to use buildkite to test the exisining cases are unchanged

PR #21, opened 2021-07-19
Remove unused code and comments

charleskawczynski:


szy21:
I found some other unused codes when trying to run it. Could I just comment here? For example the [two lines](https://github.com/CliMA/TurbulenceConvection.jl/blob/6e25dd017ed029076b68c0617a99f81209a0b3ad/integration_tests/utils/Cases.jl#L345) here.

charleskawczynski:
> For example the two lines here.

Fixed.

trontrytel (approved):

PR #23, opened 2021-07-19
Replace tdma with built-in Julia implementation

charleskawczynski:
🎉

### DIFF
--- a/integration_tests/ARM_SGP.jl
+++ b/integration_tests/ARM_SGP.jl
@@ -13,13 +13,13 @@ using .NameList
 using .ParamList
 
 best_mse = OrderedDict()
-best_mse["qt_mean"] = 6.7939652328911526e+00
-best_mse["updraft_area"] = 2.6728907094911216e+02
-best_mse["updraft_w"] = 3.3246977024115265e-01
-best_mse["updraft_qt"] = 6.0684798609715308e+01
-best_mse["updraft_thetal"] = 6.9056681132480222e+01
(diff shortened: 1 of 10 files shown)


charleskawczynski:
I locally ran
```julia
function tridiag_solve(nz, b_rhs, a, b, c)
    # Note that `1:end` is zero-based indexing.
    A = LinearAlgebra.Tridiagonal(a[1:end], parent(b), c[0:(end - 1)])
    x_new = off_arr(A \ parent(b_rhs))
    x = deepcopy(b_rhs)
    tridiag_solve_old(nz, x, a, b, c)
    if !(x ≈ x_new)
        error("Not matching")
        @show a
        @show b
        @show c
        @show b_rhs
        @show x
        @show x_new
    else
    end
    return x
end


function tridiag_solve_old(nz, x, a, b, c)
    scratch = pyzeros(length(x))
    scratch[0] = c[0] / b[0]
    x[0] = x[0] / b[0]
    @inbounds for i in xrange(1, nz)
        m = 1.0 / (b[i] - a[i] * scratch[i - 1])
        scratch[i] = c[i] * m
        x[i] = (x[i] - a[i] * x[i - 1]) * m
    end
    # TODO: verify translation
    @inbounds for i in revxrange(nz - 2, 0, -1)
        x[i] = x[i] - scratch[i] * x[i + 1]
    end
    return
end
```
and found that the error is never hit, so the mse table changes are required due to machine precision errors.

haakon-e (approved):
Looks good.

PR #25, opened 2021-07-19
Use Thermodynamics.jl for reference state profiles

charleskawczynski:
This PR adds the use of Thermodynamics.jl to help construct the reference state profiles.


yairchn:
** Do not merge yet **

Examining the Artifacts, there are many in which the difference is acceptable but some that are not acceptable.
The most outstanding are:
Soares: Updraft area (note that updraft theta here is ok in lower levels even though updraft area is off which is unclear to me how).
TRMM: TKE, theta_updraft, updraft area, v_mean, (might be off due to plotting as its off also in the SCAMPy plot)

src/ReferenceState.jl
```diff
@@ -101,18 +102,17 @@ function initialize(self::ReferenceState, Gr::Grid, Stats::NetCDFIO_Stats)
...
-        temperature[k] = ret.T
-        ql[k] = ret.ql
-        qv[k] = self.qtg - (ql[k] + qi[k])
-        alpha[k] = alpha_c(p_[k], temperature[k], self.qtg, qv[k])
-        ret = eos(t_to_entropy_c, eos_first_guess_entropy, p_half_[k], self.qtg, self.sg)
-        temperature_half[k] = ret.T
-        ql_half[k] = ret.ql
-        qv_half[k] = self.qtg - (ql_half[k] + qi_half[k])
-        alpha_half[k] = alpha_c(p_half_[k], temperature_half[k], self.qtg, qv_half[k])
+        ts = TD.PhaseEquil_pTq(param_set, p_[k], T_g, q_tot_g)
```
yairchn, line 106: here we need to add a 'PhaseEquil_pSq' with 'S' as moist entropy to reproduce the scampy reference profile.

charleskawczynski: We've decided to forgo the entropy constructors in favor of the existing ones.

charleskawczynski:
I think we've worked out most of the kinks. We'll come back to the reference state to address #98 later.

PR #28, opened 2021-07-20
fix: indexing error

haakon-e:


charleskawczynski:
Good catch! I just fixed this a few minutes ago in #27!

PR #29, opened 2021-07-20
Add reference state profile plots to docs

charleskawczynski:
This will be helpful to verify that the reference state profiles don't change too dramatically in #25.


haakon-e (approved):
It seems like the bottom and left sides of the figures don't fit the documentation frame, i.e. the plot labels are slightly hidden. Otherwise, looks good.

charleskawczynski:
Yeah, let me see if I can fix that

charleskawczynski:
Thanks for taking a look!

PR #30, opened 2021-07-20
fix: write namelist and paramlist to file

haakon-e:
Store namelist and paramlist in output folder together with the NetCDF datafile.

### DIFF
--- a/src/NetCDFIO.jl
+++ b/src/NetCDFIO.jl
@@ -1,4 +1,5 @@
 using NCDatasets
+using JSON
 
 # TODO: remove `vars` hack that avoids https://github.com/Alexander-Barth/NCDatasets.jl/issues/135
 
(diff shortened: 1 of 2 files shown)


src/NetCDFIO.jl
```diff
@@ -52,10 +52,13 @@ mutable struct NetCDFIO_Stats
...
-        # TODO: this was commented:
-        # cp(joinpath(inpath, "$simname.in"), joinpath(outpath, "$simname.in"))
-        # cp(joinpath(inpath, "paramlist_$casename.in"), joinpath(outpath, "paramlist_$casename.in"))
+        # Write namefile and paramfile to output directory
+        open(joinpath(outpath, "$simname.in"), "w") do io
+            JSON.print(io, namelist, 4)
+        end
+        open(joinpath(outpath, "paramlist_$casename.in"), "w") do io
+            JSON.print(io, namelist, 4)
+        end
```
charleskawczynski, line 61: It looks like scampy used to copy the file from the folder to a new folder. Should we follow this to avoid duplicating the namelist/paramlist? They should already be generated in, for example, [default_namelist](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/integration_tests/utils/generate_namelist.jl#L102)

charleskawczynski (approved):
LGTM

PR #31, opened 2021-07-20
Cleanup

charleskawczynski:
This PR:
 - Deletes some unused and commented code
 - Moves some of the outer constructors to where the types are defined.


charleskawczynski:
MSE tables had to be udpated because we are removing the pi definition, which had less precision than Julia's builtin definition.

PR #34, opened 2021-07-21
Merge paramlist into namelist

charleskawczynski:
@yairchn suggested we merge the namelist and paramlist. @haakon-e, I believe this will affect the calibration interface.


szy21:
I think this is a good idea. I may call it "namelist" rather than "paramlist" because it's not just parameters. I don't have a strong preference though.

szy21 (approved):

yairchn (approved):
LGTM - I am joining Zhaoyi's comment that namelist is a better name for the combined file

charleskawczynski:
Fine by me, I'll do a replace-all to "namelist"

src/NetCDFIO.jl
```diff
@@ -57,7 +57,7 @@ mutable struct NetCDFIO_Stats
         open(joinpath(outpath, "$simname.in"), "w") do io
             JSON.print(io, namelist, 4)
         end
```
haakon-e, line 59:
```suggestion
```
remove this and change the comment

haakon-e (approved):
Looks good. Merge and I'll fix things on the calibration side.

PR #36, opened 2021-07-21
Add grid abstractions

charleskawczynski:
This PR adds some convenience methods for handling traversing the grid.


charleskawczynski:
This is a rather large amount of work, so I think I'd prefer to merge things incrementally.

PR #39, opened 2021-07-22
Remove unused functions

charleskawczynski:
Remove unused functions, fix some typos. I don't think these extra functions are needed, but it would be nice to get confirmation.

### DIFF
--- a/src/turbulence_functions.jl
+++ b/src/turbulence_functions.jl
@@ -220,9 +220,6 @@ function get_mixing_tau(zi, wstar)
     return zi / (wstar + 0.001)
 end
 
-
-
-
 # MO scaling of near surface tke and scalar variance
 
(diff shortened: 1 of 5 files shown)


trontrytel (approved):
LGTM. Not sure about the `entr_detr_env_moisture_deficit_b_ED_MF` not being needed at some point @yairchn ?

src/turbulence_functions.jl
```diff
@@ -1,50 +1,5 @@
 # Entrainment Rates
 
-function entr_detr_env_moisture_deficit_b_ED_MF(entr_in::entr_struct)
```
yairchn, line 3: I would like to keep this version of the entrainment closure - we might find ourselves using it

yairchn (commented):
please keep the entrainment functions that are there

PR #40, opened 2021-07-22
Try thermodynamics outside of ref state

charleskawczynski:


yairchn (commented):
The code changes look correct to me but the simulation results are not good. 
I am afraid we will need to make an offline comparison of 'eos' and 'PhaseEquil_pθq' to make sure they operate the same way.
We should also take a close look at the parameters to make sure there is no difference there.

charleskawczynski:
This is pretty stale and #196 hit a lot of replacements

PR #42, opened 2021-07-22
move max_supersaturation to namelist

szy21:
This is one step towards moving the microphysics parameters to the namelist, so that we can use the calibration pipeline for those parameters. I have confirmed the results using the rain_model = 'cutoff' are the same after this change. @trontrytel @charleskawczynski Could you please take a look and let me know whether it makes sense? Thanks!


trontrytel (approved):
LGTM!

szy21:
@charleskawczynski I wonder what the codecov failure is?

charleskawczynski:
No need to worry about codecov because we have no tests in runtests.jl. We could probably turn it off, but it’ll be useful once we start adding unit tests

charleskawczynski (approved):
LGTM

PR #43, opened 2021-07-26
Add thermo extension with constant entropy ref state

charleskawczynski:
This PR adds some pressure-entropy-total specific humidity constructors, and uses them for constructing the reference state to ensure a constant entropy reference state.


charleskawczynski:
Some notes/findings with this PR:

 - The reference states seem to match (visually indistinguishable) with all cases except for `TRMM_LBA` and `GATE_III`. Top candidates for why this happens: 1) different latent heats (easy to check), 2) scampy bug (see below) regarding liquid condensate at freezing temperatures
 - The newly added method, `air_density_equil`, may be useful to Thermodynamics.jl in its own right.
 - All cases are converging, but `TRMM_LBA` results (and potentially `GATE_III`) have worsened. There were deviations in `TRMM_LBA` after the initial direct translation to Julia, so there are likely bugs in TC.jl's `TRMM_LBA`, and perhaps these deviations are amplified by this PRs changes.
 - This begs the question, should we be running `GATE_III`? Perhaps as a weekly cron job, or a QA run? @trontrytel, is this something that you could set up?

This needs to be confirmed, but I believe I've found a bug in SCAMPy: liquid condensate (`q_liq`)-- not ice condensate--develops even when `eos` returns freezing temperatures (T < 273.15 K). I believe this case occurs in `TRMM_LBA`, will hopefully confirm soon.

charleskawczynski:
It seems that the thermo functions are not yet self-consistent (by about an entropy value of ~13). Perhaps this is why there's still a discrepency? This may be easier to look into in Thermodynamics.jl.

src/ReferenceState.jl
```diff
@@ -44,17 +44,18 @@ specific initialization function defined in Initialization.pyx
...
-        ret = eos(t_to_entropy_c, eos_first_guess_entropy, exp(p), self.qtg, self.sg)
-        q_i = 0.0
-        q_l = ret.ql
-        T = ret.T
-        return -g / (Rd * T * (1.0 - self.qtg + eps_vi * (self.qtg - q_l - q_i)))
+    function rhs(logp, u, z)
+        ts = PhaseEquil_psq(param_set, exp(logp), self.sg, self.qtg)
+        T = TD.air_temperature(ts)
+        R_m = TD.gas_constant_air(ts)
+        return -FT(grav(param_set)) / (T * R_m)
```
tapios, line 58: This really is `-1/H`, where `H = R * T/g` is the scale height. It would be clearer to return the scale height `H` and call it that.

charleskawczynski:
Updates after #47 and #51: Most results appear fine. Rico and ARM have slightly noticeable differences, but qualitatively and quantitatively satisfactory. TRMM_LBA is the one outlier, where updraft w and tke become undesirably large in the upper atmos.

charleskawczynski:
I think we’re forgoing this PR in favor of #25.

PR #48, opened 2021-07-27
Add field abstraction

charleskawczynski:
This PR adds a field constructor, just so that we can get a handle on the number of places where dofs are specified for center and face fields.

This is needed for #50.


charleskawczynski:
Would be nice for someone to review that center and face fields are properly assigned to each variable, since it's not clear to me based on the number of grid points (most of the time it's the same, but sometimes, e.g., with `f_rad`, it's different), cc @trontrytel, @yairchn, @szy21

charleskawczynski:
I guess any individual field can be easily fixed if something is wrong. This is still a no-op PR.

PR #51, opened 2021-07-28
Fix inversion height bug, read Ri bulk crit from namelist, add Nieuwstadt case

charleskawczynski:
I believe this fixes the Nieuwstadt case. This may also fix / improve other cases. cc @yairchn. Looking forward to re-try #25, #40, and #43.


charleskawczynski:
Partial fix to #3

PR #53, opened 2021-07-29
Fix translation bug in ref state bcs

charleskawczynski:
This is likely not what we want in the end, but for early-on comparisons, and perhaps if there are assumptions elsewhere about how the ref state bcs are applied, this PR fixes a translation bug in the ref state bcs: in SCAMPy, the `p_half` field has two missing elements at the start and end of the array, and _both_ points are _mirrored_ in _log-space_ about the boundary. This enters the computations [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/Turbulence_PrognosticTKE.jl#L511).

### DIFF
--- a/integration_tests/TRMM_LBA.jl
+++ b/integration_tests/TRMM_LBA.jl
@@ -11,14 +11,14 @@ include(joinpath("utils", "compute_mse.jl"))
 using .NameList
 
 best_mse = OrderedDict()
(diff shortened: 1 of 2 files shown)


charleskawczynski:
With these changes, ~120 fields in TRMM_LBA match scampy to ~1e-9 (max mse) at the 10th timestep.

PR #54, opened 2021-07-29
Remove nc file if it exists before creating it

charleskawczynski:
Sometimes julia complains that I don't have permission to create the nc file (sometimes when simulations crash). This should smooth the workflow a bit.

### DIFF
--- a/src/NetCDFIO.jl
+++ b/src/NetCDFIO.jl
@@ -62,6 +62,9 @@ mutable struct NetCDFIO_Stats
         cinterior = Gr.cinterior
         finterior = Gr.finterior
 
+        # Remove the NC file if it exists, in case it accidentally wasn't closed
+        isfile(path_plus_file) && rm(path_plus_file; force = true)
+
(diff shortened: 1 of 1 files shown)


haakon-e:
Looks good to me. Is it possible to write the `.nc` files in a context to make sure they're always closed in case anything happens? (I don't like the idea of accidentally keeping files open anyways)

haakon-e (approved):

charleskawczynski:
> Looks good to me. Is it possible to write the .nc files in a context to make sure they're always closed in case anything happens? (I don't like the idea of accidentally keeping files open anyways)

Yes, it is possible and I agree, but we [can't easily](https://github.com/Alexander-Barth/NCDatasets.jl/issues/135) open-close the file on the fly, so this open-close has to be done at a high level for now. We could wrap the whole solve in a `do`-block that opens the nc file, but that might be a bit of work (I haven't tried though).

Let's merge this for now, I'll open an issue on improving things.

PR #56, opened 2021-07-30
Inspect artifact file

charleskawczynski:


charleskawczynski:
I can confirm that, running Bomex on the `main` branch on my mac results in no artifact file change, however, running on my windows laptop _does_ result in an artifact file change. I think this may be enough motivation to git ignore the artifact files, to smoothen developer's workflows.

PR #60, opened 2021-07-30
Apply theta-l bcs where it seems to be missing

charleskawczynski:
I'm hoping that this will fix the theta-L plots in our comparisons.

### DIFF
--- a/integration_tests/TRMM_LBA.jl
+++ b/integration_tests/TRMM_LBA.jl
@@ -11,14 +11,14 @@ include(joinpath("utils", "compute_mse.jl"))
 using .NameList
 
 best_mse = OrderedDict()
-best_mse["qt_mean"] = 3.9046069426789907e+00
-best_mse["updraft_area"] = 2.8342045114790926e+04
+best_mse["qt_mean"] = 3.9046069426558234e+00
+best_mse["updraft_area"] = 2.8331099674735910e+04
 best_mse["updraft_w"] = 6.5629758352987142e+02
-best_mse["updraft_qt"] = 2.5132229019169145e+01
-best_mse["updraft_thetal"] = 1.1134328331311413e+02
-best_mse["v_mean"] = 2.9404204554100676e+02
-best_mse["u_mean"] = 1.6903145439466905e+03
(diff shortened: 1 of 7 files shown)


haakon-e (approved):
The plots look nice! 

Do you fix the same random seed for each simulation when test plots are created? That may be the cause of some small discrepancies too.

charleskawczynski:
> The plots look nice! ... Do you fix the same random seed for each simulation when test plots are created? That may be the cause of some small discrepancies too.

Thanks :slightly_smiling_face:. There's currently no random seed, but we've [suppressed random numbers](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/utility_functions.jl#L5-L11) for now (and have TODOs to add them back in). I believe the small differences are due to the BCs being applied.

Hmm, I thought that some of the plots would be different. E.g., Bomex theta-L contours.

haakon-e:
> > The plots look nice! ... Do you fix the same random seed for each simulation when test plots are created? That may be the cause of some small discrepancies too.
> 
> Thanks 🙂. There's currently no random seed, but we've [suppressed random numbers](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/utility_functions.jl#L5-L11) for now (and have TODOs to add them back in). I believe the small differences are due to the BCs being applied.
> 
> Hmm, I thought that some of the plots would be different. E.g., Bomex theta-L contours.

I only looked at TRMM, which failed.. But I can't see any difference in theta_l in Bomex. Are you saying that you'd expect a difference due to this change?

charleskawczynski:
> I only looked at TRMM, which had the failed tests.. But I can't see any difference in theta_l in Bomex. Are you saying that you'd expect a difference due to this change?

Yeah, the contour plots look like solid planes. I just found out why, I think, though. The ICs for THL are not defined, so it exports zeros at t=0, which kind of ruins the contour plots. Trying it out now.

charleskawczynski:
Yeah, that was it. Seems to look okay now. I had to add a if-branch in `compute_mse` for this anyway since the Artifacts still have uninitialized temperatures at t=0.

charleskawczynski:
Forgot to update the tables.. bors r+

PR #62, opened 2021-07-30
Move clima_1m parameters to the namelist

szy21:
This PR moves all clima_1m parameters to the namelist except for rho-cloud_liq, nu_air, K_therm, and D_vapor. I assume these parameters won't change but can move them as well. The results for the Rico test case (which uses clima_1m) are identical before and after the change.


charleskawczynski (commented):
I think I'd prefer to pass `param_set` through as an argument. This will allow us to call, e.g., `C_drag(::AbstractEarthParameterSet)` and friends, and keep the argument list a bit more compact. This also means that we don't need to store the parameters in `RainVariables`, since we can just call the methods with the param set when needed. @trontrytel, what do you think?

szy21:
I agree. The only reason I did this is to use the calibration pipeline (soonish). Could we still do that with the `param_set`?

charleskawczynski:
> I agree. The only reason I did this is to use the calibration pipeline (soonish). Could we still do that with the `param_set`?

Yep! You just need to grab the parameters from the namelist [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/integration_tests/utils/parameter_set.jl#L13), and overload the methods [like this](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/integration_tests/utils/parameter_set.jl#L9).

Then, any calls to, e.g., `MSLP(param_set)` will return the value that is grabbed from the namelist.

szy21:
Great, let's see what Anna thinks, and I'm happy to change anything!

charleskawczynski:
I just spoke with @szy21, let's just merge this and we can move the parameters to clima parameters later.

charleskawczynski (approved):

PR #72, opened 2021-08-03
second round of cleanups

yairchn:
This second cleanup PR is a second step for simplification of the code. Here I am removing some straightforward if commands and arg that are not used in the current step (old entrainment functions and such) as well as the similarity diffusivity option in Turbulence. In a following PR after similarity diffusivity is removed we should merge Turbulence and TurbulencePrognosticTKE into a single file.


integration_tests/utils/Artifacts.toml
```diff
@@ -2,4 +2,4 @@
 git-tree-sha1 = "4a54eb3dfa48035d7a4318005963408a4e94c6d5"
 
 [SCAMPy_output]
-git-tree-sha1 = "fb9af006a3ca35cb713caa00ef06db4047bede07"
+git-tree-sha1 = "6b82611f969d3cf99102baa7d898e3c759cf014c"
```
charleskawczynski, line 5: Let's undo this. I'll see if we can git ignore this file..

charleskawczynski (approved):
Looks amazing.

PR #74, opened 2021-08-03
Use bcs and avoid accessing ghost points

charleskawczynski:
This should help us move towards removing ghost points.


charleskawczynski:
So, this is technically behavior changing because, previously, `∇U` reached to ghost points, which did not necessarily enforce 0 Neumann BCs. But I think that this is close enough--we could also change it to `Extrapolate` down the road if needed.

PR #76, opened 2021-08-03
Fix and improve dycoms radiation

charleskawczynski:
This PR:
 - Makes `f_rad` a `face_field` (with the now correct size)
 - Removes a call to `revxrange`, which we need to phase out
 - Uses OrdinaryDiffEq to integrate the radiation profile, instead of doing it ourselves
 - Fixes the upper radiation BC, which didn't seem to match the literature:

f_rad


yairchn (approved):
LGTM

trontrytel (approved):

tapios (approved):
LGTM

haakon-e (approved):

PR #77, opened 2021-08-04
Ignore artifact file to avoid git conflicts

charleskawczynski:
This artifact file has proven to be problematic for people's workflow. I think everything should work fine if we git-ignore it.


yairchn (approved):
LGTM

haakon-e (approved):

PR #78, opened 2021-08-04
Use more grid and field abstractions

charleskawczynski:
This PR removes more calls to python primitives (xrange), and uses the field constructors.


charleskawczynski:
Wow, worked on the first shot! 👊🏻

yairchn (approved):

PR #81, opened 2021-08-04
Remove `revxrange`

charleskawczynski:
I think that this is done correctly, note that we're not traversing the `term_vel[k] = terminal_velocity(Rain.C_drag, Rain.MP_n_0, QR.values[k], self.Ref.rho0_half[k])` in reverse order, since I don't think that should have any impact on the result.

### DIFF
--- a/src/EDMF_Rain.jl
+++ b/src/EDMF_Rain.jl
@@ -111,28 +111,27 @@ function solve_rain_fall(
     QR::RainVariable,
     RainArea::RainVariable,
 )
-    gw = self.Gr.gw
-    nzg = self.Gr.nzg
-    dz = self.Gr.dz
+    grid = get_grid(GMV)
+    dz = grid.dz
     dt_model = TS.dt
     CFL_limit = 0.5
 
-    term_vel = center_field(self.Gr)
-    term_vel_new = center_field(self.Gr)
+    term_vel = center_field(grid)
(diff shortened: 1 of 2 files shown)


trontrytel:
> I think that this is done correctly, note that we're not traversing the `term_vel[k] = terminal_velocity(Rain.C_drag, Rain.MP_n_0, QR.values[k], self.Ref.rho0_half[k])` in reverse order, since I don't think that should have any impact on the result.

Not sure. The idea was to integrate the rain from top down because it is falling down and we are using upwind to do it

charleskawczynski:
> Not sure. The idea was to integrate the rain from top down because it is falling down and we are using upwind to do it

We're still looping in reverse order in the loop below, I think the loop I changed the order on simply populates `term_vel`--it doesn't look like that loop depends on the order (unless it's for a simplified case, in which case I'm more than happy to keep it in reverse order).

trontrytel (approved):

trontrytel:
> > Not sure. The idea was to integrate the rain from top down because it is falling down and we are using upwind to do it
> 
> We're still looping in reverse order in the loop below, I think the loop I changed the order on simply populates `term_vel`--it doesn't look like that loop depends on the order (unless it's for a simplified case, in which case I'm more than happy to keep it in reverse order).

You are right. I think this should be ok

PR #82, opened 2021-08-05
fuse updraft closures

yairchn:
This PR merges four updraft closures into a single function as part of the fuse operations task. 
It is a first step in replacing these functions with the point wise functions coded in ClimateMachine.


charleskawczynski (approved):
Looks great!

PR #90, opened 2021-08-06
feat: lognormal stochastic closure

haakon-e:
implement framework for stochastic closures extending the entrainment and detrainment parameters,
and implement the lognormal closure.


integration_tests/utils/parameter_set.jl
```diff
@@ -10,12 +10,32 @@ end
 CLIMAParameters.Planet.MSLP(ps::EarthParameterSet) = ps.nt.MSLP
 CLIMAParameters.Atmos.Microphysics.τ_cond_evap(ps::EarthParameterSet) = ps.nt.τ_cond_evap
 
+# stochastic
+CLIMAParameters.Atmos.EDMF.stoch_closure(ps::EarthParameterSet) = ps.nt.stoch_closure
+CLIMAParameters.Atmos.EDMF.stoch_ε_lognormal_var(ps::EarthParameterSet) = ps.nt.stoch_ε_lognormal_var
+CLIMAParameters.Atmos.EDMF.stoch_δ_lognormal_var(ps::EarthParameterSet) = ps.nt.stoch_δ_lognormal_var
```
charleskawczynski, line 16: I think these methods need to be defined in CLIMAParameters for this to work.

src/stochastic_closures.jl
```diff
@@ -0,0 +1,44 @@
+
```
charleskawczynski, line 1:
```suggestion
import Distributions
```

src/stochastic_closures.jl
```diff
@@ -0,0 +1,44 @@
...
+function lognormal_closure(param_set::APS, t::EntDet) where EntDet
+    lognormal_var = stoch_lognormal_var(param_set, t)
+    return lognormal_sampler(1.0, lognormal_var)
+end
+
+function lognormal_sampler(m, var)
+    μ = log(m^2 / √(m^2 + var))
+    σ = √(log(1 + var / m^2))
+    return rand(LogNormal(μ, σ))
```
charleskawczynski, line 31:
```suggestion
    return rand(Distributions.LogNormal(μ, σ))
```

src/stochastic_closures.jl
```diff
@@ -0,0 +1,44 @@
...
+#
+## Stochastic differential equations
+#
+function sde(a, b, u0, dt)
+    f(u,p,t) = a*u
+    g(u,p,t) = b*u
+    sub_dt = dt // 8
+    tspan = (0.0,dt)
+    prob = SDEProblem(f,g,u0,tspan)
+    sol = solve(prob,EM(),dt=sub_dt)
```
charleskawczynski, line 43: What's `EM` here?

integration_tests/utils/parameter_set.jl
haakon-e: hmm. it seems like quite a bit of hassle to update CLIMAParameters at this stage of testing. I would rather want to create some definitions "locally" in TC.jl, at least for now.

src/stochastic_closures.jl
haakon-e: commented out for now.
(EM is a time-stepper for SDEs in DiffEq)

integration_tests/utils/parameter_set.jl
charleskawczynski: Yeah, let's just define these methods in a new file, `clima_params.jl` in `src/`, and add a TODO comment to move to CLIMAParameters.

src/stochastic_closures.jl
```diff
@@ -0,0 +1,47 @@
+import Distributions
+
+include("clima_params.jl")
```
charleskawczynski, line 3: Since these may not be the only clima parameters, let's `include` this in `src/TurbulenceConvection.jl`.

src/clima_params.jl
```diff
@@ -0,0 +1,5 @@
+
+# stochastic
+stoch_closure(ps::APS) = ps.nt.stoch_closure
+stoch_ε_lognormal_var(ps::APS) = ps.nt.stoch_ε_lognormal_var
+stoch_δ_lognormal_var(ps::APS) = ps.nt.stoch_δ_lognormal_var
```
charleskawczynski, line 5: Let's put this in a module, `ClimaParams`, and then, in `src/TurbulenceConvection.jl`, we can do

```julia
include("ClimaParams.jl")
import ClimaParams
const ICP = ClimaParams # maybe "i" for internal?

# later use, e.g., ICP. stoch_closure(param_set)
```
That way, we can easily grep `ICP.` to find all of the parameters that need to eventually move to CLIMAParameters.jl.

charleskawczynski (approved):
This LGTM--let's squash first, though. Did we want to uncomment the `sde(a, b, u0, dt)` function?

haakon-e:
I think I might wait to merge until #100 is merged, to see if that impacts this PR.

> This LGTM--let's squash first, though. Did we want to uncomment the `sde(a, b, u0, dt)` function?

I'll squash and work on implementing `sde`. For `sde` I need to retain some form of memory with regards to the previous value of my parameters as well as fetch the current time step.

haakon-e:
could you have another look @charleskawczynski ? Then we can merge.

src/stochastic_closures.jl
```diff
@@ -0,0 +1,74 @@
...
+    return u
+end
+
+""" Solve the Ornstein-Uhlenbeck process `du = f(u,p,t)⋅dt + g(u,p,t)⋅dW` numerically 
+
+This formulation solves the Cox-Ingersoll-Ross model, a variant of the Vasicek model
+which ensures that `u` remains positive so long as `a,b>0`.
+
+ref: https://en.wikipedia.org/wiki/Cox–Ingersoll–Ross_model
+"""
```
charleskawczynski, line 64:
```suggestion
"""
    Solve the Ornstein-Uhlenbeck process `du = f(u,p,t)⋅dt + g(u,p,t)⋅dW` numerically 

This formulation solves the Cox-Ingersoll-Ross model, a variant of the Vasicek model
which ensures that `u` remains positive so long as `a,b>0`.

ref: https://en.wikipedia.org/wiki/Cox–Ingersoll–Ross_model
"""
```

src/stochastic_closures.jl
```diff
@@ -0,0 +1,74 @@
...
+    closure = ICP.stoch_closure(param_set)
+    u = if closure == "none"
+        1.0  # By default, unity scaling (i.e. entr/detr are unaffected)
+    elseif closure == "lognormal"
+        lognormal_closure(param_set, term)
+    elseif closure == "sde"
+        sde_closure(param_set, sde_model, term)
+    else
+        error("Unknown stochastic closure $closure.")
+    end
```
charleskawczynski, line 18: Why does `stoch_closure` need to be a clima parameter? I think I would prefer adding a type to `sde_struct`, e.g.:

```julia
struct NoneClosureType end
struct LogNormalClosureType end
struct SDEClosureType end

Base.@kwdef mutable struct sde_struct{T}
    u0::Float64
    dt::Float64
end

# where sde_struct is initialized, for example use:
sde_model = sde_struct{SDEClosureType}(u0 = 1, dt = dt)
# The type can be determined by a namelist parameter and put into `sde_struct{T}`
```

This way the compute kernel doesn't need to branch off at runtime:

```julia
stochastic_closure(::APS, ::sde_struct{NonClosureType}, ::AbstractEntDet) = 1
function stochastic_closure(param_set::APS, ::sde_struct{LogNormalClosureType}, term::AbstractEntDet)
    u = lognormal_closure(param_set, term)
    if u < 0
        @warn("Negative stochastic parameter ⟹ negative $(typeof(term)). Setting to 0")
        u = 0
    end
    return u
end
function stochastic_closure(param_set::APS, sde_model::sde_struct{SDEClosureType}, term::AbstractEntDet)
    u = sde_closure(param_set, sde_model, term)
    if u < 0
        @warn("Negative stochastic parameter ⟹ negative $(typeof(term)). Setting to 0")
        u = 0
    end
    return u
end
```

All that said, if this is just experimental, then I think this is all fine and we can refactor later.

charleskawczynski (approved):
Overall looks good to me, let's just for sure fix the one doc string

PR #93, opened 2021-08-08
Consolidate turbulent entrainment

yairchn:
This PR consolidate turbulent entrainment to a single term that is effectively added to the dynamic entrainment. It is a step on the way to merging the two into a single closure file. 

** this branch has behavioral changes as it computes turbulent entrainment only at half levels and interpolates it to full levels **


yairchn:
looking through all the changes in the Artifacts I do not find any meaningful change from main.

charleskawczynski (approved):
Code changes and results both look fine to me 🚀

yairchn:
Bors r+

PR #99, opened 2021-08-09
Make les comparison optional, + more cases plots

charleskawczynski:
Big improvement on #3, here. This PR makes LES data optional, so this PR adds regression tests for
 - DryBubble
 - life_cycle_Tan2018
 - SP

And compares against SCAMPy


charleskawczynski:
Note that I had to change `Hvar_mean` in TRMM_LBA because, while it showed up in the MSE table, we didn't have a regression test on it.

PR #100, opened 2021-08-09
Move entr and detr to separate file 

yairchn:
This PR moves the computation of entrainment and detrainment to a separate file  - while breaking of the non dimensional functions so that they could replaced by NN or any time of ML in the future.

Both this PR and the the recent changes to pressure should have all parameters read of CliMA parameters in the future


src/closures/entr_detr.jl
```diff
@@ -0,0 +1,101 @@
...
+    ql_up,
+    ql_env,
+    RH_upd,
+    RH_env,
+    M,
+    dMdz,
+    tke,
+    N_up,
+    ρ,
+)
```
charleskawczynski, line 70: I think we need to reduce the number of vars passed in. A big risk (because it can be a really difficult bug to catch) of this is mixing up the order of the inputs.

src/types.jl
```diff
@@ -813,24 +804,24 @@ mutable struct EDMF_PrognosticTKE{PS, A1, A2}
...
+        #     entr_detr_env_moisture_deficit
+        # elseif entr_detr_fp_str == "moisture_deficit_b_ED_MF"
+        #     entr_detr_env_moisture_deficit_b_ED_MF
+        # elseif entr_detr_fp_str == "moisture_deficit_div"
+        #     entr_detr_env_moisture_deficit_div
+        # elseif entr_detr_fp_str == "none"
+        #     entr_detr_none
+        # else
+        #     error("Turbulence--EDMF_PrognosticTKE: Entrainment rate namelist option is not recognized")
+        # end
```
charleskawczynski, line 824: Let's delete this instead of commenting

charleskawczynski (commented):
After deleting the block of commented code, I think this looks ready to go.

PR #101, opened 2021-08-09
Replace more thermodynamics

charleskawczynski:
This seems to ruin the results of a bunch of cases.


charleskawczynski:
This is pretty stale and #196 hit a lot of replacements

PR #107, opened 2021-08-10
make use of the correct qt in extra buoyancy

yairchn:
Behavioral changes but plots look fine

### DIFF
--- a/src/EDMF_Updrafts.jl
+++ b/src/EDMF_Updrafts.jl
@@ -393,11 +393,10 @@ function buoyancy(
                     UpdVar.B.values[i, k] = buoyancy_c(self.Ref.rho0_half[k], rho)
                     UpdVar.RH.values[i, k] = relative_humidity_c(self.Ref.p0_half[k], qt, qt - qv, 0.0, t)
                 elseif UpdVar.Area.values[i, k - 1] > 0.0 && k > kc_surf
-                    qt = UpdVar.QT.values[i,k-1]
-                    h  = UpdVar.H.values[i,k-1]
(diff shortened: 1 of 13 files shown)


charleskawczynski:
> Behavioral changes but plots look fine

They don't look fine to me! Bomex updraft w looks like it's zero!

src/EDMF_Updrafts.jl
```diff
@@ -393,11 +393,10 @@ function buoyancy(
...
                     UpdVar.RH.values[i, k] = relative_humidity_c(self.Ref.p0_half[k], qt, qt - qv, 0.0, t)
                 elseif UpdVar.Area.values[i, k - 1] > 0.0 && k > kc_surf
-                    qt = UpdVar.QT.values[i,k-1]
-                    h  = UpdVar.H.values[i,k-1]
+                    qt = UpdVar.QT.values[i, k - 1]
+                    h = UpdVar.H.values[i, k - 1]
                     sa = eos(self.Ref.p0_half[k], qt, h)
-                    qt -= sa.ql
-                    qv = qt
+                    qv = qt = sa.ql
```
charleskawczynski, line 399:
```suggestion
                    qv = qt - sa.ql
```

trontrytel (approved):
The only difference I see is in the dry bubble case. But I don't know if it was always different between scampy and TC.jl or if it's caused by this change?

charleskawczynski (approved):
Whatever I saw before seems to be fixed 🎉

trontrytel:
> Whatever I saw before seems to be fixed 🎉

What has been seen, cannot be unseen...

yairchn:
Bors r+

PR #109, opened 2021-08-11
Add mse wrapper to make optional files flexible

charleskawczynski:
This should allow us to optionally, and safely, use in any of the nc files (les, scampy, tc, tc-main) in a more flexible way.

### DIFF
--- a/integration_tests/utils/compute_mse.jl
+++ b/integration_tests/utils/compute_mse.jl
@@ -72,18 +72,50 @@ function get_time(ds, var)
     error("No key for $var found in the nc file.")
 end
 
-function compute_mse(
-    experiment,
+append_dict(::Nothing, sym::Symbol, dict = Dict()) = dict
+function append_dict(filename::AbstractString, sym::Symbol, dict = Dict())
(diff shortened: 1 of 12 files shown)


integration_tests/utils/compute_mse.jl
```diff
@@ -72,18 +72,50 @@ function get_time(ds, var)
...
+    ds_dict = append_dict(ds_tc_main_filename, :ds_tc_main, ds_dict)
+
+    try
+        computed_mse = compute_mse(all_args...; ds_dict, kwargs...)
+    finally
+        for ds in values(ds_dict)
+            close(ds)
+        end
+    end
+end
```
charleskawczynski, line 110: @jakebolewski, does this look like a _safe_ way to open/close multiple (where each file is optional)? When looking at the `Dataset(::Function)` method (for the `do`-syntax), it uses the `try-finally` like this.

charleskawczynski: Going to merge for now since I think it's doing what we want

PR #111, opened 2021-08-11
unify timesteps in the model

yairchn:
This PR unifies the time steps in the model across the updrafts and the grid mean the environment and the microphysics. The tilmestep is set to 3 sec which works. Next step would be to add a CFL condition that combines updraft velocity and rainfall velocity to get the optimal tilmestep. 

** behavioral changes !! ** 

take a close look

### DIFF
--- a/integration_tests/GABLS.jl
+++ b/integration_tests/GABLS.jl
@@ -12,14 +12,14 @@ using .NameList
 
 best_mse = OrderedDict()
 
-best_mse["updraft_thetal"] = 5.0248694008424222e+00
-best_mse["v_mean"] = 4.4630163253493214e+00
-best_mse["u_mean"] = 9.6412500983148846e+00
-best_mse["tke_mean"] = 2.4683562232142533e+00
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Can we update the description? #115 shrunk the timestep, and this is just removing substepping, correct? Also, we should rebase because I just fixed a bug in #127--we previously were not comparing with tc main (we are now, though), we were stuck comparing against an older commit.

charleskawczynski (approved):
The results seem virtually unchanged, so LGTM. @trontrytel, I'm not sure that the rain part is exercised yet, do these code changes look okay?

trontrytel:
LGTM too. Lets merge it. Precipitation was never tested and fully integrated into scampy. So whatever this PR changes, I will have to deal with after I'm done switching to TD.jl

charleskawczynski:
@yairchn let’s rebase/squash first, then merge

yairchn:
Bors r+

PR #116, opened 2021-08-12
Group plots

charleskawczynski:
This should help in reducing the number of clicks required to review all the results.

I think one more thing we could do is separate mse tables from what is plotted, as well as flexibility with, for example, plotting contours but not profiles.

This PR also fixed a bug in the variable map (`updraft_qr`)


trontrytel:
This looks great for the profiles. So nice to have everything in one place and to have the TC.jl main. Thank you so much for doing this!

Good luck with the contours!

integration_tests/utils/compute_mse.jl
```diff
@@ -206,7 +218,7 @@ function compute_mse(experiment, best_mse, plot_dir; ds_dict, plot_comparison =
         data_tcm_arr = get_data(ds_tc_main, tc_var)'
         data_tcc_arr = get_data(ds_tc, tc_var)'
         data_scm_arr = get_data(ds_scampy, scm_var)'
-        @info "     Data sizes (les,scm,tcc,tcm): $(size(data_les_arr)), $(size(data_scm_arr)), $(size(data_tcc_arr)), $(size(data_tcm_arr))"
+        # @info "     Data sizes (les,scm,tcc,tcm): $(size(data_les_arr)), $(size(data_scm_arr)), $(size(data_tcc_arr)), $(size(data_tcm_arr))"
```
haakon-e, line 221: this should either be not commented out or removed?

haakon-e (approved):
looks good. exactly where can I see all plots now?

integration_tests/utils/compute_mse.jl
charleskawczynski: I commented it because I think I was using it for debugging, I guess we could change it to `@debug ...`, but I've not been good with using the logger in TC.jl.

charleskawczynski:
> looks good. exactly where can I see all plots now?

 - Click on `details` for the buildkite github action status (`buildkite/turbulenceconvection-ci`)
 - Click on a case (e.g., Bomex)
 - Click on Artifacts

There should be two figs per case: `contours.png` and `profiles.png`

integration_tests/utils/compute_mse.jl
haakon-e: Okay, perhaps we could add logging to a long-term todo-list for TC.jl.

haakon-e:
It seems like the contours needs to be reviewed.
image

Profiles looks reasonably good, but wouldn't hurt to have them a bit bigger, and also extend the figure boundary so all text is visible.
image

charleskawczynski:
Yeah, we definitely need some tweaking here

charleskawczynski:
Ok, we're now filtering out `["thetal_mean", "updraft_thetal", "u_mean", "v_mean", "qt_mean", "updraft_qt", "temperature_mean"]`, I think the plots look reasonable.

jakebolewski:
It might also be useful to annotate the build with these plots:  https://buildkite.com/docs/agent/v3/cli-annotate#embedding-and-linking-artifacts-in-annotations

trontrytel (approved):
Thank you!

charleskawczynski:
> It might also be useful to annotate the build with these plots: https://buildkite.com/docs/agent/v3/cli-annotate#embedding-and-linking-artifacts-in-annotations

This PR is pretty big, but I'll take a look at this in a subsequent one.

I think this is as good as I can get it on the first pass. There's still room for improvement on the profiles.

charleskawczynski:
Actually, I'd like to see the remote results before merging.

PR #117, opened 2021-08-13
Make sure that qt in updraft is non negative 

yairchn:
This PR makes the EDMF more resilient to different parameter values by preventing negative qt in the updraft.


charleskawczynski (approved):
LGTM

yairchn:
Bors r+

PR #122, opened 2021-08-13
Reorganize scalar equations to look normal

yairchn:


charleskawczynski (approved):
Looks fantastic. We'll need to update the tables after 115 merges, but these changes look great.

PR #123, opened 2021-08-13
organize scalar equations

yairchn:


charleskawczynski (approved):
LGTM so long the results look good!

haakon-e:
Should TC main and this PR have exactly the same results?
It seems like some profiles are different.

yairchn:
Bors r+

PR #126, opened 2021-08-14
Squeeze out space in profile plots

### LINKED ISSUE #124: Improve profile plots
After #116, the profiles plots have relatively large margins, and all the xlabels/ylabels are repeated for each plot, which takes up room that could otherwise make the plots themselves bigger.

charleskawczynski:
This PR updates the layout of the profile plots to squeeze out the lost white space. I also changed the layout to be 5 columns (it was previously 4). We can change it back if people prefer 4, most of the gained space was in the horizontal direction, so I think this actually preserves the look of the figures over sticking with 4, but I have no strong feelings about this.

Closes #124.

### DIFF
--- a/integration_tests/utils/compute_mse.jl
+++ b/integration_tests/utils/compute_mse.jl
@@ -277,20 +277,20 @@ function compute_mse(
         if plot_comparison
             p = Plots.plot()
             if have_pycles_ds
-                Plots.plot!(data_les_cont_mapped, z_tcc ./ 10^3, xlabel = tc_var, ylabel = "z [km]", label = "PyCLES")
+                Plots.plot!(data_les_cont_mapped, z_tcc ./ 10^3, title = tc_var, ylabel = "z [km]", label = "PyCLES")
             end
-            Plots.plot!(data_scm_cont_mapped, z_tcc ./ 10^3, xlabel = tc_var, ylabel = "z [km]", label = "SCAMPy")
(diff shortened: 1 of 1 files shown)


haakon-e:
Could you add a top-level title which states which simulation case the subplots are showing? e.g. "Bomex contours", etc.

Also, I think we should consider making all plots larger (extend width & height). If I zoom in to study details of one profile, it becomes apparent that the resolution could be improved.
image
In safari, this also seems to be maximum zoom level.
I realize you may have chosen this size to fit all subplots in a typical browser window, but I'm not sure if this is strictly necessary.

haakon-e (approved):
Other than that, LGTM!

charleskawczynski:
Ahhh, I see what you mean, we’ll need to increase the size and font size, since increasing the size has the effect of making the font look smaller. I’ll try this out soon

charleskawczynski:
Alright, updated! Would be nice to hear feedback on these changes when people get a chance. I think it's looking near complete 🚀

haakon-e:
Seems like you need to pass through the case name to the plotting functions.

charleskawczynski:
> Seems like you need to pass through the case name to the plotting functions.

Whoops, updated, hopefully this works.

charleskawczynski:
Going to merge this for now. I think we can do tweaking beyond this if people wish.

PR #129, opened 2021-08-16
Merge mse tables->1 file, print mse table post run

charleskawczynski:
This PR should be a nice quality of life improvement for updating the MSE tables.

Previously, we had to click into each tab (for each case) and copy the table to the corresponding file

With this PR: a new job prints the entire mse table file contents after all cases are complete--so we only need to copy these file contents into the single mse tables file.


yairchn (approved):
L-GREAT-TM

PR #131, opened 2021-08-16
add Ri and Pr functions

yairchn:
This PR adds standalone gradient Richardson number and turbulent Prandtl number functions that are called from compute_mixing_length. It is the first step to organizing mixing length function.

### DIFF
--- a/src/turbulence_functions.jl
+++ b/src/turbulence_functions.jl
@@ -101,3 +101,17 @@ function set_cloudbase_flag(ql, current_flag)
     end
     return new_flag
 end
+
+function gradient_Richardson_number(∂b∂θ_l, ∂b∂q_tot, Shear², ϵ)
+    return min(∂b∂θ_l / max(Shear², ϵ) + ∂b∂q_tot / max(Shear², ϵ), 0.25)
+end
+
+function turbulent_Prandtl_number(obukhov_length, ∇Ri, Pr, ω_pr)
+    if obukhov_length > 0.0 && ∇Ri > 0.0 #stable
+        # CSB (Dan Li, 2019), with Pr_neutral=0.74 and w1=40.0/13.0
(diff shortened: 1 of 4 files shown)


src/turbulence_functions.jl
```diff
@@ -100,3 +100,17 @@ function set_cloudbase_flag(ql, current_flag)
     end
     return new_flag
 end
+
+function gradient_Richardson_number(∂b∂θ_l, ∂b∂q_tot, Shear², ϵ)
+    return min(∂b∂θ_l / max(Shear², ϵ) + ∂b∂q_tot / max(Shear², ϵ), 0.25)
```
charleskawczynski, line 105: Is the 0.25 empirical?

src/turbulence_functions.jl
```diff
@@ -100,3 +100,17 @@ function set_cloudbase_flag(ql, current_flag)
...
+end
+
+function turbulent_Prandtl_number(obukhov_length, ∇Ri, Pr, ω₁)
+    if obukhov_length > 0.0 && ∇Ri > 0.0 #stable
+        # CSB (Dan Li, 2019), with Pr_neutral=0.74 and w1=40.0/13.0
+        prandtl_nvec = Pr * (2.0 * ∇Ri / (1.0 + ω₁ * ∇Ri - sqrt((1.0 + ω₁ * ∇Ri)^2.0 - 4.0 * ∇Ri)))
```
charleskawczynski, line 111: This is fine for now, but for all of this to be type-stable we'll want to write it as
```julia
prandtl_nvec = Pr * (2 * ∇Ri / (1 + ω₁ * ∇Ri - sqrt((1 + ω₁ * ∇Ri)^2 - 4 * ∇Ri)))
```

src/turbulence_functions.jl
yairchn: it should be a namelist parameter - will expose it after 134 is merged

yairchn: should I change it now ?
is there anything else we should do?

charleskawczynski (approved):
I think we're good to merge this 🚀

yairchn:
Bors r+

PR #134, opened 2021-08-16
Add internal experimental clima parameters

charleskawczynski:


yairchn (approved):
LGTM

haakon-e:
Should `c_div` also be added to 
https://github.com/CliMA/TurbulenceConvection.jl/blob/c0e2df602d592b30ee44b896effabf08a7eb547e/integration_tests/utils/parameter_set.jl#L15-L23
otherwise it would always be 0.4, regardless of the namelist value, no?

charleskawczynski:
> Shouldn't `c_div` also be added to

Good catch, thanks!

integration_tests/utils/parameter_set.jl
```diff
@@ -22,6 +22,7 @@ function create_parameter_set(namelist)
         MSLP = 100000.0, # or grab from, e.g., namelist[""][...]
         τ_cond_evap = TC.parse_namelist(namelist, "microphysics", "τ_cond_evap"; default = 10.0),
         c_ε = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment_factor"),
+        c_div = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment_massflux_div_factor"; default = 0.4),
```
haakon-e, line 25: As a small detail, if only DryBubble uses `c_div=0.4`, perhaps we should have `0.0` be the default? (especially since it has seemed like simulations are more unstable with non-zero c_div)

charleskawczynski: Oh, yeah, I had no idea, will update

yairchn: it is 0 by default 

https://github.com/CliMA/TurbulenceConvection.jl/blob/7ed6136fd93caf8aea35a4cbc3c8952c4b13bdf5/integration_tests/utils/generate_namelist.jl#L75

haakon-e (approved):
LGTM

PR #136, opened 2021-08-17
Separate the update and initialization in TurbulencePrognosticTKE

yairchn:
In this PR I am taking the first step to writing the model equations as RHS and LHS. To that end I am starting by taking anything relating initialization out of the update function in TurbulencePrognosticTKE. 
Doing so, I found that some cases with variable ustar has zero ustar at the initialization which leads to nan in the surface covariances. To avoid this I added a ustar value for these cases that is over written by compute_ustar after initialization. 

The PR has behavioral changes that are negligible to my opinion.


src/Turbulence_PrognosticTKE.jl
```diff
@@ -1,10 +1,24 @@
...
+        decompose_environment(self, GMV, "values")
+        saturation_adjustment(self.EnvThermo, self.EnvVar)
+        buoyancy(self.UpdThermo, self.UpdVar, self.EnvVar, GMV, self.extrapolate_buoyancy)
     else
         initialize(self.UpdVar, GMV)
+        decompose_environment(self, GMV, "values")
     end
+    update_inversion(self, GMV, Case.inversion_option)
+    self.wstar = get_wstar(Case.Sur.bflux, self.zi)
+    microphysics(self.EnvThermo, self.EnvVar, self.Rain, TS.dt)
```
yairchn, line 20: @trontrytel I think this microphysics call is needed - but I am not sure

yairchn: I will start by including it and maybe remove it later PR

src/Turbulence_PrognosticTKE.jl
```diff
@@ -260,7 +279,8 @@ function update(self::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBas
     buoyancy(self.UpdThermo, self.UpdVar, self.EnvVar, GMV, self.extrapolate_buoyancy)
     compute_eddy_diffusivities_tke(self, GMV, Case)
     update_GMV_ED(self, GMV, Case, TS)
-    compute_covariance(self, GMV, Case, TS)
+    compute_covariance_rhs(self, GMV, Case, TS)
```
yairchn, line 282: here 'compute_covariance' was split into 'compute_covariance_rhs' and 'update_covariance' as these two functions will be called separately in a future PR

haakon-e (approved):
Let's merge

yairchn:
Bors r+

costachris (approved):

PR #138, opened 2021-08-18
Organize RHS round 2

yairchn:
In this PR I have organized the function calls in 'update' function in TurbulencePrognosticTKE in three chunks:
1. compute RHS (for Upd, Env and GMV) 
2. update prognostic variables  (for Upd, Env and GMV) 
3. set values (back-out env values, set old with value and set value with new before the next step, cleanups etc.)

Tables are updated and there are behavioral changes - but these are small and I believe can be addressed by the tuning.


src/Turbulence_PrognosticTKE.jl
```diff
@@ -8,13 +8,12 @@ function initialize(
...
     if Case.casename == "DryBubble"
         initialize_DryBubble(self.UpdVar, GMV, Ref)
-        decompose_environment(self, GMV, "values")
-        saturation_adjustment(self.EnvThermo, self.EnvVar)
-        buoyancy(self.UpdThermo, self.UpdVar, self.EnvVar, GMV, self.extrapolate_buoyancy)
     else
         initialize(self.UpdVar, GMV)
-        decompose_environment(self, GMV, "values")
     end
+    decompose_environment(self, GMV, "values")
```
yairchn, line 14: here I found that the diagnostic calls can be done for both case in the initialization without change to the model outcome

haakon-e (approved):

yairchn:
Bors r+

PR #139, opened 2021-08-19
replace latent heat

yairchn:
This PR replaces Latent heat only throughout the code


src/Surface.jl
```diff
@@ -1,3 +1,5 @@
+import Thermodynamics
```
charleskawczynski, line 1: I think this, and the next line, are defined in `src/TurbulenceConvection.jl`, so this is probably not needed.

src/microphysics_functions.jl
```diff
@@ -1,18 +1,21 @@
 
+import Thermodynamics
```
charleskawczynski, line 2: This and next line are needed— we only need it once in `src/`

yairchn:
@charleskawczynski  do you know what all tests are red here ? CI and all ?

trontrytel (approved):
The artifact plots look reasonable. The CI is broken, but it does not look to be related to this PR.

PR #141, opened 2021-08-21
Remove latent_heat, cpm_c and pv_star_c

yairchn:


yairchn:
small behavior change in GABLS is observed when removing qv_star_t in Cases.jl

yairchn:
Not to be merged until finished

PR #142, opened 2021-08-21
move all upd advections to extren funcs

yairchn:
In this PR the advection of updraft variables are computed by external functions as preparation to a RHS - LHS separation of the equations


src/utility_functions.jl
```diff
@@ -34,3 +34,36 @@ function lamb_smooth_minimum(l, lower_bound, upper_bound)
...
     return smin
 end
+
+function upwind_advection_area(self, i, k, dzi, ρ0_half)
+    whalf_kp = interp2pt(self.UpdVar.W.values[i, k - 1], self.UpdVar.W.values[i, k])
+    whalf_k = interp2pt(self.UpdVar.W.values[i, k - 2], self.UpdVar.W.values[i, k - 1])
+
+    m_kp = (ρ0_half[k] * self.UpdVar.Area.values[i, k] * whalf_kp)
+    m_k = (ρ0_half[k - 1] * self.UpdVar.Area.values[i, k - 1] * whalf_k)
+    return -dzi * (m_kp - m_k) / ρ0_half[k]
```
yairchn, line 44: moving this 1/ρ0_half[k] out to the equation (consistently) will cause some changes in Soares. I am not sure why. If we could move it I think we could unify 'upwind_advection_area' and 'upwind_advection_scalar'

src/utility_functions.jl
```diff
@@ -34,3 +34,36 @@ function lamb_smooth_minimum(l, lower_bound, upper_bound)
     smin = num / den
     return smin
 end
+
+function upwind_advection_area(self, i, k, dzi, ρ0_half)
```
charleskawczynski, line 38: I would prefer that we write this in the form

```julia
upwind_advection_area(w_vals, a_vals, ρ0_half, grid, i, k)
```
We don't want the operator to be aware of `EDMF_PrognosticTKE` (`self`). And passing in `grid` allows us to flexibly change the form of the operator.

charleskawczynski: And similar below

haakon-e (approved):
LGTM

yairchn:
Bors r+

PR #153, opened 2021-08-24
Use normal thermodynamic functions in dycoms

yairchn:


yairchn:
It turns out that dycoms is not really changed by using the normal thermodynamic functions in its initialization and radiation.

charleskawczynski (approved):
LGTM

PR #156, opened 2021-08-25
use thermodynamics.jl

yairchn:
In this PR I am incrementally switching from thermodynamic_functions.jl to thermodynamics.jl
So far I have switched:
- latent_heat
- cpm_c
- pv_star_c
- qv_star_c
- RH
- pv_c
- exner_c
- theta_c
- t_to_thetali_c
- alpha_c
- theta_rho_c
- thetali_c
- thetali_virt_c (using PhaseEquil_pTq in TurbProgTKE 363 for now.)
- remove rho_c (using 'PhaseEquil_pTq' as an intermediary between eos and rho). Removing eos would be a next stage.)
* I am using (pp = 0,0,0) now only in buoyancy gradients computed in mixing length and TKE buoy source term


src/EDMF_Environment.jl
```diff
@@ -356,11 +353,16 @@ function sgs_quadrature(self::EnvironmentThermodynamics, EnvVar::EnvironmentVari
...
+            self.th_dry[k] =
+                TD.dry_pottemp_given_pressure(param_set, outer_env[i_T_dry], self.Ref.p0_half[k], phase_part)
+
             self.t_cloudy[k] = outer_env[i_T_cld]
             self.qv_cloudy[k] = outer_env[i_qt_cld] - outer_env[i_ql]
             self.qt_cloudy[k] = outer_env[i_qt_cld]
-            self.th_cloudy[k] = theta_c(self.Ref.p0_half[k], outer_env[i_T_cld])
+            phase_part = TD.PhasePartition(self.qt_cloudy[k], 0.0, 0.0)
+            self.th_cloudy[k] =
+                TD.dry_pottemp_given_pressure(param_set, outer_env[i_T_cld], self.Ref.p0_half[k], phase_part)
```
charleskawczynski, line 365:
```suggestion
            ts = TD.PhaseEquil_pTq(param_set, self.Ref.p0_half[k], outer_env[i_T_cld], self.qt_cloudy[k])
            self.th_cloudy[k] = TD.dry_pottemp(ts)
```
You'll want to test this out, but I believe that this should give the same result (`T, p` and `q_tot` are specified).

src/EDMF_Environment.jl
```diff
@@ -356,11 +353,16 @@ function sgs_quadrature(self::EnvironmentThermodynamics, EnvVar::EnvironmentVari
             # update cloudy/dry variables for buoyancy in TKE
             EnvVar.cloud_fraction.values[k] = outer_env[i_cf]
             self.qt_dry[k] = outer_env[i_qt_dry]
-            self.th_dry[k] = theta_c(self.Ref.p0_half[k], outer_env[i_T_dry])
+            phase_part = TD.PhasePartition(self.qt_dry[k], 0.0, 0.0)
+            self.th_dry[k] =
+                TD.dry_pottemp_given_pressure(param_set, outer_env[i_T_dry], self.Ref.p0_half[k], phase_part)
```
charleskawczynski, line 358:
```suggestion
            ts = TD.PhaseEquil_pTq(param_set, self.Ref.p0_half[k], outer_env[i_T_dry], self.qt_dry[k])
            self.th_dry[k] = TD.dry_pottemp(ts)
```
You'll want to test this out, but I think this should work.

integration_tests/utils/Cases.jl
```diff
@@ -190,7 +185,9 @@ function initialize_profiles(self::CasesBase{Soares}, Gr::Grid, GMV::GridMeanVar
 
     @inbounds for k in real_center_indicies(Gr)
         GMV.H.values[k] = theta[k]
-        GMV.T.values[k] = theta[k] * exner_c(Ref.p0_half[k])
+        phase_part = TD.PhasePartition(GMV.QT.values[k], GMV.QL.values[k], 0.0)
+        Π = TD.exner_given_pressure(param_set, Ref.p0_half[k], phase_part)
+        GMV.T.values[k] = theta[k] * Π
```
charleskawczynski, line 190:
```suggestion
        ts = TD.PhaseEquil_pθq(param_set, Ref.p0_half[k], theta[k], GMV.QT.values[k])
        Π = TD.exner(ts)
        GMV.T.values[k] = TD.air_temperature(ts)
```
Does this work?

charleskawczynski (commented):
Overall, I like the idea of incrementally using Thermodynamics, but this PR uses a _lot_ of internal thermodynamic methods, and I'm not sure that this is a step in the right direction. I understand that there are special cases where one needs to use internal Thermo methods, but there are many instances in this PR where they can be avoided.

Since this PR is so large, but nicely preserves the physical results, perhaps we can keep it open and slowly peel off pieces in smaller separate PRs (that avoid using internal thermodynamic methods). That is to say, always create a thermo state from available quantities, and then call whatever method is needed, e.g., `TD.air_temperature(ts::ThermodynamicState)`.

PR #170, opened 2021-08-27
Use more discrete operators

charleskawczynski:


charleskawczynski:
MSE tables needed to be updated, I believe due to precision errors (the order of operations has changed).

PR #173, opened 2021-08-28
Add fcut and f∇, and use appropriately

charleskawczynski:
When trying to add bcs to the interp operators, I realized that `cut` was specifically written for cell centers--and we need them to also work for cell faces.

This PR
 - Adds `fcut` and `f∇` and uses them appropriately
 - Splits pressure calculations so that the correct indices are traversed (we want to use `real_center_indices` when updating cell center data and `real_face_indices` when updating cell face data)
 - Adds `f`/`c` prefix to other methods to make the use more clear.
 - `∇w_up` in `compute_updraft_closures` is now computed from a centered difference with a 3-point stencil


charleskawczynski:
Results look fine AFAICT

PR #178, opened 2021-08-29
Use more centered gradients

charleskawczynski:
`c∇_onesided` is currently a _downwind_ scheme, which means that it can't be used in the upwind advection operators. Rather than adding yet another stencil/operator category, I figured it'd be easier to switch calls from `c∇_onesided` to `c∇`, and then change the implementation of `c∇_onesided` so that it's actually upwind, and then a new stencil/operator is not needed.

I'm not exactly sure if this is consistent with how the time-implicit operators are defined but, AFAIK only the advection terms _really_ need to be upwinded, and these changes are only applied to other sources.


charleskawczynski:
The Dycoms results look a bit bad.

In the sake of time, let's just go with the refactoring approach.

PR #184, opened 2021-08-30
Replace interp2pt with bc ops

charleskawczynski:


charleskawczynski:
I'm realizing that we really can't continue until we fix a lot of problems with the `solve_updraft` function..

charleskawczynski:
Closing in favor of more updated/less stale PRs

PR #186, opened 2021-08-30
Restructure main to have single update function

charleskawczynski:
This PR restructures `main.jl` so that there is a single update function. This means that
```julia
        TC.zero_tendencies(self.GMV)
        Cases.update_surface(self.Case, self.GMV, self.TS)
        Cases.update_forcing(self.Case, self.GMV, self.TS)
        Cases.update_radiation(self.Case, self.GMV, self.TS)

        TC.update(self.GMV, self.TS)
        TC.update_GMV_diagnostics(self.Turb, self.GMV)
```
moves inside `TC.update`. This way, we can decompose `TC.update` into
 - Update aux (compute diagnostic and non-local quantities)
 - Compute RHS (compute `tendencies`, not `new`)
 - Solve eqs (`values = values + Δt*tendencies` + implicit solves)


charleskawczynski:
Results look fine

PR #187, opened 2021-08-30
Restructure update function

charleskawczynski:


charleskawczynski:
A good deal of this was fixed in #192.

PR #188, opened 2021-08-31
Fix regression test for cases without LES data

charleskawczynski:
I realized that some of our MSE tables were showing zero, which didn't make sense. It turns out that those cases without LES results (SP, life_cycle_Tan2018, DryBubble) were able to regress without failing. This PR fixes this issue and ensures that we compare with at least one dataset:
 - LES takes first precedence
 - SCAMPy takes second precedence
 - TC.jl main takes third precedence


yairchn (approved):
LGTM

PR #192, opened 2021-08-31
Fuse grid mean tendencies into single loop

charleskawczynski:
Looking through the code, I realized that _all_ grid mean quantities seem to be updated in a standardized way, and I think we can fuse all the tendencies into a single loop.

This is still not a great final version, but we can start thinking about how we want to dispatch to add sources for a given model.

### DIFF
--- a/src/Turbulence_PrognosticTKE.jl
+++ b/src/Turbulence_PrognosticTKE.jl
@@ -237,17 +237,82 @@ update_forcing(Case::CasesBase, GMV::GridMeanVariables, TS::TimeStepping) =
 update_radiation(Case::CasesBase, GMV::GridMeanVariables, TS::TimeStepping) =
     error("update_radiation should be overloaded in case-specific methods (in Cases.jl)")
 
+function compute_gm_tendencies!(grid, Case, GMV, Ref, TS)
+    GMV.U.tendencies .= 0
+    GMV.V.tendencies .= 0
+    GMV.QT.tendencies .= 0
+    GMV.H.tendencies .= 0
+
+    @inbounds for k in real_center_indicies(grid)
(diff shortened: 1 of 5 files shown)


yairchn (approved):
LGTM

src/Turbulence_PrognosticTKE.jl
```diff
@@ -237,17 +237,82 @@ update_forcing(Case::CasesBase, GMV::GridMeanVariables, TS::TimeStepping) =
...
+        end
+
+        if force_type(Case.Fo) <: ForcingStandard
+            if Case.Fo.apply_subsidence
+                GMV.H.tendencies[k] -= ∇H * Case.Fo.subsidence[k]
+                GMV.QT.tendencies[k] -= ∇q_tot * Case.Fo.subsidence[k]
+            end
+            GMV.H.tendencies[k] += Case.Fo.dTdt[k] / Π
+            GMV.QT.tendencies[k] += Case.Fo.dqtdt[k]
+        end
```
haakon-e, line 276: A bit late to the party here, but isn't this kind of pattern ripe for multiple dispatch?
I'm assuming at most one of `force_type(Case.Fo) <: ...` should ever be applied

haakon-e: Oh, I see you mentioned this in the description. If we're returning to this later, LGTM for now.

charleskawczynski: Yes, definitely. That will be the long term solution. I figure that collecting all of these sources into a single loop is a good first step. Then we can dispatch on which sources to include (within the loop). This is basically how ClimateMachine did it

PR #193, opened 2021-09-01
Split some loops in solve_updraft

charleskawczynski:


charleskawczynski:
Results still look good! 🚀

PR #196, opened 2021-09-01
Replace thermodynamics except for rho_c and eos 

yairchn:
In this PR I am incrementally switching from thermodynamic_functions.jl to thermodynamics.jl
So far I have switched:

latent_heat
cpm_c
pv_star_c
qv_star_c
RH
pv_c
exner_c
theta_c
t_to_thetali_c
alpha_c
theta_rho_c
thetali_c
thetali_virt_c (using PhaseEquil_pTq in TurbProgTKE 363 for now.)
remove rho_c (using 'PhaseEquil_pTq' as an intermediary between eos and rho). Removing eos would be a next stage.)
I am using (pp = 0,0,0) now only in buoyancy gradients computed in mixing length and TKE buoy source term


charleskawczynski (approved):
So long the results look okay, these changes look fantastic

charleskawczynski:
Looking through the results now, they look great! Thanks, @yairchn!

PR #197, opened 2021-09-01
Use RootSolvers.jl in eos

charleskawczynski:
This PR adds the use of RootSolvers.jl in `eos`, so that we disentangle the numerical method with the equations being solved. This also changes the numerical method from Secant method to Newton method (with auto-diff).


charleskawczynski:
Results look solid! 🚀

PR #203, opened 2021-09-02
add a pointwise buoy gradinet function

yairchn:


src/types.jl
```diff
@@ -152,6 +152,51 @@ Base.@kwdef struct MinDisspLen{FT, T}
...
+    ∂qt∂z::FT
+    "potential temperature gradient"
+    ∂θl∂z::FT
+    "reference pressure"
+    p0::FT
+    "cloud fraction"
+    en_cld_frac::FT
+    "specific volume"
+    alpha0::FT
+end
```
charleskawczynski, line 198: I think it'd be really nice if we could use this data structure instead:

```julia
Base.@kwdef struct Tan2018{TS, FT}
    "non cloudy thermodynamic state"
    ts_dry::TS
    "cloudy thermodynamic state"
    ts_cloudy::FT
    "total specific humidity gradient"
    ∂qt∂z::FT
    "potential temperature gradient"
    ∂θl∂z::FT
    "cloud fraction"
    en_cld_frac::FT
end
```

charleskawczynski (approved):
Overall, these changes are definitely an improvement. One additional step that we could take (I'm fine with doing it in this PR or saving for a subsequent PR) is storing thermodynamic states in `Tan2018` (I commented there).

src/types.jl
charleskawczynski: Let's leave this for a later PR.

PR #214, opened 2021-09-03
Avoid accessing ghost points in Microphysics

charleskawczynski:
I'm not exactly sure how this algo works, but it's currently accessing ghost points, and we're trying to avoid them. This PR adds a hack to avoid accessing ghost points.


trontrytel:
This is supposed to be an upstream advection of rain falling down in the domain. 

I would say that whatever is done for upstream advection up in the domain of our EDMF updrafts should be done here as well. (Just remembering that the rain is falling down)

charleskawczynski:
Sounds good, I think this looks good, thanks!

PR #215, opened 2021-09-03
Replace more operators

charleskawczynski:
This PR
 - uses unicode
 - Replaces a gradient operator
 - uses `findlast`

The remaining `k - 1` calls do not necessarily access ghost points, but could and we'll see what happens.


charleskawczynski:
Closing in favor of #216.

PR #216, opened 2021-09-03
Replace more operators

charleskawczynski:
This builds on top of #215.


charleskawczynski:
There's some small differences, but overall the results look good. There's a tiny regression with SP tke profiles, but I think it's acceptable for now. Going to squash and merge.

PR #217, opened 2021-09-03
Replace gradient with operator

charleskawczynski:


charleskawczynski:
I think that this may be the last one! 🎉

charleskawczynski:
Nooooooo bors r-

PR #218, opened 2021-09-03
Restructure grid to not have ghost points

charleskawczynski:
Well, it's running but it fails in saturation adjustment somewhere. I think there are some bug fixes in here, but I don't think they'll have an impact on `main` if we fix in a separate PR.

This may take a while. Here's a sketch:
Screen Shot 2021-09-05 at 2 06 42 PM


src/EDMF_Updrafts.jl
```diff
@@ -375,16 +373,21 @@ function buoyancy(
...
-                    qt = UpdVar.QT.values[i, k - 1]
-                    h = UpdVar.H.values[i, k - 1]
-                    sa = eos(param_set, self.Ref.p0_half[k], qt, h)
-                    qv = qt - sa.ql
-                    t = sa.T
-                    rho = rho_c(self.Ref.p0_half[k], t, qt, qv)
-                    UpdVar.B.values[i, k] = buoyancy_c(param_set, self.Ref.rho0_half[k], rho)
-                    ts = TD.PhaseEquil_pθq(param_set, self.Ref.p0_half[k], h, qt)
-                    UpdVar.RH.values[i, k] = TD.relative_humidity(ts)
+                elseif k > kc_surf
```
yairchn, line 376: why did you remove the area fraction condition here?
this could cause the code to run eos on bad values with very small area fraction.

charleskawczynski: With the new grid, we can no longer access `UpdVar.Area.values[i, k - 1]` at `k = 1`. So this had to become a nested `if` block. It's functionally the same as before, but now it protects against accessing `UpdVar.Area.values[i, k - 1]`.

That said, we should really fix this logic.

src/Turbulence_PrognosticTKE.jl
```diff
@@ -6,18 +6,28 @@ function initialize(
     Ref::ReferenceState,
     TS::TimeStepping,
 )
+    check_nans(GMV); check_nans(self)
+    set_updraft_surface_bc(self, GMV, Case)
```
charleskawczynski, line 10: Since we're only changing the grid, this really shouldn't be necessary, and may be hiding the root problem..

charleskawczynski:
For TRMM, we'll likely need to change `hcat(map(xrange(0, 36)) do tt....` to first exclude ghost points in `Gr.z_half`.

src/Turbulence_PrognosticTKE.jl
charleskawczynski: Removed this, and fixed the issue.

charleskawczynski:
It looks like this is working!

charleskawczynski:
Also, I realized after looking at results-- this PR introduces purely refactoring changes, but since the size of the face fields changes, our MSE values will change because their normalized by the data scale (which depends on the _length_ of the vectors).

charleskawczynski:
Hmm, after looking a bit closer, I think there are some behavioral changes-- but it seems not worth investigating because
 - The results in this PR look great
 - Running life cycle has _exactly_ the same results for all fields after 5000 timesteps (which our plots don't fully seem to indicate). And, because our comparison requires shifting indices (SCAMPy skipped plotting the top face, and plotted cell face data on cell centers). One thought was that this could be a plotting artifact but, actually, there are negative velocities in main, whereas this PR seems to _better_ filter.

I'm going to update the tables and move forward with this! 🚀

yairchn:
The figures looks great!
I will review the PR today

integration_tests/utils/generate_namelist.jl
```diff
@@ -120,7 +120,7 @@ function default_namelist(case_name::String)
     # From namelist
     namelist_defaults["grid"] = Dict()
     namelist_defaults["grid"]["dims"] = 1
-    namelist_defaults["grid"]["gw"] = 2
+    namelist_defaults["grid"]["gw"] = 0
```
yairchn, line 123: Yeh!!

charleskawczynski:
On a final thought, I really suspect that these changes are some sort of diagnostic artifact--it would be mind boggling if any differences in `w_updraft` (the only affected field) did not impact _any_ other fields.

yairchn (approved):
LGTM! great work

charleskawczynski:
Just to note, the next PR _may_ run into issue with `compute_mse.jl` not correctly handling main data when its vector sizes have changed

PR #223, opened 2021-09-04
replace Magnus formula with TD.saturation_vapor_pressure

yairchn:


charleskawczynski (approved):
Wow, I thought we previously saw notable differences, here. Maybe we've fixed some bugs that previously had bad interaction with these changes?

Overall I'm fine with the PR-- but I'm also now curious if we can replace more thermodynamics wholesale 🤔

yairchn:
Bors r+

PR #225, opened 2021-09-04
Avoid OOB (without ghost points) in buoyancy computation

charleskawczynski:
~Looking at this part in the code that I found a bit confusing, I realized that the `elseif` block seems to be valid at a single (special) point above which the updraft area fraction is non-zero, then the `else` block handles _every_ point after that. This PR tries to simply remove this logic--since we're updating buoyancy there based on neighboring data.~

Let's just avoid out of bounds indexing in these computations for now..


yairchn:
I agree that it could be made better but if you look at the ql_mean profile in DYCOMS you will see that we have a behavioral change that is not desired.

PR #226, opened 2021-09-05
fuse operations in environment rho functions 

yairchn:


src/Turbulence_PrognosticTKE.jl
```diff
@@ -461,14 +450,14 @@ function compute_mixing_length(self, obukhov_length, ustar, GMV::GridMeanVariabl
         ∂θv∂z = c∇(thv_cut, grid, k; bottom = SetGradient(0), top = Extrapolate())
         # compute ∇Ri and Pr
         ∇_Ri = gradient_Richardson_number(bg.∂b∂z_θl, bg.∂b∂z_qt, Shear², eps(0.0))
-        self.prandtl_nvec[k] = turbulent_Prandtl_number(obukhov_length, ∇_Ri, Pr_n, ω_pr)
+        self.prandtl_nvec[k] = turbulent_Prandtl_number(Case.Sur.obukhov_length, ∇_Ri, Pr_n, ω_pr)
```
charleskawczynski, line 453: I'd prefer the unpacking approach, as this variable is used multiple times, and we should minimize the dependence on the structure of our data structures. So we should probably revert these types of changes

src/Turbulence_PrognosticTKE.jl
```diff
@@ -1304,20 +1232,21 @@ function compute_covariance_rhs(self::EDMF_PrognosticTKE, GMV::GridMeanVariables
...
+    @inbounds for k in real_center_indicies(grid)
+        compute_covariance_entr(self, k, grid, ref_state, 0.5, en.TKE, up.W, up.W, en.W, en.W, gm.W, gm.W)
+        compute_covariance_entr(self, k, grid, ref_state, 1.0, en.Hvar, up.H, up.H, en.H, en.H, gm.H, gm.H)
+        compute_covariance_entr(self, k, grid, ref_state, 1.0, en.QTvar, up.QT, up.QT, en.QT, en.QT, gm.QT, gm.QT)
+        compute_covariance_entr(self, k, grid, ref_state, 1.0, en.HQTcov, up.H, up.QT, en.H, en.QT, gm.H, gm.QT)
+        compute_covariance_shear(self, k, grid, ref_state, KM, 0.5, gm, en.TKE, up.W.values, up.W.values, en.W.values, en.W.values)
+        compute_covariance_shear(self, k, grid, ref_state, KH, 1.0, gm, en.Hvar, up.H.values, up.H.values, en.H.values, en.H.values)
+        compute_covariance_shear(self, k, grid, ref_state, KH, 1.0, gm, en.QTvar, up.QT.values, up.QT.values, en.QT.values, en.QT.values)
+        compute_covariance_shear(self, k, grid, ref_state, KH, 1.0, gm, en.HQTcov, up.H.values, up.QT.values, en.H.values, en.QT.values)
+    end
```
charleskawczynski, line 1249: I'm not sure I like this pattern-- I like the idea of fusing these loops, but I think we can cook up a better solution.

src/Turbulence_PrognosticTKE.jl
```diff
@@ -497,22 +486,11 @@ function compute_mixing_length(self, obukhov_length, ustar, GMV::GridMeanVariabl
...
-    param_set = parameter_set(self)
-    c_m = CPEDMF.c_m(param_set)
-    compute_mixing_length(self, Case.Sur.obukhov_length, Case.Sur.ustar, GMV)
-    KM = diffusivity_m(self)
-    KH = diffusivity_h(self)
-    @inbounds for k in real_center_indicies(grid)
-        lm = self.mixing_length[k]
-        pr = self.prandtl_nvec[k]
-        KM.values[k] = c_m * lm * sqrt(max(self.EnvVar.TKE.values[k], 0.0))
-        KH.values[k] = KM.values[k] / pr
```
charleskawczynski, line 515: I like how this is fused 🚀

src/Turbulence_PrognosticTKE.jl
yairchn: Do you prefer to have a "model" for each functions ? that would be much - right ?

charleskawczynski: I think let’s leave it the way it is for now. When we are using ClimaCore, we may be able to write a cleaner solution

src/Turbulence_PrognosticTKE.jl
```diff
@@ -846,6 +822,10 @@ function compute_updraft_closures(self::EDMF_PrognosticTKE, GMV::GridMeanVariabl
                 self.nh_pressure_drag[i, k] = 0.0
             end
             self.nh_pressure[i, k] = self.nh_pressure_b[i, k] + self.nh_pressure_adv[i, k] + self.nh_pressure_drag[i, k]
+            w_up_c = interpf2c(self.UpdVar.W.values, grid, k, i)
+            w_en_c = interpf2c(self.EnvVar.W.values, grid, k)
+            press_c = interpf2c(self.nh_pressure, grid, k, i)
+            self.EnvVar.TKE.press[k] += (w_en_c - w_up_c) * press_c
```
charleskawczynski, line 828: We can’t do this, EnvVar.TKE.press is a cell center field, and we’re looping over cell faces

src/Turbulence_PrognosticTKE.jl
```diff
@@ -816,6 +791,7 @@ function compute_updraft_closures(self::EDMF_PrognosticTKE, GMV::GridMeanVariabl
     kf_surf = kf_surface(grid)
     kc_toa = kc_top_of_atmos(grid)
     @inbounds for k in real_face_indicies(grid)
+        self.EnvVar.TKE.press[k] = 0.0
```
charleskawczynski, line 794: We can’t do this, EnvVar.TKE.press is a cell center field, and we’re looping over cell faces

src/Turbulence_PrognosticTKE.jl
```diff
@@ -1749,49 +1653,46 @@ end
...
-            ∇w_en = f∇(self.EnvVar.W.values, grid, k; w_bcs...)
-            Envcov_ = -self.horiz_K_eddy[i_last, k] * ∇w_en
-        else
-            Envcov_ = env_covar.values[k]
-        end
+    # TODO: report bug: i used outside of scope.
+    # This is only valid (assuming correct) for 1
+    # updraft.
+    i_last = last(xrange(self.n_updrafts))
+    if tke_factor == 0.5
```
charleskawczynski, line 1675: We should use `is_tke` here instead

PR #227, opened 2021-09-05
remove one k loop in compute_updraft_closures

yairchn:


charleskawczynski (commented):
Oh, we can't do this, actually. One loop is over `real_face_indices`, and the other over `real_center_indices`.

PR #220, opened 2021-09-04
Apply some bug fixes

charleskawczynski:
This PR applies some bug fixes to bugs found in #218. I think this needs careful review.

PR #231, opened 2021-09-06
Remove center_indices and face_indices

charleskawczynski:
This PR removes `center_indices` and `face_indices` in preparation for #218


charleskawczynski:
This PR needs to wait on #220.

PR #240, opened 2021-09-08
Fix another grid traversal

charleskawczynski:
Peeled off from #218.


charleskawczynski:
Results look fine with this fix.

PR #244, opened 2021-09-09
Fix fields accessed in center loop traversals

charleskawczynski:
Another peel off from #218, thanks @trontrytel for the rubber duck!


charleskawczynski:
The most notable difference here seems to be in the DYCOMS tke mean, but it looks reasonable enough (and closer to SCAMPy, actually).

PR #249, opened 2021-09-10
Error on non-convergence in eos, up tolerance

charleskawczynski:
Right now, in `main`, life cycle, and several other cases are printing warnings about saturation adjustment non-convergence at many points in space across many timesteps. This has really bloated the log, and we should really error on non-convergence (and adjust the tolerance / number of iterations if needed).

This PR changes eos from using Newtons method to Secant method and errors on non-convergence.


charleskawczynski:
Newtons method was not converging for some inputs, so let's Secant method. At first I tested Secant method locally with a higher tolerance but the results looked terrible, so I lowered it back down.

charleskawczynski:
Results look okay with these changes.

PR #252, opened 2021-09-11
Add dependence on ClimaCore branch

charleskawczynski:
This PR adds the dependence on the `turbulence_convection_checkpoint_01` branch. The idea is that we can increment our branch (next to `turbulence_convection_checkpoint_02`) as we need/want updates from clima core. Depending on a specific branch, and not on `main`, allows us to update clima core when we want. That way, our CI isn't affected by something failing in ClimaCore `main`.

This will somewhat change user/developer workflows, since we're now checking the Manifest files into the repo--developers shouldn't have local version issues anymore, since the concrete package versions should be specified (except when, of course, their trying to upgrade packages). This leads me to a proposal:

Because we must update packages in the Manifest any time we want updates (breaking or non-breaking), I propose that we do these updates in _separate_ PRs from any development PRs. That is, if we're updating (a) package(s), we should be making changes _only required to upgrade the package_.

cc @yairchn, @haakon-e, @trontrytel, @ilopezgp, @jiahe23, @szy21


haakon-e:
seems reasonable to me.

yairchn (approved):
Just to double check you meant to add these Manifest and Project files to the code, right ?

charleskawczynski:
> you meant to add these Manifest and Project files to the code, right ?

Yes

charleskawczynski:
Unfortunately, this means that we'll need to depend on Plots (at least it's the simplest solution). I'll open an issue to revert these changes once ClimaCore is first released.

haakon-e:
As an aside (if there's a desire to keep large dependencies to a minimum), I think I'm the one who added all of DiffEq as a dependency. If there's a way to add sub modules only for Stochastic DiffEq, we could look into that at some point.

jakebolewski:
We plan on registering the ClimaCore package and making a release the coming week to hopefully (largely) alleviate the need for this setup.  Then probably you would want to pin a version and work with upstream to manage update.

charleskawczynski:
> As an aside (if there's a desire to keep large dependencies to a minimum), I think I'm the one who added all of DiffEq as a dependency. If there's a way to add sub modules only for Stochastic DiffEq, we could look into that at some point.

I think it's fine for now. As we improve things I'm hoping that we'll be able to trim out some of the deps and move them into the test environment

PR #255, opened 2021-09-13
Use ClimaCore grid

charleskawczynski:
This PR switches over to using a ClimaCore.jl grid and changes some variable names
 - `dz` -> `Δz`
 - `z_half` -> `zc`
 - `z` -> `zf`

I think that this should all work with [ClimaCore.jl's 173](https://github.com/CliMA/ClimaCore.jl/pull/173). We'll need to create a new branch to update to, or just point to a new SHA, maybe that's easier than creating new branches.

Since this PR changes `z_half` / `z` to `zc` and `zf`, we'll need a follow-up PR to change the names for the main branch.

### DIFF
--- a/docs/src/ReferenceStates.md
+++ b/docs/src/ReferenceStates.md
@@ -21,8 +21,8 @@ function export_ref_profile(case_name::String)
     case = Cases.CasesFactory(namelist, grid, ref_state)
     Cases.initialize_reference(case, grid, ref_state, Stats)
     NCDatasets.Dataset(joinpath(Stats.path_plus_file), "r") do ds
-        zc = ds.group["profiles"]["z_half"][:]
-        zf = ds.group["profiles"]["z"][:]
+        zc = ds.group["profiles"]["zc"][:]
+        zf = ds.group["profiles"]["zf"][:]
         ρc_0 = ds.group["reference"]["rho0_half"][:]
         pc_0 = ds.group["reference"]["p0_half"][:]
         αc_0 = ds.group["reference"]["alpha0_half"][:]
--- a/src/Grid.jl
+++ b/src/Grid.jl
@@ -1,62 +1,62 @@
-struct Grid{A1, FT}
+
+Base.@propagate_inbounds function Base.getindex(field::CC.Fields.FiniteDifferenceField, i::Integer)
+    Base.getindex(CC.Fields.field_values(field), i)
+end
+
+
+struct Grid{FT, SC, SF}
     zmin::FT
     zmax::FT
-    dz::FT
-    dzi::FT
+    Δz::FT
+    Δzi::FT
     nz::Int
-    nzg::Int
-    z_half::A1
-    z::A1
+    zc::SC
+    zf::SF
     function Grid(namelist)
 
         #Get the grid spacing
-        dz = namelist["grid"]["dz"]
+        Δz = namelist["grid"]["dz"]
+        nz = namelist["grid"]["nz"]
+        FT = Float64
+        z₀, z₁ = FT(0), FT(nz * Δz)
+        domain = CC.Domains.IntervalDomain(z₀, z₁, x3boundary = (:bottom, :top))
+        mesh = CC.Meshes.IntervalMesh(domain, nelems = nz)
+
+        cs = CC.Spaces.CenterFiniteDifferenceSpace(mesh)
+        fs = CC.Spaces.FaceFiniteDifferenceSpace(cs)
+        zc = CC.Fields.coordinate_field(cs)
+        zf = CC.Fields.coordinate_field(fs)
 
         #Set the inverse grid spacing
+        Δzi = 1.0 / Δz
 
-        dzi = 1.0 / dz
-
-        # Get the grid dimensions
-        nz = namelist["grid"]["nz"]
-        nzg = nz
-
-        # TODO: make cell centers and cell faces different sizes
-        z_half = zeros(nz)
-        z = zeros(nz + 1)
-
-        z .= map(1:(nz + 1)) do k
-            (k - 1.0) * dz
-        end
-        z_half .= map(1:nz) do k
-            (z[k] + z[k + 1]) / 2.0
-        end
-
-        zmin = z[1]
-        zmax = z[nzg + 1]
-        A1 = typeof(z)
+        zmin = minimum(parent(zf))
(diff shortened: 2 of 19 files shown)


src/Grid.jl
```diff
@@ -1,62 +1,58 @@
...
-    dz::FT
-    dzi::FT
+    Δz::FT
+    Δzi::FT
     nz::Int
-    nzg::Int
-    z_half::A1
-    z::A1
+    zc::SC
+    zf::SF
```
haakon-e, line 10: Are these in practice regular arrays, or some CC construct?

src/Grid.jl
```diff
@@ -1,62 +1,58 @@
...
 is_surface_center(grid::Grid, k::Int) = k == kc_surface(grid)
 is_toa_center(grid::Grid, k::Int) = k == kc_top_of_atmos(grid)
 is_surface_face(grid::Grid, k::Int) = k == kf_surface(grid)
 is_toa_face(grid::Grid, k::Int) = k == kf_top_of_atmos(grid)
 
-zc_surface(grid::Grid) = grid.z_half[kc_surface(grid)]
-zf_surface(grid::Grid) = grid.z[kf_surface(grid)]
-zc_toa(grid::Grid) = grid.z_half[kc_top_of_atmos(grid)]
-zf_toa(grid::Grid) = grid.z[kf_top_of_atmos(grid)]
+zc_surface(grid::Grid) = vec(grid.zc)[kc_surface(grid)]
```
haakon-e, line 49: Is the `vec` function related to the lack of support for slicing? (https://github.com/CliMA/ClimaCore.jl/pull/173#pullrequestreview-752224612)

src/Turbulence_PrognosticTKE.jl
```diff
@@ -1794,7 +1792,8 @@ function update_inversion(self::EDMF_PrognosticTKE, GMV::GridMeanVariables, opti
             ∇θ_liq = c∇_downwind(∇θ_liq_cut, grid, k; bottom = FreeBoundary(), top = SetGradient(0))
             if ∇θ_liq > ∇θ_liq_max
                 ∇θ_liq_max = ∇θ_liq
-                self.zi = grid.z[k]
+                # TODO: bugfix: we should be using grid.zc[k]
```
haakon-e, line 1795: do we need an issue for this?

docs/src/ReferenceStates.md
```diff
@@ -21,7 +21,7 @@ function export_ref_profile(case_name::String)
     case = Cases.CasesFactory(namelist, grid, ref_state)
     Cases.initialize_reference(case, grid, ref_state, Stats)
     NCDatasets.Dataset(joinpath(Stats.path_plus_file), "r") do ds
-        zc = ds.group["profiles"]["z_half"][:]
+        zc = ds.group["profiles"]["zc"][:]
         zf = ds.group["profiles"]["z"][:]
```
haakon-e, line 25: should this be `"zf"` instead of `"z"`?

haakon-e (approved):
Overall, LGTM

src/Grid.jl
charleskawczynski: They’re types— finite difference spaces, I think

src/Grid.jl
charleskawczynski: Yeah, I forgot to revert that, thanks!

src/Turbulence_PrognosticTKE.jl
charleskawczynski: Yeah, opened #256.

docs/src/ReferenceStates.md
charleskawczynski: Yes! Good catch!

src/NetCDFIO.jl
```diff
@@ -78,23 +78,23 @@ mutable struct NetCDFIO_Stats
 
         NC.Dataset(path_plus_file, "c") do root_grp
 
-            zf = Gr.z
-            zc = Gr.z_half
+            zf = vec(Gr.zf)
+            zc = vec(Gr.zc)
```
haakon-e, line 82: did you mean to remove `vec` here too?

src/ReferenceState.jl
```diff
@@ -76,8 +76,8 @@ function initialize(self::ReferenceState, grid::Grid, Stats::NetCDFIO_Stats)
     prob = ODEProblem(rhs, logp, z_span)
     sol = solve(prob, Tsit5(), reltol = 1e-12, abstol = 1e-12)
 
-    p .= sol.(grid.z)
-    p_half .= sol.(grid.z_half)
+    p .= sol.(vec(grid.zf))
+    p_half .= sol.(vec(grid.zc))
```
haakon-e, line 80: and here

src/NetCDFIO.jl
charleskawczynski: No, I imagine that NCDatasets needs vectors here

src/ReferenceState.jl
charleskawczynski: No, I think that DiffEq’s solution type wouldn’t work without calling `vec` on the field.

charleskawczynski:
@haakon-e, @yairchn, one thing that I'm not crazy about with this PR is that we have to leave `z_half` in a few places, where we read from NC files. Thoughts?

haakon-e:
I don't have a full overview of all these places -- can we update any .nc files if they're static? or would we have to update pycles too?

charleskawczynski:
> I don't have a full overview of all these places -- can we update any .nc files if they're static? or would we have to update pycles too?

Technically we only need to update the NC files, which would change our Artifact.toml (currently [git ignored](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/.gitignore#L34)).

There are some consequences of changing the NC file links, though. After changing it, we no longer can checkout older branches and expect them to run. We could simply add a new set of NC files to the cloud with this name change, but we obviously don't want to do that many times.

I'm almost in favor of leaving legacy names around for some time, idk. Going to sleep on it.

charleskawczynski:
> I don't have a full overview of all these places -- can we update any .nc files if they're static? or would we have to update pycles too?

It’s only for strings in accessing NC files.

I think it’s fine—our data structures will still be consistent, and I think we’ll always be interfacing with data/simulations that have other naming conventions/formulations

PR #257, opened 2021-09-13
move saturation adjusment to solve updrafts

yairchn:


charleskawczynski (commented):
This is moving things in the wrong direction, these are effectively computing diagnostics. We should move this up to the `update_aux` part of this function.

PR #258, opened 2021-09-13
Fix 256 bug

### LINKED ISSUE #256: Update inversion uses wrong coordinate
We should be using z at cell centers [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/Turbulence_PrognosticTKE.jl#L1797) since we're traversing cell centers.

charleskawczynski:
Closes #256.


charleskawczynski:
Results look good!

PR #264, opened 2021-09-15
Make reference state immutable

charleskawczynski:
This PR:
 - Makes `ReferenceState` immutable
 - Moves the reference state plots in the docs into a .jl file (easier to edit with syntax highlighting)


yairchn (approved):
LGTM

PR #265, opened 2021-09-15
Finalize the steps in thermodynamic change

yairchn:
As a compromise measure between using only thermodynamic states from Thermodynamics.jl (which requires massive changes to make it Anelastic) and using internal Thermodynamics.jl methods we added here a thermodynamic constructor in the new file 'src/inelastic_thermodynamics.jl' with several in-house functions that allow our thermodynamic state to be Anelastic. Mind that we are still using TD.PhaseEqauil_pθq in ReferenceState.jl


src/anelastic_thermodynamic.jl
```diff
@@ -0,0 +1,202 @@
...
+
+"""
+    q_vap_saturation_from_partial_pressures(param_set, q_tot, p, p_v_sat)
+Compute the saturation specific humidity, given
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `q_tot` total water specific humidity,
+ - `p` air pressure,
+ - `p_v_sat` saturation vapor pressure
+"""
+function q_vap_saturation_from_partial_pressures(param_set::APS, q_tot::FT, p::FT, p_v_sat::FT) where {FT <: Real}
```
trontrytel, line 118: Could this function be part of Thermodynamics.jl? I will need something similar when computing supersaturation for coupling CloudMicrophysics.jl with TC.jl

src/anelastic_thermodynamic.jl
```diff
@@ -0,0 +1,202 @@
...
+    p::FT,
+    q_tot::FT,
+    phase_type::Type{<:TD.ThermodynamicState},
+) where {FT <: Real}
+
+    p_v_sat = TD.saturation_vapor_pressure(param_set, T, TD.Liquid())
+    q_v_sat = q_vap_saturation_from_partial_pressures(param_set, q_tot, p, p_v_sat)
+
+    _liquid_frac = 1
+    # # TODO: write and use saturation_excess_given_p (that calls q_vap_saturation_from_partial_pressures)
```
trontrytel, line 159: Why do we need to hardcode liquid_fraction = 1 here?

src/anelastic_thermodynamic.jl
trontrytel: Also, does this by chance assume that 1 = q_d + q_v (i.e. no liquid and ice?)

src/thermodynamic_functions.jl
```diff
@@ -1,72 +0,0 @@
-
```
trontrytel, line 1: awesome!

src/microphysics_functions.jl
```diff
@@ -139,7 +139,8 @@ function microphysics_rain_src(
     phase_part = TD.PhasePartition(qt, ql, qi)
     _ret.thl = TD.liquid_ice_pottemp_given_pressure(param_set, T, p0, phase_part)
     _ret.th = TD.dry_pottemp_given_pressure(param_set, T, p0, phase_part)
-    _ret.rho = rho_c(p0, T, qt, _ret.qv)
+    # added the original rho_c here until thermodynamics is swaped
+    _ret.rho = p0 / ((Rd * T) * (1.0 - qt + eps_vi * _ret.qv))
```
trontrytel, line 144: Why can't we use one of the anelastic thermodynamics functions here? We should not have to calculate density differently here

trontrytel (approved):
I left some questions. Might be good to chat?

(didn't mean to approve yet 😂 )

src/anelastic_thermodynamic.jl
yairchn: its the current assumption in TC.jl that will need to change when we introduce ice

src/microphysics_functions.jl
yairchn: It's strange to run PhaseEquil when we already have the temperature. I figured this entire code will be switched with CloudMicrophysics.jl soon.

src/EDMF_Updrafts.jl
```diff
@@ -342,23 +346,22 @@ function buoyancy(
...
                         qt = UpdVar.QT.values[i, k - 1]
                         h = UpdVar.H.values[i, k - 1]
-                        sa = eos(param_set, self.Ref.p0_half[k], qt, h)
-                        qv = qt - sa.ql
-                        t = sa.T
-                        rho = rho_c(self.Ref.p0_half[k], t, qt, qv)
+                        ts = AnelasticPhaseEquil_pθq(param_set, self.Ref.p0_half[k], h, qt)
+                        t = TD.air_temperature(ts)
+                        # qv = qt - TD.liquid_specific_humidity(ts)
+                        # rho = rho_c(self.Ref.p0_half[k], t, qt, qv)
```
charleskawczynski, line 362: Let's delete this

src/TurbulenceConvection.jl
```diff
@@ -69,7 +69,7 @@ include("NetCDFIO.jl")
 include("ReferenceState.jl")
 include("types.jl")
 
-include("thermodynamic_functions.jl")
+include("anelastic_thermodynamic.jl")
```
charleskawczynski, line 72:
```suggestion
include("anelastic_thermodynamics.jl")
```
And change filename accordingly

src/microphysics_functions.jl
charleskawczynski: Since `qt,ql,qi` are known (they're passed in), we should really change this to pass in a thermodynamic state and just call thermo methods.

charleskawczynski (commented):
Changes look great AFAICT, let's not forget to delete the `eos_struct`

```julia
Base.@kwdef mutable struct eos_struct
    T::Float64 = 0
    ql::Float64 = 0
end
```
in `types.jl`

yairchn:
We are waiting for a corresponding PR https://github.com/CliMA/Thermodynamics.jl/pull/52 before we can run this

integration_tests/utils/main.jl
```diff
@@ -4,6 +4,12 @@ import TurbulenceConvection
 import ClimaCore
 const CC = ClimaCore
 
+#------ TODO: Remove this when branch is ready to be merged
+import Pkg
+Pkg.rm("Thermodynamics")
```
charleskawczynski, line 9: Hmm, maybe `Pkg.rm("Thermodynamics")` just needs to be deleted? I don't see a `force` option (to not fail if the package doesn't exist or something).

yairchn: If its too complicated we can just keep this PR updated until we merge the Changes in Thermodynamics.jl

charleskawczynski: Hmm, it seems to be working, now?

charleskawczynski: Ah, only for Bomex. Hmm, yeah. Let's remove this line, I think that should do it.

yairchn (commented):
great catch

yairchn:
See https://github.com/CliMA/TurbulenceConvection.jl/pull/325

PR #268, opened 2021-09-16
Add function for handling aliased names

charleskawczynski:
This PR replaces the variable maps, which were used to allow us to compare with other datasets with potentially different names.

The main new piece in this PR is the `name_aliases` function, which returns a `Dict` whose keys are the "core variable name", which we'll hopefully use for the property names of our structs. The values are all of the possible aliases that we may need for comparing with other simulations (e.g., pycles, scampy, etc.). These aliases should refer to the same variable, but with a different name.

This doesn't really help us right now but this will help/allow us to loop over property names and significantly trim down the size of our io/initialize_io functions. For example,
```julia
write_profile(Stats, "mixing_length_ratio", self.ml_ratio)
```
We might want to add a dict entry `"ml_ratio" => ("mixing_length_ratio", )`.

Side note: because of these changes, it seems simplest to remove the LES variable name in the mse tables.


src/name_aliases.jl
```diff
@@ -0,0 +1,34 @@
...
+
+function get_nc_data(ds, var::String)
+    dict = name_aliases()
+    key_options = haskey(dict, var) ? (var, dict[var]...) : (var,)
+
+    for key in key_options
+        if haskey(ds, key)
+            return Array(ds[key])
+        else
+            for group_option in ["profiles", "reference", "timeseries"]
```
haakon-e, line 25: Seems like you added `timeseries` despite this not being part of the old function?

src/name_aliases.jl
```diff
@@ -0,0 +1,34 @@
+function name_aliases()
```
haakon-e, line 1: This function (or `get_nc_data` could do well with a one-sentence documentation), depending on where it is put, it could look something like, 

> any aliases for a variable name are stored as tuples in the dictionary. `get_nc_data` checks all these aliases when retrieving a data given a TC.jl variable (keys in this dictionary)

src/name_aliases.jl
haakon-e: Is it currently the case that variable names in `reference` always form mutually exclusive sets from variable names in `timeseries`?

haakon-e (commented):
Overall, looks good. Since these are functions that may be useful to external users (if you want to use TC.jl to fetch data from nc-files), it may be useful with some minimal documentation about these methods.

src/name_aliases.jl
charleskawczynski: Oh, yeah, I had thought about making this compatible with `get_time(ds, var)` in `compute_mse.jl`.

I'm not sure about mutually exclusive part--I'd have to sift through the nc files.

src/name_aliases.jl
charleskawczynski: Agreed.

src/Forcing.jl
```diff
@@ -21,16 +21,16 @@ function initialize(self::ForcingBase{ForcingLES}, GMV::GridMeanVariables, Gr::G
...
-        self.dtdt_hadv = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "dtdt_hadv", imin, imax))
-        self.dtdt_nudge = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "dtdt_nudge", imin, imax))
-        self.dtdt_fluc = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "dtdt_fluc", imin, imax))
-        self.dqtdt_hadv = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "dqtdt_hadv", imin, imax))
-        self.dqtdt_nudge = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "dqtdt_nudge", imin, imax))
-        self.dqtdt_fluc = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "dqtdt_fluc", imin, imax))
-        self.subsidence = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "ls_subsidence", imin, imax))
-        self.u_nudge = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "u_mean", imin, imax))
-        self.v_nudge = pyinterp(Gr.zc, z_les_half, get_nc_data(data, "profiles", "v_mean", imin, imax))
+        self.dtdt_hadv = pyinterp(Gr.zc, zc_les, mean_nc_data(data, "profiles", "dtdt_hadv", imin, imax))
```
yairchn, line 26: @costachris can you take a look at this change?

charleskawczynski: This should be purely refactoring, so we should be fine.

costachris (approved):

PR #273, opened 2021-09-16
Split out more tendencies from update

charleskawczynski:
This PR
 - Splits out the construction of the implicit operator from the update of the 2nd order moment tdma solves
 - Moves all implicit operators (construction of tridiagonal matrices) into `Operators.jl`, since they will (hopefully) be eventually replaced with a fully implicit solve.


src/Operators.jl
```diff
@@ -517,3 +517,123 @@ function dual_centers(f::AbstractMatrix, grid, k::Int, i_up::Int)
...
     end
 end
+
+
+#####
+##### Implicit operators
+#####
+
+
+function construct_tridiag_diffusion_O1(grid::Grid, dt, ρ_ae_K_m, ρ_0, ae)
```
yairchn, line 527: maybe we can come up with a more informative name?

yairchn (approved):
LGTM

src/Operators.jl
charleskawczynski: I welcome better name suggestions!

yairchn (approved):
LGTM

PR #274, opened 2021-09-16
Centralize computation of GVM 2nd 3rd moments

yairchn:
This PR make order and when how and where second and Thord moments in the grid are computed


src/Turbulence_PrognosticTKE.jl
```diff
@@ -725,76 +717,76 @@ function get_GMV_CoVar(
...
-                covar_e.values[k] = covar_e.values[k] / ae[k]
-            else
-                covar_e.values[k] = 0.0
-            end
-        end
-    end
-
-    return
-end
+# function get_env_covar_from_GMV(
```
charleskawczynski, line 720: Let's just delete this.

yairchn:
Bors r+

PR #276, opened 2021-09-16
cleanup quadrature

yairchn:
these two variables are computed in quadrature but are not used


charleskawczynski (approved):
LGTM, let's just squash, first.

yairchn:
Bors r+

PR #279, opened 2021-09-17
Use generic methods in Cases.jl

charleskawczynski:
I didn't think that this would help us delete this much code, but 🎉!


charleskawczynski:
I think that this is going to help with some subsequent work I'd like to tackle next.

PR #282, opened 2021-09-17
add compute_diffusive_fluxes and place it with aux updates in the mas…

yairchn:
…ter update function


charleskawczynski (approved):
LGTM. Hopefully we can combine some of these functions down the road so that we can fuse more grid traversals.

yairchn:
Bors r+

yairchn:
Bors r+

PR #283, opened 2021-09-18
add compose_tridiagonal_terms function

yairchn:


src/Turbulence_PrognosticTKE.jl
```diff
@@ -397,6 +397,61 @@ function update(self::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBas
...
+        GMV.QT.tendencies[k] += (GMV.QT.new[k] - GMV.QT.values[k]) * TS.dti
+        GMV.U.tendencies[k] += (GMV.U.new[k] - GMV.U.values[k]) * TS.dti
+        GMV.V.tendencies[k] += (GMV.V.new[k] - GMV.V.values[k]) * TS.dti
+    end
+
+    return
+end
+
+
+function compose_tridiagonal_terms(self::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBase, TS::TimeStepping)
```
charleskawczynski, line 449: Can we maybe call this `compose_time_implicit_terms`, or `configure_time_implicit_eq` (Since it also constructs `b` in `Ax=b`)?

charleskawczynski (commented):
These changes look fine to me— I think we can find a better function name, though. I left some suggestions, but I’m open to other names.

yairchn:
Bors r+

PR #287, opened 2021-09-20
Refactor Cases

charleskawczynski:
This PR refactors Cases.jl by:
 - Removing unnecessary `theta` fields
 - Using local variables to reduce `getproperty` calls
 - Fuses some loops

So far, this branch only applies this pattern to a small set of Cases.jl, but it'd be nice if someone could take this over and apply to the rest of the file.

I know that this may interfere with the thermodynamics PR (one reason why I didn't finish), so maybe we can wait until that merges, first, before continuing with this.

### DIFF
--- a/integration_tests/utils/Cases.jl
+++ b/integration_tests/utils/Cases.jl
@@ -203,18 +203,19 @@ function initialize_surface(self::CasesBase{Soares}, grid::Grid, ref_state::Refe
     self.Sur.qsurface = 5.0e-3
     theta_flux = 6.0e-2
     qt_flux = 2.5e-5
+    ρ0_surf = ref_state.rho0[kf_surf]
     ts = TD.PhaseEquil_pTq(param_set, ref_state.p0[kf_surf], self.Sur.Tsurface, self.Sur.qsurface)
-    theta_surface = TD.liquid_ice_pottemp(ts)
-    self.Sur.lhf = qt_flux * TC.surface_value(ref_state.rho0, grid) * TD.latent_heat_vapor(ts)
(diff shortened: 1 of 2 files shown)


integration_tests/utils/Cases.jl
```diff
@@ -403,32 +389,36 @@ function initialize_forcing(self::CasesBase{Bomex}, grid::Grid, ref_state::Refer
...
+        if zc_k <= 300.0
+            dqtdt = -1.2e-8   #kg/(kg * s)
+        elseif zc_k > 300.0 && zc_k <= 500.0
+            dqtdt = -1.2e-8 + (zc_k - 300.0) * (0.0 - -1.2e-8) / (500.0 - 300.0) #kg/(kg * s)
+        else
+            dqtdt = 0.0
         end
+        self.Fo.dqtdt[k] = dqtdt
 
         #Set large scale subsidence
```
yairchn, line 414: subsidence = 0.0

charleskawczynski: I'm not sure I understand this comment

charleskawczynski: Ah, by default, yes.

yairchn: above 'zc_k <= 2100.0' --> subsidence = 0.0

charleskawczynski:
I don't like that these changes are making this file even larger. In the long term, when more loops are fused, this packing/unpacking pattern will probably pay off, but I'm not sure it's paying off with the limited scope at the moment. I may try to peel off pieces of this PR that seem less iffy

charleskawczynski:
Superseded by #302

PR #290, opened 2021-09-22
aj/use cm

trontrytel:
First try at using CloudMicrophysics.jl in TC.jl. 

Lets see what the CI has to say


src/EDMF_Rain.jl
```diff
@@ -171,29 +171,24 @@ function solve_rain_evap(
     @inbounds for k in real_center_indices(self.grid)
         flag_evaporate_all = false
 
-        tmp_evap = max(
-            0,
-            conv_q_rai_to_q_vap(
+        q = TD.PhasePartition(GMV.QT.values[k], GMV.QL.values[k], 0.0)
+
+        tmp_evap_rate =
+            -CM1.evaporation_sublimation(
```
yairchn, line 177: why is there a '-' sign here?
to the best of my understanding both 'evaporation_sublimation' and 'conv_q_rai_to_q_vap' return a positive value - no?

trontrytel: I always confuse myself there. So lets double check together:

1. In EDMF/Rain I want to compute the qt due to evaporation tendency:
https://github.com/CliMA/TurbulenceConvection.jl/blob/da88225d5e0de203772fac96f6b996cc4c8b1172/src/EDMF_Rain.jl#L191

2. The CloudMicrophysics.jl returns the rain tendency due to evaporation
https://github.com/CliMA/CloudMicrophysics.jl/blob/169c19f2765f7ccdf08bf7fbbdbbda8d67dffda9/src/Microphysics_1M.jl#L702

So I think that the signs are correct there?

src/TurbulenceConvection.jl
```diff
@@ -25,6 +26,14 @@ const SA = StaticArrays
 const TD = Thermodynamics
 const RS = RootSolvers
 
+const CM = CloudMicrophysics
+const CM0 = CloudMicrophysics.Microphysics_0M
+const CM1 = CloudMicrophysics.Microphysics_1M
+const liq_type = CM1.LiquidType()
+const ice_type = CM1.IceType()
+const rai_type = CM1.RainType()
```
yairchn, line 34: 'rain_type' ?

trontrytel: I need the types for dispatch over parameters and microphysics functions. And I'm obsessive about line alingment. So all the type names shall have the same length ;)

PR #292, opened 2021-09-22
move mf_tend and nph_tend to GMV.tendencies

yairchn:
prepare to handle the tridiagonal solvers in the grid mean


src/Turbulence_PrognosticTKE.jl
```diff
@@ -246,13 +246,15 @@ function update_cloud_frac(self::EDMF_PrognosticTKE, GMV::GridMeanVariables)
     GMV.cloud_cover = min(self.EnvVar.cloud_cover + sum(self.UpdVar.cloud_cover), 1)
 end
 
-function compute_gm_tendencies!(grid, Case, GMV, ref_state, TS)
-    param_set = parameter_set(GMV)
+function zero_gm_tendencies!(self::EDMF_PrognosticTKE)
```
charleskawczynski, line 249: Let's keep these assignments in the `compute_gm_tendencies!`.

src/Turbulence_PrognosticTKE.jl
```diff
@@ -368,6 +376,7 @@ function update(self::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBas
     # Some of these methods should probably live in `compute_tendencies`, when written, but we'll
     # treat them as auxiliary variables for now, until we disentangle the tendency computations.
     set_old_with_values(self.UpdVar)
+    zero_gm_tendencies!(self)
```
charleskawczynski, line 379: Why would we call this, here?

charleskawczynski: Let's remove it.

yairchn: its called here as the tendencies are updated in 'compute_GMV_MF' so if the first thing in 'compute_gm_tendencies' will zero out tendencies these will be removed

charleskawczynski: All tendencies should be computed in the same function--ideally the same loop

charleskawczynski: If that's the case, then let's move the contents of `compute_GMV_MF` into `compute_gm_tendencies!`.

src/Turbulence_PrognosticTKE.jl
```diff
@@ -309,15 +312,69 @@ function compute_gm_tendencies!(grid, Case, GMV, ref_state, TS)
...
+        m_bcs = (; bottom = SetValue(0), top = SetValue(0))
+        h_en_f = interpc2f(self.EnvVar.H.values, grid, k; m_bcs...)
+        qt_en_f = interpc2f(self.EnvVar.QT.values, grid, k; m_bcs...)
+        @inbounds for i in xrange(self.n_updrafts)
+            h_up_f = interpc2f(self.UpdVar.H.values, grid, k, i; m_bcs...)
+            qt_up_f = interpc2f(self.UpdVar.QT.values, grid, k, i; m_bcs...)
+            self.massflux_h[k] += self.m[i, k] * (h_up_f - h_en_f)
+            self.massflux_qt[k] += self.m[i, k] * (qt_up_f - qt_en_f)
+        end
+    end
```
charleskawczynski, line 352: This is non-blocking, but we could fuse two loops by moving this block above the first `@inbounds for k in real_center_indices(grid)` in this function, and then move the contents of the loop below into the loop above that traverses cell centers. We can always do that in a later PR, though.

charleskawczynski (approved):
These changes look fantastic. Thanks for connecting the dots, here!

yairchn:
Bors r+

yairchn:
Bors r+

yairchn:
Bors r+

PR #293, opened 2021-09-22
Remove constants, move to CLIMAParameters

charleskawczynski:
This PR removes some of the constants and moves them into calls to CLIMAParameters.jl methods. This effectively unifies parameters that were previously inconsistent.


charleskawczynski:
Results look fine

yairchn (approved):
LGTM

PR #294, opened 2021-09-22
Fix contour plot time for tc main

charleskawczynski:
Thanks @trontrytel, for finding this!


charleskawczynski:
We might want to instead used the `_mapped` data, so that everything is compared on the same playing field-- except, it is nice to plot the raw data (rather than interpolated data). I think these changes are fine for now since we don't change the temporal bounds very much.

PR #297, opened 2021-09-22
compute exner function from thermodynamic state

yairchn:
On main, currently we enforce a PhasePartition(0,0,0) in the computation of exner function using the reference pressure p0. This was added as I thought that it is required in order to get the buoyancy gradients right, as using 'TD.air_pressure(ts) ' resulted in behavioral changes. It seems that we can use the full PhasePartition(qt,ql,qi=0) in the Exner computation in buoyancy gradients - the problem with the computation of TD.air_pressure(ts) is that it is not Anelastic. Once we have a full Anelastic ThermoState that can produce exner we should switch to using TD.exner(ts).


ilopezgp:
@yairchn Are we not using anelastic thermodynamics now? What are we using?

charleskawczynski (approved):
I have no strong opinions, here. Is this improving the model somehow? And why are we making these changes? I think the PR description could be improved (what the PR does and why).

yairchn:
> @yairchn Are we not using anelastic thermodynamics now? What are we using?

here we are using 'TD.PhasePartition' which is not exact as it 

> @yairchn Are we not using anelastic thermodynamics now? What are we using?

to compute exner we go directly to 'TD.exner_given_pressure' which ensure that the right pressure is used. our hope is to have a 'ts = PhaseEquil...()' that is ensuring the correct pressure and avoid calling other internal Thermodynamics.jl methods in the future. 

> @yairchn Are we not using anelastic thermodynamics now? What are we using?

yairchn:
> I have no strong opinions, here. Is this improving the model somehow? And why are we making these changes? I think the PR description could be improved (what the PR does and why).

see the updated text above, basically the code is main is not fully correct, a more correct version is in this PR, pending on an Anelastic ThermoState.

ilopezgp:
I went through the PR and here are my thoughts:

1. The Exner function is only used in the current formulation to construct the `prefactor`, which we should rename to `∂b∂θv`. This is equal to `∂b∂θv=g/θv_0`, where we should make sure we are using the `θv_0` that corresponds to our general definition of buoyancy. If we use `alpha0` as the specific volume that defines all subgrid-scale buoyancies, then we should the reference state `θv` as our `θv_0`. This means that in this case, the correct thing is to use `TD.PhasePartition` with the reference phase partition, which I guess is all dry. This means the master version is correct.
2. The Exner function is not used to compute the cloudy contribution to the buoyancy flux. Note that this is because we have used `Π=t_cloudy/θ_cloudy` instead, so we do not need to compute the Exner function here. We should make sure that `t_cloudy` and `θ_cloudy` are computed correctly, though.

So to sum up, these changes as they are proposed are not correct. The former version is correct.

ilopezgp (changes requested):
See my comments above, I suggest closing this PR.

PR #299, opened 2021-09-22
Remove more constant parameters and use climaparameters

charleskawczynski:


charleskawczynski:
Only had to change the Rico mse table, results look fine.

PR #301, opened 2021-09-23
add qr plot in Rico

trontrytel:
Trying to add qr plot to contours and profiles for Rico.


charleskawczynski:
Looks good! All you need to do is update the Mae tables, now

charleskawczynski (approved):

PR #303, opened 2021-09-23
bugfix to cutoff autoconversion

trontrytel:


yairchn (approved):
changes are minor and bring it closer to SCAMPy, we should update the tables and merge

PR #305, opened 2021-09-23
use 0-moment microphysics from CM.jl

trontrytel:
This PR replaces the instantaneous autoconversion rate with the 0-moment microphysics scheme from CloudMicrophysics.jl.

To retain the same "instanteneous" rain removal the CLIMAParameters precipitation timescale is overwritten to be equal to the model time step. 

I have also removed the cond_evap parameter from the parameter_set file. That parameter is related to condensation timescale for non-equilibrium cloud formation. Right now TC.jl is formulated assuming equilibrium (saturation adjustment). So that parameter should not be used anywhere.


integration_tests/utils/parameter_set.jl
```diff
@@ -45,7 +46,7 @@ function create_parameter_set(namelist)
         R_d = 287.1,
         R_v = 461.5,
         molmass_ratio = 461.5/287.1,
-        τ_cond_evap = TC.parse_namelist(namelist, "microphysics", "τ_cond_evap"; default = 10.0),
+        τ_precip = TC.parse_namelist(namelist, "microphysics", "τ_precip"; default = 1000.0),
```
yairchn, line 49: it's a large difference in 'τ'. Does it has the same meaning? was the default ever used?

yairchn (approved):
LGTM - tables needs update but TRMM looks great!

integration_tests/utils/parameter_set.jl
trontrytel: Before in TC.jl we were computing the delta qr and dividing by dt. We never assumed any relaxation timescale. 

In CM.jl it is coded as a relaxation to some state with a timescale given as parameter. By default the timescale is large to actually allow some precipitable water to remain in the domain. I think it was tested in some old ClimateMachine setups. But the timescale itself was never properly calibrated against anything. 

So overall to reproduce the TC-main behaviour it needs to be set to dt. For other cases it has to be calibrated anyway

trontrytel: Also, the tau_cond_evap and tau_precip are two different parameters. The first one is used with non-equilibrium cloud formation. And should not be used in TC.jl right now, cause we only assume equilibrium and run with saturation adjustment. The second one is used for rain removal timescale.

PR #306, opened 2021-09-24
use CM 1-moment microphysics

trontrytel:
This PR replaces the remaining microphysics functions with CloudMicrophysics.jl 1-moment microphysics functions.

We are evaporating too much rain. I think its the same issue as we have discussed before - i.e. how to compute the q_vap_sat when in anelastic approximation.


charleskawczynski (commented):
The code changes look great, but are we okay with the new rain profiles? The rain profile in Rico looks like it's zero everywhere

src/EDMF_Rain.jl
```diff
@@ -213,7 +208,7 @@ function solve_rain_evap(
             # TODO: assuming that rain evaporation doesn"t change
             # rain area fraction
             # (should be changed for prognostic rain area fractions)
-            QR.values[k] -= tmp_evap
+            QR.values[k] -= tmp_evap_rate * Δt
```
charleskawczynski, line 211: Are we time-stepping `QR`, here? If so, can we write this as `QR.values[k] += (-tmp_evap_rate) * Δt` so it's clear that `(-tmp_evap_rate)` is the (only) tendency?

src/EDMF_Rain.jl
```diff
@@ -171,29 +171,24 @@ function solve_rain_evap(
...
-            ) * Δt,
-        )
+                GMV.T.values[k],
+            )
 
-        if tmp_evap > QR.values[k]
+        if tmp_evap_rate * Δt > QR.values[k]
             flag_evaporate_all = true
-            tmp_evap = QR.values[k]
+            tmp_evap_rate = QR.values[k] / Δt
```
charleskawczynski, line 188: Is there a timescale that we can divide by, rather than the timestep?

charleskawczynski: Or is this just for CFL-type calculations?

trontrytel:
> The code changes look great, but are we okay with the new rain profiles? The rain profile in Rico looks like it's zero everywhere

I'm still debugging this. The autoconverion, accretion and advection are ok. The problem is with rain evaporation. Maybe it's the q_vap_sat calculation again, or maybe something else. Will let you know

src/EDMF_Rain.jl
trontrytel: In this PR I wanted to keep the "logic" that was there. In the next one we should change it to only return the tendency and remove all that

src/microphysics_functions.jl
```diff
@@ -173,35 +85,5 @@ function rain_area(source_area, source_qr, current_area, current_qr)
         _ret.qr = current_qr + source_area * source_qr
```
yairchn, line 85: Is this an update of qr? if so the function should return the src and the update should be done elsewhere with all prognostic variable

src/EDMF_Rain.jl
trontrytel: Yeah - those are just safeguards to decrease the evaporation rate if we would evaporate more that what we have. Should be removed and handled better. But I would prefer to do it in the next PR and focus on just replacing the functions in this one

src/microphysics_functions.jl
trontrytel: This will be cleaned up in the next PR. In this one I just want to replace the functions we are already using and not change anything else

trontrytel:
Just to clarify: The goal of this PR is to replace 4 functions: autoconversion, accretion, terminal velocity and rain_evaporation with their equivalents from CloudMicrophysics.jl I'm done with the first 3.

In the next PR I will remove the leftover code related to rain area fraction, and delete the leftower new/old stuff for rain. But since those can bring behavioural changes I want to do them after this PR

charleskawczynski (approved):

ilopezgp:
> Just to clarify: The goal of this PR is to replace 4 functions: autoconversion, accretion, terminal velocity and rain_evaporation with their equivalents from CloudMicrophysics.jl I'm done with the first 3.
> 
> In the next PR I will remove the leftover code related to rain area fraction, and delete the leftower new/old stuff for rain. But **since those can bring behavioural changes** I want to do them after this PR

I take that you want this PR to not incur any behavioral changes? In that case, I suggest making it a draft PR until we get Rico with rain.

trontrytel:
> > Just to clarify: The goal of this PR is to replace 4 functions: autoconversion, accretion, terminal velocity and rain_evaporation with their equivalents from CloudMicrophysics.jl I'm done with the first 3.
> > In the next PR I will remove the leftover code related to rain area fraction, and delete the leftower new/old stuff for rain. But **since those can bring behavioural changes** I want to do them after this PR
> 
> I take that you want this PR to not incur any behavioral changes? In that case, I suggest making it a draft PR until we get Rico with rain.

Yeh - I'm not gonna merge it untill I figure out the rain evaporation

charleskawczynski:
Alternatively, you could break off parts of this PR into a separate one (e.g., the part that migrates the CLIMAParameters 🙂)

trontrytel:
An update on rain debugging front. The rain fall velocity, autoconversion and accretion are fine. I now think there is a mistake in the TC-main function for rain evaporation. 

In the `EDMF_rain.jl` and `microphysics_functions.jl` I now have both formulations coded.  When I print the actual rates for the both versions, the old version (the one from TC-main) is non-zero only 3 times during the entire simulation. That is not right. Also, in documentation build for CM.jl we compare the CM.jl evaporation rate against an empirical parameterization and they are close.

I tried figuring out where is the difference coming from. Unfortunately the TC-main version is coded using different notation than the one we have right now in CM.jl. I'll post an image below with my comparisons of both equations. The end result is that the coefficients multiplying the ventilation factor are different in the two formulations. And again, I think that what we have in CM.jl is correct. 

It's worth noting that in TC-main, our autoconversion timescale was set to be 10x slower than the defaults from CM.jl. So making the autoconversion process faster gives you more precipitation. I can't set it to be exactly the default value from CM.jl I start running into too fast terminal velocities for the given timestep length. And I don't want to change the timestep in this PR so that we keep comparing apples to apples between TC-main and TC-PR. 

What is currently running in buildkite is the CM.jl version of rain evaporation plus autoconversion timescale changed to be very close to the default number. We get some rain, not as much as before. But I think with proper calibration, even just taking the dt and autoconversion timescale we should be able to get the qr profiles we want. 

So let me know if the equations and plots convince you, or if you want to check some more equations by yourself.

trontrytel:
CM_1_moment

TC_main

src/EDMF_Rain.jl
```diff
@@ -115,7 +115,7 @@ function solve_rain_fall(
     # TODO: assuming GMV.W = 0
     # TODO: verify translation
     @inbounds for k in real_center_indices(grid)
-        term_vel[k] = terminal_velocity(param_set, Rain.C_drag, Rain.MP_n_0, QR.values[k], self.ref_state.rho0_half[k])
+        term_vel[k] = CM1.terminal_velocity(param_set, rai_type, self.ref_state.rho0_half[k], QR.values[k])
```
ilopezgp, line 118: is `rai_type` the name we are set with?

src/EDMF_Rain.jl
```diff
@@ -171,29 +171,44 @@ function solve_rain_evap(
...
+        tmp_evap_rate = tmp_evap_rate_clima
+
+        if tmp_evap_rate * Δt > QR.values[k]
             flag_evaporate_all = true
-            tmp_evap = QR.values[k]
+            tmp_evap_rate = QR.values[k] / Δt
         end
 
-        self.rain_evap_source_qt[k] = tmp_evap * RainArea.values[k]
+        self.rain_evap_source_qt[k] = tmp_evap_rate * RainArea.values[k]
```
ilopezgp, line 211: So the units were wrong in master?

src/microphysics_functions.jl
```diff
@@ -1,70 +1,38 @@
-
 """
-Source term for thetal because of qr transitioning between the working fluid and rain
-(simple version to avoid exponents)
+It computes the tendency term to thetal because of qr transitioning between
+the working fluid and rain
```
ilopezgp, line 3: can you add directionality? Instead of between, from... to, so we understand the sign better.

src/microphysics_functions.jl
```diff
@@ -1,70 +1,38 @@
...
+    D_vapor = CP.Atmos.Microphysics.D_vapor(param_set)
+    ρ_cloud_liq = CP.Planet.ρ_cloud_liq(param_set)
+    K_therm = CP.Atmos.Microphysics.K_therm(param_set)
+    N_Sc = ν_air / D_vapor
 
-    av_param = sqrt(2.0 * π) * a_vent * sqrt(rho / rho_cloud_liq)
-    bv_param =
-        2^(7 / 16) * gamma_11_4 * π^(5 / 16) * b_vent * N_Sc^(1 / 3) * sqrt(v_c) * (rho / rho_cloud_liq)^(11 / 16)
+    av_param = sqrt(2.0 * π) * a_vent * sqrt(rho / ρ_cloud_liq)
+    bv_param = 2^(7 / 16) * gamma_11_4 * π^(5 / 16) * b_vent * N_Sc^(1 / 3) * sqrt(v_c) * (rho / ρ_cloud_liq)^(11 / 16)
```
ilopezgp, line 35: Are these functional groups coded up in CloudMicrophysics? Can we directly import them from there?

src/microphysics_functions.jl
```diff
@@ -74,41 +42,70 @@ function conv_q_rai_to_q_vap(param_set, C_drag, MP_n_0, a_vent, b_vent, q_rai, q
...
+    q::TD.PhasePartition{FT},
+    q_rai::FT,
+    ρ::FT,
+    p::FT,
+    T::FT,
+) where {FT <: Real}
+    evap_subl_rate = FT(0)
+
+    #_S::FT = TD.supersaturation(param_set, q, ρ, T, TD.Liquid())
+    #q_sat::FT = TD.q_vap_saturation_generic(param_set, T, ρ, TD.Liquid())
```
ilopezgp, line 65: Remove?

src/EDMF_Rain.jl
charleskawczynski: `main`* 🙂

src/EDMF_Rain.jl
trontrytel: I define them here:
https://github.com/CliMA/TurbulenceConvection.jl/blob/277f5a5e38fe316a2a832f3890e1070b7162e8e5/src/TurbulenceConvection.jl#L31-L37

Would you like a different name?

src/EDMF_Rain.jl
trontrytel: I think so. Would be good to go through the code together and check. Maybe once we get rid of the remaining new/old leftovers

src/microphysics_functions.jl
trontrytel: It is the same term for autoconversion (qt -> qr) and rain evaporation (qr -> qt). The sign of `dqr_dt` will make the difference.

src/microphysics_functions.jl
trontrytel: cleaned up

src/microphysics_functions.jl
trontrytel: cleaned up

src/EDMF_Rain.jl
ilopezgp: If you want to stick to three letters, that's fine with me. I would prefer rain_type and snow_type.

trontrytel: I'm a little obsessive with line alignment. But I probably should not. Will switch to full name for the precipitation ones

ilopezgp (approved):

PR #308, opened 2021-09-24
Refactor Cases.jl

charleskawczynski:
I noticed that `initialize_profiles` doesn't actually need `CasesBase` anywhere, it only needs the particular case (`<: AbstractCaseType`). This PR passes in the case, instead. We also add `case` as a property to `CasesBase`.


charleskawczynski:
Ah, my grepping failed me. `LES_driven_SCM` is an exception. bors r-

charleskawczynski:
Going to close this for now..

PR #309, opened 2021-09-24
Try to improve ci build time

charleskawczynski:


jakebolewski:
This biggest band for your buck would be to add a fixed JULIA_DEPOT_PATH env variable  to the toplevel buildkite pipeline, that would prevent downloading and precompiling every ci run: https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH

jakebolewski:
We currently set this to: export `JULIA_DEPOT_PATH="${CI_BUILD_DIR}/depot/{cpu/gpu}"` where CI_BUILD_DIR is 

    export CI_BUILD_DIR="${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/${BUILDKITE_BUILD_NUMBER}"

charleskawczynski:
> This biggest bang for your buck would be to add a fixed JULIA_DEPOT_PATH env variable to the toplevel buildkite pipeline, that would prevent downloading and precompiling every ci run: https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH

Yeah, I was just reading through this

> We currently set this to: `export JULIA_DEPOT_PATH="${CI_BUILD_DIR}/depot/{cpu/gpu}"` where CI_BUILD_DIR is
> 
> ```
> export CI_BUILD_DIR="${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/${BUILDKITE_BUILD_NUMBER}"
> ```

Which repo do we set this in? ClimaCore? ClimateMachine? The only mention of `JULIA_DEPOT_PATH` I see is in ClimateMachine.jl

charleskawczynski:
I just updated things, but still not sure if this is right..

jakebolewski:
you want to remove `BUILDKITE_BUILD_NUMBER `as that will be unique to every build, I created a directory here:

`/central/scratch/climaci/turbulenceconvection-ci/depot` that you could re-use

`BUILDKITE_BUILD_PATH` is set to `/central/scratch/climaci` in the config

jakebolewski:
These settings are part of the hooks to set the environment for the buildkite runner (where we load modules, etc.)

jakebolewski:
Ping me if this doesn't work as we might have to adjust how we set some of these env variables in the hook (only define them if they are not specified).

charleskawczynski:
> you want to remove `BUILDKITE_BUILD_NUMBER `as that will be unique to every build

Right, makes sense

> I created a directory here:
> 
> `/central/scratch/climaci/turbulenceconvection-ci/depot` that you could re-use
> 
> `BUILDKITE_BUILD_PATH` is set to `/central/scratch/climaci` in the config
> `turbulenceconvection-ci` is the pipeline slug

Ok. If this results in caching pkg-related chunks, will it auto-update when new deps are added?

jakebolewski:
Yeah it's not overriding it hook:
https://buildkite.com/clima/turbulenceconvection-ci/builds/3186#96fd31a7-a3c7-4382-97fb-9a994c505f11/95-101

Let me make an adjustment

jakebolewski:
ok let's try again

.buildkite/pipeline.yml
```diff
@@ -6,13 +6,16 @@ env:
   CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED: "true"
   BUILDKITE_COMMIT: "${BUILDKITE_COMMIT}"
   BUILDKITE_BRANCH: "${BUILDKITE_BRANCH}"
+  CI_BUILD_DIR: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}"
+  JULIA_DEPOT_PATH: "${CI_BUILD_DIR}/depot/cpu"
```
jakebolewski, line 10: let's only set this variable to not clash with the above variable (which is used everywhere in the runner, the later is julia specific)

jakebolewski:
Yes it will update

.buildkite/pipeline.yml
charleskawczynski: I'm not sure I follow your suggestion, here. Should we
 - [ ] just remove `CI_BUILD_DIR: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}"`?
 - [ ] Remove `CI_BUILD_DIR: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}"` and set `JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/cpu"`?

jakebolewski: the second one

charleskawczynski:
Hmm, I'm a bit confused why we would get this error in the `print_new_mse.jl` job

```
ERROR: LoadError: ArgumentError: Package OrderedCollections [bac558e1-5e72-5ebc-8fee-abe8a469f55d] is required but does not seem to be installed:
--
  | - Run `Pkg.instantiate()` to install all recorded dependencies.
```

OrderedCollections is a dep in both the Project.toml and test/Project.toml (we run `print_new_mse.jl` with the test environment).

Going to simply retry..

charleskawczynski:
The good news is that this is reducing CI time to ~15 min, which is nice.

charleskawczynski:
Maybe, since the Project environment is precompiled, we need to use that instead (not sure why the integration test jobs would work, though)

jakebolewski:
I would try that, not sure why the test env wouldn't work though unless you are mucking around with the load path.

jakebolewski:
Ah you are no longer instantiating the env, you have different compat entries in the test project vs the pkg project which might be the cause of the resolution issue.

charleskawczynski:
> Ah you are no longer instantiating the test env, you have different compat entries in the test project vs the pkg project which might be the cause of the resolution issue.

Ahh, yes. Makes sense

PR #312, opened 2021-09-25
remove rain area fractions

trontrytel:
First of my cleanup steps. 

This removes the environmental and updraft rain and rain area fractions. They were there mostly for debugging, but I think they bring more confusion to the code than benefits. In the current iteration, the precipitation should only be considered in the grid mean. 

I still want to come back to stratiform vs convective precipitation. But I want to do it when we have the grid mean precipitation coupling figured out.


yairchn (approved):
LGTM, I am assuming that the behavioral change is due to the change in 'τ_acnv', anything else was non behavioral changing?

charleskawczynski (approved):

ilopezgp:
Same question as @yairchn about the changes, it would be good to make sure they are due only to the change in tau. Otherwise LGTM.

ilopezgp (approved):

trontrytel:
It's a little more than that. In the previous version updrafts were interacting with rain created in the updrafts and environment was interacting only with the rain created in the environment when considering autoconversion and accretion. But the whole partitioning for stratiform and convective rain was never done fully. So we have decided for now to just do the precipitation on the grid mean everywhere. And come back to partitioned rain later

ilopezgp:
Sounds good then.

PR #313, opened 2021-09-26
Clean up Cases.jl, call satadjust outside of initialize_profiles

charleskawczynski:


charleskawczynski:
Changes in dycoms are miniscule.

PR #315, opened 2021-09-27
remove old

yairchn:
ever since we unified the time steps in the model old was effectively equal to values. 
It can be removed, along with 'set_old_with_values' function and relevant type definitions.

### DIFF
--- a/src/EDMF_Updrafts.jl
+++ b/src/EDMF_Updrafts.jl
@@ -215,16 +215,6 @@ function set_new_with_values(self::UpdraftVariables)
     return
 end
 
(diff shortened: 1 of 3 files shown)


charleskawczynski (approved):
No way! If this is possible, then let's absolutely do this. It may be that this results in a slightly different time discretization with the values/old area fraction, but if it's not crippling us, then I think this is still better because all our timestepping will be eventually revamped.

yairchn:
> ing will be eventually revampe

Just a realization I had will dealing with the tridiagonal solvers. Results are unchanged on my machine but we can wait for the buildkite

yairchn:
Bors r+

PR #316, opened 2021-09-27
Compute GMV diffusive tendencies implicitly but remove update_GMV_ED

yairchn:
An intermediated step between solving everything explicitly


src/Variables.jl
```diff
@@ -1,14 +1,3 @@
-function update(self::GridMeanVariables, TS::TimeStepping)
```
yairchn, line 1: I moved this to Turbulence_PrognosticTKE.jl under the name 'update_GMV'

yairchn (commented):
This is a draft PR where I was able to organize the GMV equations while still computing the diffusive flux in the grid mean implicitly (inside 'compute_gm_tendencies'). To compute this source explicitly making the grid mean equations explicit would require making sure that all diffusive computations are correctly translated from the implicit equation

PR #317, opened 2021-09-27
Refactor Surface.jl

charleskawczynski:
I know that this adds a few lines of code, but it _really_ reduces the number of `getproperty`/`getindex` calls, which I think will help a lot as we're revamping data structures.

This PR also throws an error after the warnings are printed--so that we're not suddenly flooding the logs like we did before.


charleskawczynski:
Results look the same.

PR #322, opened 2021-09-27
move thermodynamic eos  to Upd buoyancy function 

yairchn:
This is PR sets the ground to remove UpdVar.H.new (for all ups variables). However the remove of new will hav to come in a second PR that reorganizes the updraft equations solver due to the use of both values and new in the RHS in both scalar and w equations


src/EDMF_Updrafts.jl
```diff
@@ -317,38 +317,38 @@ function buoyancy(
...
+                    rho = TCTD.air_density(ts)
                     UpdVar.B.values[i, k] = buoyancy_c(param_set, self.ref_state.rho0_half[k], rho)
-                    ts = TD.PhaseEquil_pθq(param_set, self.ref_state.p0_half[k], h, qt)
+                    # this is here just for RH
+                    ts = TD.PhaseEquil_pθq(
+                        param_set,
+                        self.ref_state.p0_half[k],
+                        UpdVar.H.values[i, k],
+                        UpdVar.QT.values[i, k],
+                    )
```
charleskawczynski, line 350: Can we just use the same thermo state?

charleskawczynski (commented):
These changes look good to me overall, but it would be nice if we could use the same thermodynamic state for the relative humidity.

Also, can we just delete the `!extrap` branch and remove the associated bool? It doesn't seem to be used anywhere-- will we ever need it?

src/EDMF_Updrafts.jl
```diff
@@ -317,38 +317,38 @@ function buoyancy(
         @inbounds for i in xrange(self.n_updraft)
             @inbounds for k in real_center_indices(grid)
                 if UpdVar.Area.values[i, k] > 0.0
+                    # saturation adjustment
+                    ts = TCTD.eos(param_set, self.ref_state.p0_half[k], UpdVar.H.values[i, k], UpdVar.QT.values[i, k])
+                    UpdVar.QL.values[i, k] = TCTD.liquid_specific_humidity(ts)
+                    UpdVar.T.values[i, k] = TCTD.air_temperature(ts)
                     qv = UpdVar.QT.values[i, k] - UpdVar.QL.values[i, k]
-                    rho = TCTD.rho_c(
```
charleskawczynski, line 321: Let's also delete `qv = UpdVar.QT.values[i, k] - UpdVar.QL.values[i, k]`, since it will no longer be used.

src/EDMF_Updrafts.jl
```diff
@@ -357,6 +357,7 @@ function buoyancy(
                         ts = TCTD.eos(param_set, self.ref_state.p0_half[k], h, qt)
                         rho = TCTD.air_density(ts)
                         UpdVar.B.values[i, k] = buoyancy_c(param_set, self.ref_state.rho0_half[k], rho)
+                        # this is here just for RH
```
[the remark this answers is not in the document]
yairchn: 'ts = TCTD.eos(...' has now RH

yairchn:
> extrap

about !extrap, that is the option we would like to ultimately have (after figuring out how to make dycoms work with it), so I would leave it for now

src/EDMF_Updrafts.jl
yairchn: 'ts = TCTD.eos(...' has no RH

src/EDMF_Updrafts.jl
```diff
@@ -317,38 +317,37 @@ function buoyancy(
...
+                    rho = TCTD.air_density(ts)
                     UpdVar.B.values[i, k] = buoyancy_c(param_set, self.ref_state.rho0_half[k], rho)
-                    ts = TD.PhaseEquil_pθq(param_set, self.ref_state.p0_half[k], h, qt)
+                    # this is here just for RH
+                    ts = TD.PhaseEquil_pθq(
+                        param_set,
+                        self.ref_state.p0_half[k],
+                        UpdVar.H.values[i, k],
+                        UpdVar.QT.values[i, k],
+                    )
```
charleskawczynski, line 349:
```suggestion
                    args = (param_set, self.ref_state.p0_half[k], UpdVar.H.values[i, k], UpdVar.QT.values[i, k])
                    ts = TCTD.eos(args...)

                    UpdVar.QL.values[i, k] = TCTD.liquid_specific_humidity(ts)
                    UpdVar.T.values[i, k] = TCTD.air_temperature(ts)
                    rho = TCTD.air_density(ts)
                    UpdVar.B.values[i, k] = buoyancy_c(param_set, self.ref_state.rho0_half[k], rho)
                    # this is here just for RH
                    ts = TD.PhaseEquil_pθq(args...)
```

src/EDMF_Updrafts.jl
```diff
@@ -317,38 +317,37 @@ function buoyancy(
...
+                    # saturation adjustment
+                    ts = TCTD.eos(param_set, self.ref_state.p0_half[k], UpdVar.H.values[i, k], UpdVar.QT.values[i, k])
+                    UpdVar.QL.values[i, k] = TCTD.liquid_specific_humidity(ts)
+                    UpdVar.T.values[i, k] = TCTD.air_temperature(ts)
+                    rho = TCTD.air_density(ts)
                     UpdVar.B.values[i, k] = buoyancy_c(param_set, self.ref_state.rho0_half[k], rho)
+                    UpdVar.RH.values[i, k] = TD.relative_humidity(ts)
                 else
                     UpdVar.B.values[i, k] = EnvVar.B.values[k]
+                    UpdVar.RH.values[i, k] = GMV.RH.values[k]
```
charleskawczynski, line 329: Why `GMV.RH` instead of `EnvVar.RH`?

charleskawczynski (approved):
I'm not thrilled about adding more thermo constructor calls, but this does help eliminate use of `new`, so I'm fine with merging. We need to fix/resolve the inconsistency of existing thermo calls anyway.

PR #326, opened 2021-09-30
Fuse some loops

charleskawczynski:


charleskawczynski:
This is a bit related to #316, but this PR doesn't have any behavioral changes because the order of calls is effectively the same.

yairchn:
> This is a bit related to #316, but this PR doesn't have any behavioral changes because the order of calls is effectively the same.

#316 has no behavioral changes either ....

PR #332, opened 2021-10-01
Merge methods into update_aux

charleskawczynski:
This PR moves the call to set_updraft_surface_bc(self, GMV, Case)` and `diagnose_GMV_moments(self, GMV, Case, TS)` into `update_aux!`, and moves `update_aux!` to its own file.

This will help reduce the number of aliases and function calls, so changing data structures will be a bit easier.


charleskawczynski:
The HVariance isn't looking good. Going to re-try this, it wasn't difficult but it might be nice to split this into refactor+changes that will require changes (due to re-ordering of computations).

PR #333, opened 2021-10-01
use fixed surface flux for LES driven case

costachris:
A minor change that ensures surface flux is being read from the LES file.


yairchn (approved):
Great Catch! could this be related to the resolution dependance?

charleskawczynski (commented):
Looks fine to me, sorry if I translated this wrong!

costachris:
@charleskawczynski This is what I used initially, it wasn't you!

PR #73, opened 2021-08-03
The great rename

charleskawczynski:
I figure it's not a bad idea to open this as a PR now :slightly_smiling_face:

PR #335, opened 2021-10-02
Rename some variable names

charleskawczynski:
I'm starting to wonder if #73 is the way to go. There's really so many names that need to be re-thought. In addition, it no longer really makes sense to use `self` as a name--since we're using multiple dispatch. This PR makes a few name changes, including

 - GMV -> gm
 - EnvVar -> en
 - EnvThermo -> en_thermo
 - UpdThermo -> up_thermo
 - UpdVar -> up
 - tptke -> edmf
 - self -> appropriate names listed above
 - Rain -> rain

This doesn't actually change any struct field names, which would be a bit more strict, but I'm wondering what people think about migrating names a bit more incrementally.

cc @trontrytel, @yairchn, @ilopezgp, @haakon-e, @costachris

### DIFF
--- a/src/EDMF_Environment.jl
+++ b/src/EDMF_Environment.jl
@@ -1,4 +1,4 @@
-function initialize_io(self::EnvironmentVariables, Stats::NetCDFIO_Stats)
+function initialize_io(en::EnvironmentVariables, Stats::NetCDFIO_Stats)
     add_profile(Stats, "env_w")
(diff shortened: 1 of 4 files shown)


yairchn (approved):
LGTM! great work 'self' was definitely a pythonian creation

trontrytel:
Looks good to me!

charleskawczynski (commented):
bors r+

Issue #337, opened 2021-10-02
Should we revert #332?


charleskawczynski:
I just realized that #332 actually results in bad `HVar` in DYCOMS (and likely other cases). It merged without requiring to update the mse tables because, I believe, we don't test certain second order moments. If this is important to retain then I'm fine with reverting it, but I'm also wondering if we should be testing these second order moments in simulations if it is important.

cc @yairchn, @ilopezgp, @trontrytel

ilopezgp:
Yikes! Yes, let's revert this. We should add tests for `Hvar` and `Qtvar`, those are important fields.

charleskawczynski:
Closed by #338

PR #338, opened 2021-10-02
Revert "Merge methods into update_aux"

charleskawczynski:
Reverts CliMA/TurbulenceConvection.jl#332


charleskawczynski:
Wow, this lead me to find an interesting issue with our regression tests. Let's revert and then I'll improve the mse computations (which, if originally designed better would have prevented #332 from merging).

charleskawczynski:
Closes #337.

PR #339, opened 2021-10-02
Fix bug with mse computations

charleskawczynski:
In #332, the computed mse for `Hvar_mean` turned out to be lower than the best simply because the data scale of the PR was significantly larger. We should always normalize the data by the "best" data scale, and not the PR, since it's more stable and a more accurate representation of what we want to compare against. This PR fixes this issue.

Note that the MSE tables will need to be updated, but none of the source code is being changed--so our results are actually unchanged.


charleskawczynski:
I think this leads me to another question that I'm not 100% sure about. Our current comparison precedence is:

```julia
        if have_pycles_ds # LES takes first precedence
            mse_single_var = sum((data_les_cont_mapped .- data_tcc_cont_mapped) .^ 2)
            data_scale_used = data_scale_les
        elseif have_scampy_ds # SCAMPy takes second precedence
            mse_single_var = sum((data_scm_cont_mapped .- data_tcc_cont_mapped) .^ 2)
            data_scale_used = data_scale_scm
        elseif have_tc_main # TC.jl main takes third precedence
            mse_single_var = sum((data_tcm_cont_mapped .- data_tcc_cont_mapped) .^ 2)
            data_scale_used = data_scale_tcm
        else
            error("No dataset to compute MSE")
        end
```
Now that we're getting better results in TC main compared to SCAMPy, should we swap TC main and SCAMPy? I'm inclined to say we should. The advantage of keeping using SCAMPy over TC main is that it's not a moving target. That said, comparing against TC main is effectively a traditional regression test, and the problem with leaving the comparison against SCAMPy is that we could re-introduce unwanted spikes, similar to SCAMPy, and still pass the mse tests, which seems undesirable.

cc @yairchn, @ilopezgp, @trontrytel, @haakon-e.

charleskawczynski:
I don't think it's too much of a difference either way, since any behavioral changes tend to break at least one thing. I'll change the precedence to use TC main over scampy.

PR #343, opened 2021-10-03
solve grid mean equations explicitly 

yairchn:

PR #347, opened 2021-10-04
Remove GMV new

charleskawczynski:
This PR moves the explicit part of the grid mean equations to the explicit part of the implicit solve--allowing us to remove the grid mean `new` property. This PR also inlines some functions to prepare for more loop fusing.

### DIFF
--- a/src/Turbulence_PrognosticTKE.jl
+++ b/src/Turbulence_PrognosticTKE.jl
@@ -505,10 +505,10 @@ function update(edmf::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBas
         implicit_eqs.b_q_tot_gm[k] = GMV.QT.values[k] + TS.dt * GMV.QT.tendencies[k]
         implicit_eqs.b_θ_liq_ice_gm[k] = GMV.H.values[k] + TS.dt * GMV.H.tendencies[k]
         if is_surface_center(grid, k)
-            implicit_eqs.b_u_gm[k] += TS.dt * Case.Sur.rho_uflux * Δzi * ref_state.alpha0_half[k] / edmf.ae[k]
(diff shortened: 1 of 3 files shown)


src/Turbulence_PrognosticTKE.jl
```diff
@@ -500,15 +500,15 @@ function update(edmf::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBas
     implicit_eqs.A_uv_gm .= construct_tridiag_diffusion_gm(grid, TS.dt, edmf.rho_ae_KM, ref_state.rho0_half, edmf.ae)
     Δzi = grid.Δzi
     @inbounds for k in real_center_indices(grid)
-        implicit_eqs.b_u_gm[k] = GMV.U.values[k]
-        implicit_eqs.b_v_gm[k] = GMV.V.values[k]
-        implicit_eqs.b_q_tot_gm[k] = edmf.EnvVar.QT.values[k]
```
ilopezgp, line 505: Was this an error before, where we were taking the environmental values?

src/Turbulence_PrognosticTKE.jl
```diff
@@ -500,15 +500,15 @@ function update(edmf::EDMF_PrognosticTKE, GMV::GridMeanVariables, Case::CasesBas
...
-        implicit_eqs.b_θ_liq_ice_gm[k] = edmf.EnvVar.H.values[k]
+        implicit_eqs.b_u_gm[k] = GMV.U.values[k] + TS.dt * GMV.U.tendencies[k]
+        implicit_eqs.b_v_gm[k] = GMV.V.values[k] + TS.dt * GMV.V.tendencies[k]
```
ilopezgp, line 504: So we are computing the implicit tendency in here in terms of an *intermediate step* ?

ilopezgp:
Let's document the reasons behind this clearly before changing the solver. I am not sure the proposed solution is correct.

charleskawczynski:
Superseded by #343

Issue #329, opened 2021-09-30
Inconsistent inversion heigh computations


charleskawczynski:
I just noticed that [update_inversion](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/Turbulence_PrognosticTKE.jl#L1528) computes the inversion height in different ways depending on the given `option`, however, `get_inversion` is _always_ used in [Surface.jl](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/Surface.jl#L14).

This appears to be the case in SCAMPy, too.

charleskawczynski:
Closed by #348.

PR #348, opened 2021-10-04
change initialize covariance 

yairchn:
This PR initializes the grid mean covariances in Cases.jl as functions of height only (no inversion height or wstar). It modifies 'initialize_covariance' function to set only the environmental covariances. 
It will also allow to better organized the call for 'update_aux' which relies on these changes. It would close this issue:
https://github.com/CliMA/TurbulenceConvection.jl/issues/329


ilopezgp:
You can check the initial TKE fields, they are not as close to the LES in all cases where this was used. I would keep it, since TKE=0 is not a good initialization.

yairchn:
> You can check the initial TKE fields, they are not as close to the LES in all cases where this was used. I would keep it, since TKE=0 is not a good initialization.

In which case ?

ilopezgp:
Boundary layer cases that do not have another initialization. In this case, Dycoms and Tan life cycle.

src/Turbulence_PrognosticTKE.jl
```diff
@@ -14,8 +14,6 @@ function initialize(
     param_set = parameter_set(edmf)
     grid = get_grid(edmf)
     update_aux!(edmf, gm, grid, Case, ref_state, param_set, TS)
-    update_inversion(edmf, gm, Case.inversion_option)
```
ilopezgp, line 17: I would like to keep inversion height as a diagnostic variable in the netcdf, if possible.

ilopezgp:
RICO went crazy

trontrytel:
> RICO went crazy

Wow, it's different for sure. I would say it's within the range of what can be calibrated with the autoconversion timescale. So if all the other cases are looking good/better this one could be improved by that

src/Turbulence_PrognosticTKE.jl
```diff
@@ -1162,85 +1160,16 @@ end
...
-        end
-    end
-
-    if ws > 0.0
-        @inbounds for k in real_center_indices(grid)
-            z = grid.zc[k]
-            # need to rethink of how to initilize the covarinace profiles - for now took the TKE profile
-            gm.Hvar.values[k] = gm.Hvar.values[kc_surf] * tke_gm(z)
-            gm.QTvar.values[k] = gm.QTvar.values[kc_surf] * tke_gm(z)
-            gm.HQTcov.values[k] = gm.HQTcov.values[kc_surf] * tke_gm(z)
```
ilopezgp, line 1216: @yairchn I think you forgot to add this to the cases

ilopezgp:
Judging from the contours of TKE for Bomex, it seems like the model is not using the values you initialized in Cases.jl. I don't know why, though. Maybe it is related to the other covariances being initialized differently (as zero), instead of like before (see comment above).

src/Turbulence_PrognosticTKE.jl
yairchn: 'tke_gm(z)' was defined by the ws, we are avoiding this and just defining the TKE profile case by case

ilopezgp: Yes, but I think it would still be good to define the covariances in terms of the TKE profile times the surface variance as a starting point.

yairchn:
> > RICO went crazy
> 
> Wow, it's different for sure. I would say it's within the range of what can be calibrated with the autoconversion timescale. So if all the other cases are looking good/better this one could be improved by that

What happen there is that the initial TKE was non zero in places where TKE dissipation is effectively zero (large mixing length). This TKE at higher levels just slowly mixed the cloud from the top. I don't think it microphysics to blame on this one ;-)

src/Turbulence_PrognosticTKE.jl
yairchn: Added back

charleskawczynski (approved):

integration_tests/utils/mse_tables.jl
```diff
@@ -5,33 +5,33 @@
...
+all_best_mse["ARM_SGP"]["u_mean"] = 5.76247098285726e8
+all_best_mse["ARM_SGP"]["tke_mean"] = 999.5973333265448
+all_best_mse["ARM_SGP"]["temperature_mean"] = 0.00014620899330000358
+all_best_mse["ARM_SGP"]["ql_mean"] = 229.10440600129382
+all_best_mse["ARM_SGP"]["thetal_mean"] = 0.00013751993726485138
+all_best_mse["ARM_SGP"]["Hvar_mean"] = 446.9040361383814
+all_best_mse["ARM_SGP"]["QTvar_mean"] = 159.74942479096785
 #
 all_best_mse["Bomex"] = OrderedDict()
-all_best_mse["Bomex"]["qt_mean"] = 0.09937060369154255
```
ilopezgp, line 22: Is there a reason why the mse tables for boomer/dycoms/gabls have to be changed? These should not have changed with your update.

ilopezgp (approved):

yairchn:
Bors r+

PR #350, opened 2021-10-05
DoNotMerge: Delete initializer for all cases

ilopezgp:


yairchn:
The buildkit look good to me!

PR #354, opened 2021-10-05
some more changes to the order in update 

yairchn:


charleskawczynski (approved):
Fantastic. We are moving! 🚀

We'll need to update the tables after #336 lands.

charleskawczynski:
Wow, looks like this is now breaking for `LES_driven_SCM` 😕. Idk if that was just from merging with main (which I don't think has had any major differences recently), or what.

charleskawczynski (approved):

PR #355, opened 2021-10-05
Inline several functions in preparation to fuse operations

charleskawczynski:
This PR inlines several functions into `update_aux!` in preparation to fuse more operations. It's a bit mind-bending that this resulted in changes. I did change some `1.0` to `1` to make a few things type stable, maybe that's it?


charleskawczynski:
Results look virtually unchanged

PR #362, opened 2021-10-07
add theta tendency from rain evaporation

trontrytel:
This fixes a bug where we were computing the rain evaporation tendency to theta, but we were not saving or using it anywhere. 

It changes the Rico case, so please take a look. (I'm splitting it from the other cleanup changes that I'm doing, which should not be changing results)


charleskawczynski (approved):
The code changes look fine to me. Are we okay with the changes in results? The contours appear quite different.

PR #364, opened 2021-10-08
initial TKE in Rico 

yairchn:
Using inversion height to initialize the TKE in Rico


charleskawczynski (approved):
This looks fine by me. @ilopezgp is this what you suggested we add?

ilopezgp (approved):

PR #367, opened 2021-10-08
Prep form calculating microphysics tendencies

trontrytel:
Preparation work for computing and applying all precipitation tendencies.

This PR:
- replaces the misleading `source` with `tendency` for all microphysics processes
- deletes the `clear_precip_sources` and  `update_total_precip_sources` functions (not needed now because we don't do sub-steps for updrafts)
- gets rid of the `evaporate_all` logic in rain evaporation (positivity limiters should be handled separately elsewhere) 
- stores different microphysics tendencies into helper arrays (to be changed later to only augment the main tendencies for all state variables).


src/EDMF_Rain.jl
```diff
@@ -41,23 +40,26 @@ function rain_diagnostics(
...
+function compute_subdomain_rain_tendency_sum(
     rain::RainVariables,
     up_thermo::UpdraftThermodynamics,
     en_thermo::EnvironmentThermodynamics,
     TS::TimeStepping,
 )
+    # TODO - just move it to the place where all rain tendecies are summed up
     @inbounds for k in real_center_indices(rain.grid)
-        rain.QR.values[k] -= (en_thermo.prec_source_qt[k] + up_thermo.prec_source_qt_tot[k]) * TS.dt
+        rain.QR.values[k] -=
```
yairchn, line 61: maybe '+= -RHS'

trontrytel: why is that better?

yairchn: @charleskawczynski I think I remember you saying this is better, am I wrong?

charleskawczynski: I think I'd prefer it be in the form:
```
x{n+1} = x^{n} + Δt * tendencies
```
So that all the signs are correct in `tendencies` (DifferentialEquations basically requires `tendencies`).

trontrytel: In that case, do you want me to create a new array for qr_tendency?

src/types.jl
```diff
@@ -844,7 +851,7 @@ mutable struct EDMF_PrognosticTKE{PS, A1, A2, IE}
     lambda_stab::Float64
     minimum_area::Float64
     Rain::RainVariables
-    rainphysics::RainPhysics
+    RainPhys::RainPhysics
```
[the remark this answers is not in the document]
trontrytel: I was following what we are doing for updaft and environment thermodynamics:

https://github.com/CliMA/TurbulenceConvection.jl/blob/0e5751baa5b70d50d23c90de9f7e59b105429778/src/types.jl#L856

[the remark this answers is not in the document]
trontrytel: also, I don't think that it is a variable?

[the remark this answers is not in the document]
charleskawczynski: Yeah, I've been wanting to change these for a while but it's been sitting on the back burner. Also, it is a variable name, it just happens to be the property of a struct.

src/EDMF_Rain.jl
charleskawczynski: That would be great!

That said, if it's easier to do this incrementally, I'm fine with merging now and cleaning up later.

charleskawczynski (approved):
Overall, the code changes look good to me. I think we could improve on some names and massage the form of some of the equations, but I don't think these are blockers.

PR #384, opened 2021-10-12
replacing universal functions

yairchn:
In this PR I am replacing the use of universal functions with those from SF.jl. 
Note that the universal functions in TC.jl already compute the difference in values between z1 and z0 while the universal functions in SF.jl both need to be computed individually. 
Furthermore , in SF.jl the universal functions for stable and unstable conditions are fused into a single function while in TC.jl different functions are coded for different stability conditions.


src/surface_functions.jl
```diff
@@ -40,13 +40,17 @@ function compute_ustar(param_set, windspeed, buoyancy_flux, z0, z1)
...
+            lmo = -ustar * ustar * ustar / (buoyancy_flux * vkb)
             zeta = z1 / lmo
             zeta0 = z0 / lmo
-            return windspeed - ustar / vkb * (logz - psi_m(zeta, zeta0))
+            psi_m = zeta >= 0.0 ? psi_m_stable : psi_m_unstable
+            uf = SF.Businger(param_set, lmo)
+            ζ = z1 / lmo
+            ζ₀ = z0 / lmo
+            Ψₘ₁ = SF.psi(uf, ζ, SF.MomentumTransport())
+            Ψₘ₀ = SF.psi(uf, ζ₀, SF.MomentumTransport())
```
charleskawczynski, line 52: I can't see the `Ψ` subscript (my browser doesn't recognize these characters). Let's just use `Ψ_m_1`/`Ψ_m_2` instead.

src/surface_functions.jl
```diff
@@ -69,32 +48,33 @@ function exchange_coefficients_byun(param_set, Ri, zb, z0, cm, ch, lmo)
...
+            ζ = zfactor * (-2.0 * sqrt(qb) * cos(angle / 3.0) + 1.0 / (3.0 * gamma_m))
         end
-        lmo = zb / zeta
-        zeta0 = z0 / lmo
-        psi_m = psi_m_unstable(zeta, zeta0)
-        psi_h = psi_h_unstable(zeta, zeta0)
+        lmo = zb / ζ
+        ζ_0 = z0 / lmo
     end
+    uf = SF.Businger(param_set, convert(Float64, lmo))
```
yairchn, line 70: @charleskawczynski any suggestions of a better way to deal with the type issue raised by 'uf = SF.Businger(param_set, lmo)'
the above code the way I found around it

src/surface_functions.jl
```diff
@@ -9,44 +9,23 @@ function buoyancy_flux(param_set, shf::FT, lhf, T_b, qt_b, α0_b, ts) where {FT}
...
     ustar0 = windspeed * vkb / logz
     monin_obukov(ustar) = -ustar^3 / (buoyancy_flux * vkb)
     if (abs(buoyancy_flux) > 1.0e-20)
-        ζ_1 = z1 / monin_obukov(ustar0)
-        psi_m = ζ_1 >= 0.0 ? psi_m_stable : psi_m_unstable
         function roots(ustar)
-            lmo = monin_obukov(ustar)
+            lmo = -ustar * ustar * ustar / (buoyancy_flux * vkb)
             zeta = z1 / lmo
             zeta0 = z0 / lmo
```
ilopezgp, line 22: Are `zeta` and `zeta0` used anywhere?

yairchn:
Bors r+

yairchn:
Bors r+

PR #405, opened 2021-10-15
Add consistent environmental buoyancy gradients using mean fields and quadratures.

ilopezgp:
This PR modifies the `buoyancy_gradients` closure and the struct used to compute them in the following ways:

- The outputs of `buoyancy_gradients` are modified to be the buoyancy gradients in the environment, and its dry and cloudy parts. Before, a couple of partial derivatives were outputted that were not used by the model, or just added up in functions in the code. I output the dry and moist derivatives instead because I may use them to simplify the mixing length formulation in the near future.
- Two implementations of the environmental buoyancy gradient are given. The first one corrects the implementation in Tan et al. (2018) in several ways. First, it exclusively uses environmental mean quantities for computation of buoyancy fluxes, which is consistent with not having access to a probabilistic cloud partition function. This is consistent with the original paper, although we later modified it in the code to use partial information from the quadratures. Second, it corrects the computation of the buoyancy gradient in the dry part from the one given in the paper. The derivation of this term in Tan et al (2018) is taken from Deardorff 1976 without consideration for the different assumptions allowed, and it is actually incorrect because `θ_liq_ice != θ` in the environment, just in the dry part of the environment. Third, it corrects the latent heat used to include effects of all condensate species.
- The second implementation is consistent with having `Quadratures` in the model. The derivatives are the same as in the previous case, therefore you will find that the `buoyancy_gradients` function is shared. The main distinction is in the arguments passed with the `EnvBuoyGrad` struct. In this case, we make full use of the quadrature information to compute the gradients. Safeguards are included for gradients computed in vanishing cloudy or dry regions. In these regions, we resort to the environmental mean gradients.


src/types.jl
```diff
@@ -154,6 +154,9 @@ Base.@kwdef struct GradBuoy{FT}
     ∂b∂z_θl::FT
 end
 
+Base.@kwdef mutable struct bg_struct{T}
+end
```
charleskawczynski, line 158: Let's change the name to something more descriptive, and use TitleCase. Also, let's not use `mutable` if we can avoid it.

src/update_aux.jl
```diff
@@ -288,21 +288,51 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
+                thl_cloudy = en_thermo.thl_cloudy[k],
+                ∂qt∂z_dry = ∂qt∂z_dry,
+                ∂θ∂z_dry = ∂θ∂z_dry,
+                ∂qt∂z_cloudy = ∂qt∂z_cloudy,
+                ∂θl∂z_cloudy = ∂θl∂z_cloudy,
+                p0 = p0_c[k],
+                en_cld_frac = en.cloud_fraction.values[k],
+                alpha0 = α0_c[k],
+            )
+        end
```
charleskawczynski, line 334: If we unify the propertynames, we could get away with this instead:

```julia
        kwargs = (;
                qt_dry = en_thermo.qt_dry[k],
                th_dry = en_thermo.th_dry[k],
                t_cloudy = en_thermo.t_cloudy[k],
                qv_cloudy = en_thermo.qv_cloudy[k],
                qt_cloudy = en_thermo.qt_cloudy[k],
                th_cloudy = en_thermo.th_cloudy[k],
                thl_cloudy = en_thermo.thl_cloudy[k],
                ∂qt∂z = ∂qt∂z,
                ∂θl∂z = ∂θl∂z,
                p0 = p0_c[k],
                en_cld_frac = en.cloud_fraction.values[k],
                alpha0 = α0_c[k],
            )
        bg_model = if edmf.bg_model == bg_struct{Tan2018}
            Tan2018(;kwargs...)
        elseif edmf.bg_model == bg_struct{GradBuoyQuadratures}
            GradBuoyQuadratures(;kwargs...)
        end
        bg = buoyancy_gradients(param_set, edmf.bg_model)
```

And this will be even shorter if we pass thermo states into the struct.

charleskawczynski: Yeah, or you could do `common_kwargs =`, and then `foo(; common_kwargs..., case_specific_kwargs...)`

src/update_aux.jl
```diff
@@ -329,7 +329,7 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
             THV_dry_cut = ccut(en_thermo.thv_dry, grid, k)
             ∂θv∂z_dry = c∇(THV_dry_cut, grid, k; bottom = SetGradient(0), top = SetGradient(0))
             # Gradient safeguard in inner domain -- default to environmental mean
-            if k != 0 && k != real_center_indices(grid)[end]
+            if k != 1 && k != real_center_indices(grid)[end]
```
charleskawczynski, line 332: `k` is not an `Int`, so `k != 1` will always be true. I think you want to compare `k` with the result of one of [these functions](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/Grid.jl#L63-L66)

ilopezgp:
For the MSE, I will wait until we decide whether we want to keep `quadratures` or `mean` as the default mode for the buoyancy flux. I prefer `quadratures`, since it should be more accurate.

src/closures/buoyancy_gradients.jl
```diff
@@ -1,36 +1,71 @@
...
-        ∂b∂qt_cld = 0
+        ∂b∂z_θl_cld = FT(0)
+        ∂b∂z_qt_cld = FT(0)
     end
-    ∂b∂θl = (bg_model.en_cld_frac * ∂b∂θl_cld + (1 - bg_model.en_cld_frac) * ∂b∂θl_dry)
-    ∂b∂qt = (bg_model.en_cld_frac * ∂b∂qt_cld + (1 - bg_model.en_cld_frac) * ∂b∂qt_dry)
 
-    ∂b∂z_θl = ∂b∂θl * bg_model.∂θl∂z
-    ∂b∂z_qt = ∂b∂qt * bg_model.∂qt∂z
+    ∂b∂z_dry = bg_model.en_cld_frac < FT(1) ? ∂b∂θv * bg_model.∂θv∂z_dry : FT(0)
```
yairchn, line 65: dry here means cloud free, right ? shouldn't '∂b∂z_dry' include a contribution from '∂b∂z_qt_dry' (i.e. from ) as well? wouldn't that  the third term in the middle part of equation (4.2) in your document?

trontrytel: Would it make sense to replace `dry` with `cloud_free` or something similar in the code?

src/closures/buoyancy_gradients.jl
```diff
@@ -1,36 +1,71 @@
...
+    return GradBuoy(∂b∂z, ∂b∂z_dry, ∂b∂z_cloudy)
+end
+
+"""
+    buoyancy_gradient_chain_rule(
+    bg_model::EnvBuoyGrad{FT, EBG},
+    ∂b∂θv::FT,
+    ∂b∂θl_cld::FT,
+    ∂b∂qt_cld::FT,
+) where {FT <: Real, EBG <: EnvBuoyGradClosure}
```
charleskawczynski, line 46:
```suggestion
    buoyancy_gradient_chain_rule(
        bg_model::EnvBuoyGrad{FT, EBG},
        ∂b∂θv::FT,
        ∂b∂θl_cld::FT,
        ∂b∂qt_cld::FT,
    ) where {FT <: Real, EBG <: EnvBuoyGradClosure}
```

src/update_aux.jl
```diff
@@ -325,10 +309,68 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
+            QT_cloudy_cut = ccut(en_thermo.qt_cloudy, grid, k)
+            ∂qt∂z_cloudy = c∇(QT_cloudy_cut, grid, k; bottom = SetGradient(0), top = SetGradient(0))
+            THL_cloudy_cut = ccut(en_thermo.thl_cloudy, grid, k)
+            ∂θl∂z_cloudy = c∇(THL_cloudy_cut, grid, k; bottom = SetGradient(0), top = SetGradient(0))
+            THV_dry_cut = ccut(en_thermo.thv_dry, grid, k)
+            ∂θv∂z_dry = c∇(THV_dry_cut, grid, k; bottom = SetGradient(0), top = SetGradient(0))
+            # Gradient safeguards for dry/cloudy -- default to environmental mean
+            # Assumes no fog or domain-top cloudiness
+            if k != kc_surface(grid) && k != kc_top_of_atmos(grid)
+                if cf[k - 1] * cf[k + 1] ≈ 0.0
```
charleskawczynski, line 347: This is a very strange condition. Why do the equations depend on this?

src/update_aux.jl
```diff
@@ -325,10 +309,68 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
+                    ∂qt∂z_cloudy = ∂qt∂z
+                    ∂θl∂z_cloudy = ∂θl∂z
+                end
+                ∂θv∂z_dry = (1 - cf[k - 1]) * (1 - cf[k + 1]) ≈ 0.0 ? ∂θv∂z : ∂θv∂z_dry
```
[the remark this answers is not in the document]
ilopezgp: The gradient across vanishing conditional domains has no meaning. Therefore, we only compute unsaturated and saturated gradients when there actually are unsaturated/saturated regions in the cells used to compute the gradient.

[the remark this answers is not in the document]
charleskawczynski: I would prefer that any non-collocated operators exist in Operators.jl. These conditions have assumptions about the stencils, which we've put a lot of work into abstracting away. If there is a way to recover this condition from the gradient itself (and maybe values), I think that would be a much preferred approach.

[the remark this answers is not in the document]
ilopezgp: I have moved the operator to Operators.jl now.

src/Operators.jl
```diff
@@ -250,6 +250,54 @@ function c∇(f::SA.SVector, grid::Grid, ::BottomBCTag, ::Extrapolate)
...
+function c∇_vanishing_subdomain(
+    f::SA.SVector,
+    sd::SA.SVector,
+    default∇::FT,
+    grid::Grid,
+    ::InteriorTag,
+) where {FT <: Real}
+    @assert length(f) == 3
+    @assert length(sd) == 3
+    if sd[1] * sd[3] ≈ FT(0)
```
charleskawczynski, line 292: I guess this is the condition that makes me feel uneasy. Where/how is this derived? What's the spatially continuous form of the equation being implemented?

ilopezgp: > Looks great to me!
> 
> My only comments are about the notation. I would vote for:
> 
> * using some abbreviation of saturated/unsaturated instead of cloud/dry
> * agreeing on some convention about the thetas so we don't confuse ourselves later. Maybe something like \theta, \theta_v, \theta_liq_ice
> 
> But if the others disagree, we can keep the notation we used. Don't want to block this PR

I agree. After this, we should change all the dry and cloudy variables to unsat and sat. They are also the ones with `th` instead of `\theta`, so it can be done in one go.

PR #411, opened 2021-10-16
Add pipeline to wipe the depot

charleskawczynski:
This PR adds a pipeline script (and a second TC.jl pipeline was added) to clear the Julia depot. We may need to periodically clear the depot if we want to continue caching precompiled code on central. So now, I think, we can just click the build button on buildkite to clear it and that (hopefully) will fix CI troubles.


charleskawczynski:
Whoops, I think we just want the pipeline without the yml file. Otherwise this will run on every push. Closing.

PR #416, opened 2021-10-17
fix surface sat vap pressure, remove T_freeze overwrite

trontrytel:
This PR removes the `T_freeze` overwrite for all cases except `TRMM` and `GATE`.

It also replaces the surface saturation vapour pressure calculation that assumed liquid only, with one that assumes phase partitioning depending on temperature.


ilopezgp:
The changes seem rather large...

yairchn:
if I am not wrong they still don't reproduce all changes you had in the big free PR, right ?

PR #424, opened 2021-10-19
Qualify NameList methods, improve plotting warnings

charleskawczynski:
Qualifying NameList allows us to modify the namelist in interactive sessions. Which dataset has missing data per variable is now included in the error message.


charleskawczynski:
bors ping

PR #474, opened 2021-10-29
Use OrdinaryDiffEq.jl for timestepping, rearrange update_aux!, and remove buoyancy extrapolation

charleskawczynski:
#417 inspired me to see what would happen if, when trying to use OrdinaryDiffEq for time stepping, we also try to remove buoyancy extrapolation and rearrange the update_aux! calls.


charleskawczynski:
Closing in favor of #473.

PR #476, opened 2021-10-29
Fix update_aux inconsistencies

charleskawczynski:
Similar to #417, this PR computes some grid mean quantities before they are referenced. To do this, we rely on using thermodynamic relations instead of the domain decomposition relations.

This does add a saturation adjustment, but this will likely be needed anyway, and can be performed by the dycore in the long run.

### DIFF
--- a/src/update_aux.jl
+++ b/src/update_aux.jl
@@ -84,14 +84,18 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
     #####
     ##### diagnose_GMV_moments
     #####
-    #! format: off
+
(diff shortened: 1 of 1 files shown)


trontrytel:
> > You mean the lifecycle case results?
> 
> All of them, it seems like `q_liq_mean` is consistently zero...

Dycoms still has water. So maybe its just the updrafts

charleskawczynski:
> Dycoms still has water. So maybe its just the updrafts

Oh, weird!

charleskawczynski:
I think #507 supersedes this for the most part.

PR #478, opened 2021-10-29
Compute RH, avoid re-computations

charleskawczynski:
`aux_gm.T` and `aux_gm.buoy` are already computed above in this function, so this PR removes the re-computations to avoid potential inconsistencies. This PR also adds computations for `aux_gm.RH`, which were previously computed once during initialization.

### DIFF
--- a/src/update_aux.jl
+++ b/src/update_aux.jl
@@ -84,14 +84,19 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
     #####
     ##### diagnose_GMV_moments
     #####
-    #! format: off
     get_GMV_CoVar(edmf, grid, state, :Hvar, :θ_liq_ice)
     get_GMV_CoVar(edmf, grid, state, :QTvar, :q_tot)
     get_GMV_CoVar(edmf, grid, state, :HQTcov, :θ_liq_ice, :q_tot)
     GMV_third_m(edmf, grid, state, :Hvar, :θ_liq_ice, :H_third_m)
     GMV_third_m(edmf, grid, state, :QTvar, :q_tot, :QT_third_m)
     GMV_third_m(edmf, grid, state, :tke, :w, :W_third_m)
-    #! format: on
+
+    @inbounds for k in real_center_indices(grid)
+        θ_liq_ice = prog_gm.θ_liq_ice[k]
+        q_tot = prog_gm.q_tot[k]
+        ts = thermo_state_pθq(param_set, p0_c[k], θ_liq_ice, q_tot)
+        aux_gm.RH[k] = TD.relative_humidity(ts)
+    end
 
     #####
     ##### decompose_environment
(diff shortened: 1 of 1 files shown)


charleskawczynski:
This PR is quite interesting

src/update_aux.jl
```diff
@@ -84,14 +84,19 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
-    #! format: off
     get_GMV_CoVar(edmf, grid, state, :Hvar, :θ_liq_ice)
     get_GMV_CoVar(edmf, grid, state, :QTvar, :q_tot)
     get_GMV_CoVar(edmf, grid, state, :HQTcov, :θ_liq_ice, :q_tot)
     GMV_third_m(edmf, grid, state, :Hvar, :θ_liq_ice, :H_third_m)
     GMV_third_m(edmf, grid, state, :QTvar, :q_tot, :QT_third_m)
     GMV_third_m(edmf, grid, state, :tke, :w, :W_third_m)
-    #! format: on
+
+    @inbounds for k in real_center_indices(grid)
```
ilopezgp, line 94: Can't this be added to the loop in line 80?

src/update_aux.jl
```diff
@@ -222,10 +227,10 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
     #####
     a_up_bulk = aux_tc.bulk.area
     @inbounds for k in real_center_indices(grid)
+        # TODO: Can we use thermo states here? Is this consistent with one constructed
+        #       from a thermo state?
         aux_gm.q_liq[k] = (a_up_bulk[k] * aux_tc.bulk.q_liq[k] + (1 - a_up_bulk[k]) * aux_en.q_liq[k])
         aux_gm.q_ice[k] = (a_up_bulk[k] * aux_tc.bulk.q_ice[k] + (1 - a_up_bulk[k]) * aux_en.q_ice[k])
-        aux_gm.T[k] = (a_up_bulk[k] * aux_tc.bulk.T[k] + (1 - a_up_bulk[k]) * aux_en.T[k])
```
ilopezgp, line 227: I don't see the update to `aux_gm.T[k]` anywhere before this in this file.

src/update_aux.jl
```diff
@@ -222,10 +227,10 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
-        aux_gm.buoy[k] = (a_up_bulk[k] * aux_tc.bulk.buoy[k] + (1 - a_up_bulk[k]) * aux_en.buoy[k])
```
ilopezgp, line 228: I actually think we have to keep this. If you see where it is defined above, we are computing the GMV buoyancy to subtract it to updrafts and environment. Therefore, we have to recompute it here -- it should be equal to zero in this second step.

ilopezgp: This is due to the implied balance in the SCM, see equation (47) in Cohen et al.

src/update_aux.jl
charleskawczynski: Oh, yeah, maybe I mixed this up

charleskawczynski:
I think #507 basically fixed things here. Closing.

PR #479, opened 2021-10-29
Move cloud fraction to diagnostics climacore field

charleskawczynski:


charleskawczynski:
Going to re-do this from scratch since cloud fraction is potentially an auxiliary field instead of diagnostic.

Issue #468, opened 2021-10-28
Simulations depend on export frequency


charleskawczynski:
@trontrytel and @yairchn recently discovered that results were different depending on the export frequency. This means that some of these "diagnostics" should really be auxiliary variables, which the simulation depend on. Simulations should not depend on purely diagnostic quantities, and we should probably have a clear separation between the two.

charleskawczynski:
I believe that this was fixed by #481, but I think we should hook a test in to ensure that this issue doesn't return

charleskawczynski:
Actually, I think the split between the aux and diagnostic arrays should be enough. Closing.

PR #481, opened 2021-10-29
Move auxiliary computations to update aux, consolidate io

charleskawczynski:
This PR
 - Consolidates all of the `io` (and `initialize_io`) methods into a single file, which I think we should plan to deprecate in favor of using diagnostics.jl.
 - Moves some auxiliary computations (which the simulation depends on) out from `io`, which occurs every `frequency`, to `update_aux!`

Related issue: #468


charleskawczynski:
Results actually look relatively unchanged. Merging!

PR #483, opened 2021-10-30
Make env entr-detr src terms explicit

charleskawczynski:
As a step towards making the environmental equations explicit, this PR makes the entr-detr source terms explicit.


charleskawczynski:
Superseded by #484.

PR #484, opened 2021-10-31
Make env entr-detr and mixing length src terms explicit

charleskawczynski:
This is an incremental step towards making the environment source terms all explicit.

Results look virtually unchanged.


charleskawczynski:
Supersedes #483.

PR #486, opened 2021-10-31
Make env diffusion explicit

charleskawczynski:
This PR changes the environmental diffusion to be treated explicitly in time. Similar to #485, this should help the transition to using OrdinaryDiffEq.jl for timestepping. Also, this is helpful because it removes the (decomposed) spatial operators.

~I've reduced the timestep in the drybubble case and rico. All of the results look good AFAICT, but the Hvar / QTVar prognostic variables look like they're severely diverging in ARM, Nieuwstadt, and DryBubble. We could further reduce the timestep (which I imagine would fix the problem), but I'm wondering if this is worthwhile since this doesn't technically affect our other results and we should (hopefully) go back to treating some terms implicitly once #473 merges.~

This was fixed by further reducing timesteps.


charleskawczynski:
Ok, results all look good. I had to further decrease the timestep for ARM, but this shouldn't increase the overall CI time. I'm hoping that we'll soon have implicit time marching and we can bump all of these timesteps up.

PR #487, opened 2021-11-01
Replace environment matrix inverse with explicit update

charleskawczynski:


charleskawczynski:
Results are visibly unchanged.

PR #488, opened 2021-11-01
Solve for conservative, not primitive, prognostic env variables

charleskawczynski:
This will help us be compatible with using OrdinaryDiffEq.jl for timestepping


charleskawczynski:
Results look virtually unchanged

PR #489, opened 2021-11-01
fix the use of env_b from previous timestep in buoyancy extrap

yairchn:
We should not use environment buoyancy - that is diagnosed from updraft buoyancy - as alternative. Moreover if area is smaller than minimum aux updraft values already default to grid mean in update_aux 55


charleskawczynski (approved):
LGTM

PR #494, opened 2021-11-02
Make precip variables ClimaCore fields, rm unused funcs, cleanup

charleskawczynski:
This PR
 - Makes several precipitation variables ClimaCore fields
 - Removes unused functions

I think that, for now, we can keep these tendencies (which are effectively saved for diagnostics) in `aux`. In the long run, I envision having functions that we can accumulate tendencies inside `compute_tendencies`, and they can be computed separately (using the same machinery) in the diagnostics.

### DIFF
--- a/integration_tests/utils/main.jl
+++ b/integration_tests/utils/main.jl
@@ -112,6 +112,10 @@ cent_aux_vars_edmf(FT, n_up) = (;
             QTvar = FT(0),
             HQTcov = FT(0),
         ),
+        θ_liq_ice_tendency_rain_evap = FT(0),
+        qt_tendency_rain_evap = FT(0),
+        qr_tendency_rain_evap = FT(0),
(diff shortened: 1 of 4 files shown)


trontrytel:
I'm changing some of those tendencies now. Adding snow really increases the number of partial tendencies from different processes. Maybe after I have a draft PR with snow we can discuss what should be moved to diagnostics

charleskawczynski:
> I'm changing some of those tendencies now. Adding snow really increases the number of partial tendencies from different processes. Maybe after I have a draft PR with snow we can discuss what should be moved to diagnostics

Ah, ok, sorry! Feel free to r- if you'd prefer to tackle this in a later PR!

trontrytel:
We can merge it. I will rebase carefully. 

I think I'll delete the unneeded tendencies and just augment the one we need. And later we can think of a way to handle output diagnostics

PR #502, opened 2021-11-02
Add snow

trontrytel:
@charleskawczynski  - this is how it looks right now with regards to adding the snow related processes. I mostly just have to finish the source terms to theta_liq_ice for some of the more confusing microphysics interactions.

It would be great if you could take a look now, and see how it fits with your changes in the data structures, etc

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -9,6 +9,7 @@ Base.@kwdef struct PrecipFormation{FT}
     θ_liq_ice_tendency::FT
     qt_tendency::FT
     qr_tendency::FT
+    qs_tendency::FT
 end
 
 """
@@ -181,6 +182,7 @@ end
 Base.@kwdef mutable struct PrecipVariables
     precipitation_model::String = "default_precipitation_model"
     mean_rwp::Float64 = 0
+    mean_swp::Float64 = 0
     cutoff_precipitation_rate::Float64 = 0
 end
 function PrecipVariables(namelist, grid::Grid)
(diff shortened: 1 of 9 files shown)


src/types.jl
```diff
@@ -9,6 +9,7 @@ Base.@kwdef struct PrecipFormation{FT}
     θ_liq_ice_tendency::FT
     qt_tendency::FT
     qr_tendency::FT
+    qs_tendency::FT
```
charleskawczynski, line 12: Let's add `qs` to the prognostic field in `main.jl`. (that will give us _both_ `state.prog.turbconv.qs` and `state.tendencies.turbconv.qs`)

src/types.jl
```diff
@@ -252,26 +254,37 @@ end
...
-    θ_liq_ice_tendency_rain_formation_tot::A1
-    qt_tendency_rain_formation_tot::A1
+    θ_liq_ice_tendency_precip_formation::A2
+    qt_tendency_precip_formation::A2
+    qr_tendency_precip_formation::A2
+    qs_tendency_precip_formation::A2
+    θ_liq_ice_tendency_precip_formation_tot::A1
+    qt_tendency_precip_formation_tot::A1
+    qr_tendency_precip_formation_tot::A1
+    qs_tendency_precip_formation_tot::A1
```
charleskawczynski, line 264: Can we add all of these to the aux state instead? We're trying to _deprecate_ `center_field`/`face_field` because they're ordinary arrays.

charleskawczynski (commented):
Generally speaking, I'd prefer we don't add any more fields to `types.jl`-- I'm actively trying to remove them all

charleskawczynski:
Also, which case can I look at to see snow? ❄️ ⛄ 😄

PR #503, opened 2021-11-03
revise buoy extrap to aviod k-1 index

yairchn:
we are still extrapolating buoyancy but avoiding k-1 with a dummy variable


src/update_aux.jl
```diff
@@ -121,6 +121,37 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
+        ρ = TD.air_density(ts_up)
+        q_liq_tmp = TD.liquid_specific_humidity(ts_up)
+        q_ice_tmp = TD.ice_specific_humidity(ts_up)
+        T_tmp = TD.air_temperature(ts_up)
+        buoy_tmp = buoyancy_c(param_set, ρ0_c[kc_surf], ρ)
+        RH_tmp = TD.relative_humidity(ts_up)
+        a_tmp = aux_up[i].area[kc_surf]
+        @inbounds for k in real_center_indices(grid)
+            if aux_up[i].area[k] < edmf.minimum_area && k > kc_surf && a_tmp > 0.0
+                nothing
```
yairchn, line 136: @charleskawczynski  in such case all '_tmp' vars are not updated and by default match the value of the previous k. this avoids using k-1 but still extrapolates. Not sure if you like this option

charleskawczynski: I'd prefer we have a local, point-wise, solution to removing the buoyancy extrapolation, instead of using non-local conditions. Also, won't `a_tmp > 0.0` always be true? The surface area is prescribed to be a value (greater than zero)

yairchn:
too sneaky!

PR #506, opened 2021-11-03
Prepare data structures for adding snow

trontrytel:
This PR adds all the data structures for storing snow and its tendencies.  Right now all the source terms related to snow are still set to zero. So this PR should not bring any changes.

### DIFF
--- a/src/EDMF_Environment.jl
+++ b/src/EDMF_Environment.jl
@@ -1,11 +1,12 @@
-function update_env_precip_tendencies(en_thermo::EnvironmentThermodynamics, state, k, qt_tendency, θ_liq_ice_tendency)
-
+function update_env_precip_tendencies(state, k, qt_tendency, θ_liq_ice_tendency, qr_tendency, qs_tendency)
     aux_en = center_aux_environment(state)
     tendencies_pr = center_tendencies_precipitation(state)
-    en_thermo.qt_tendency_rain_formation[k] = qt_tendency * aux_en.area[k]
-    tendencies_pr.qr[k] += -en_thermo.qt_tendency_rain_formation[k]
-    en_thermo.θ_liq_ice_tendency_rain_formation[k] = θ_liq_ice_tendency * aux_en.area[k]
 
+    aux_en.qt_tendency_precip_formation[k] = qt_tendency * aux_en.area[k]
+    aux_en.θ_liq_ice_tendency_precip_formation[k] = θ_liq_ice_tendency * aux_en.area[k]
+
+    tendencies_pr.q_rai[k] += qr_tendency * aux_en.area[k]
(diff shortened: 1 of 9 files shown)


src/EDMF_Environment.jl
```diff
@@ -1,11 +1,20 @@
-function update_env_precip_tendencies(en_thermo::EnvironmentThermodynamics, state, k, qt_tendency, θ_liq_ice_tendency)
-
+function update_env_precip_tendencies(
+    en_thermo::EnvironmentThermodynamics,
```
charleskawczynski, line 2: It looks like we can remove the `en_thermo` argument?

src/EDMF_Precipitation.jl
```diff
@@ -60,22 +73,36 @@ function compute_rain_evap_tendencies(::PrecipPhysics, grid, state, gm, TS::Time
...
+        # When we fuse loops, this should hopefully disappear
         q_tot_gm = prog_gm.q_tot[k]
         T_gm = aux_gm.T[k]
-        # When we fuse loops, this should hopefully disappear
-        ts = TD.PhaseEquil_pTq(param_set, p0_c[k], T_gm, q_tot_gm)
+        p0 = p0_c[k]
+        ρ0 = ρ0_c[k]
+        ts = TD.PhaseEquil_pTq(param_set, p0, T_gm, q_tot_gm)
         q = TD.PhasePartition(ts)
+        qv = TD.vapor_specific_humidity(ts)
```
charleskawczynski, line 86: Is this used anywhere?

src/EDMF_Precipitation.jl
```diff
@@ -60,22 +73,36 @@ function compute_rain_evap_tendencies(::PrecipPhysics, grid, state, gm, TS::Time
...
+        qr = prog_pr.q_rai[k]
+        qs = prog_pr.q_sno[k]
```
charleskawczynski, line 88: Is this used anywhere?

charleskawczynski (approved):
Overall LGTM!

src/EDMF_Precipitation.jl
trontrytel: Will be used in the following PR with source terms computations.

src/EDMF_Environment.jl
trontrytel: Thats true. I'll delete it. The corresponding function from updrafts is different. I don't think we will be dispatching over them.

trontrytel:
Let's see the CI plots. I want to double check I'm not changing anything for RiCO

charleskawczynski:
Oh, yeah, something is wrong with Rico

charleskawczynski (commented):
I think we can decompose this PR into smaller pieces.

PR #507, opened 2021-11-03
order update aux - fuse loops

yairchn:
This PR organizes update_aux in order to prevent issues of using aux vars from from previous time steps, fuse k loops and follow this general order:
* upd primitive vars
* bulk primitive vars
* env primitive vars
* ThermodynamicStates ts_gm, ts_up, ts_env
* buoyancy
* diagnostic variables (RH, T , q_liq, q_ice)
* bulk diagnostic variables
* closures

I separated the face and center updates in order to reduce the number of loops from each case


src/update_aux.jl
```diff
@@ -34,84 +34,28 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
-    #####
-    # (Find values of environmental variables by subtracting updraft values from grid mean values.
-    # Make sure the "bulkvalues" of the updraft variables are updated first)
-    # velocity (face indicies)
-    @inbounds for k in real_face_indices(grid)
         aux_tc_f.bulk.w[k] = 0
         a_bulk_bcs = (; bottom = SetValue(sum(edmf.area_surface_bc)), top = SetZeroGradient())
         a_bulk_f = interpc2f(aux_tc.bulk.area, grid, k; a_bulk_bcs...)
-        if a_bulk_f > 1.0e-20
+        if a_bulk_f > 0
```
charleskawczynski, line 58: `aux_tc.bulk.area` in `interpc2f(aux_tc.bulk.area, grid, k; a_bulk_bcs...)` is computed after it's used. Maybe it would be good to first define a `compute_bulk` function and move the bulk variables to diagnostics?

yairchn: The logic should be reversed. We should compute the center variables first so that bulk.area is available when we compute face variables. I will update.

src/update_aux.jl
```diff
@@ -34,102 +34,87 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
-        aux_tc.bulk.q_liq[k] = 0
-        aux_tc.bulk.q_ice[k] = 0
-        aux_tc.bulk.θ_liq_ice[k] = 0
-        aux_tc.bulk.T[k] = 0
-        aux_tc.bulk.RH[k] = 0
-        aux_tc.bulk.buoy[k] = 0
+        val1 = 1 / (1 - a_bulk_c)
+        val2 = a_bulk_c * val1
+
+        aux_en.area[k] = 1 - a_bulk_c
```
charleskawczynski, line 98:
```suggestion
```
This is already computed above

src/update_aux.jl
```diff
@@ -146,84 +131,105 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
+            aux_tc.bulk.T[k] = aux_en.T[k]
         end
+
+        #####
+        ##### update_GMV_diagnostics
+        #####
+        aux_gm.q_liq[k] = (aux_tc.bulk.area[k] * aux_tc.bulk.q_liq[k] + (1 - aux_tc.bulk.area[k]) * aux_en.q_liq[k])
+        aux_gm.q_ice[k] = (aux_tc.bulk.area[k] * aux_tc.bulk.q_ice[k] + (1 - aux_tc.bulk.area[k]) * aux_en.q_ice[k])
+        aux_gm.T[k] = (aux_tc.bulk.area[k] * aux_tc.bulk.T[k] + (1 - aux_tc.bulk.area[k]) * aux_en.T[k])
+        aux_gm.buoy[k] = (aux_tc.bulk.area[k] * aux_tc.bulk.buoy[k] + (1 - aux_tc.bulk.area[k]) * aux_en.buoy[k])
```
charleskawczynski, line 171: Not a blocker, but why aren't we using thermodynamics to compute these from grid mean quantities? Isn't that what the dycore will do without edmf?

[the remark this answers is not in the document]
yairchn: I think you referring to 'aux_gm.buoy[k]', its a question of which "aux_gm.buoy[k]" we would like to keep. 
the RHS of this expression
aux_gm.buoy[k] = (aux_tc.bulk.area[k] * aux_tc.bulk.buoy[k] + (1 - aux_tc.bulk.area[k]) * aux_en.buoy[k])
 is changed in 133-139 after 'aux_gm.buoy[k]' was computed the first time. Therefore 'aux_gm.buoy[k]'  is recomputed here.
All other grid mean variables in these lines are computed only here

yairchn: It is a great point. in theory the updraft and environment are not well mixed so better compute their individual thermodynamic states and get the temperature as area weighted mean. However I am not sure what the dycore will do, it should be discussed in the future.

charleskawczynski (approved):
Overall the code changes look fine to me. It would be nice to get a second set of eyes on this PR, though, since `update_aux!` is pretty intricate.

src/update_aux.jl
```diff
@@ -34,102 +34,85 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
+    @inbounds for k in real_center_indices(grid)
+        #####
+        ##### Set primitive variables
+        #####
         @inbounds for i in 1:(up.n_updrafts)
-            a_up_bcs = (; bottom = SetValue(edmf.area_surface_bc[i]), top = SetZeroGradient())
-            anew_k = interpc2f(aux_up[i].area, grid, k; a_up_bcs...)
-            if anew_k >= edmf.minimum_area
-                aux_up_f[i].w[k] = max(prog_up_f[i].ρaw[k] / (ρ0_f[k] * anew_k), 0)
+            if is_surface_center(grid, k)
```
trontrytel, line 45: Would it make sense to fill out the surface points separately. And then loop over updrafts starting from the first point above the ground?

This way we would avoid checking an if/else for every height point in this loop

src/update_aux.jl
```diff
@@ -34,102 +34,85 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
-        # Assuming gm.W = 0!
-        aux_en_f.w[k] = -a_bulk_f / (1 - a_bulk_f) * aux_tc_f.bulk.w[k]
-    end
-    @inbounds for k in real_center_indices(grid)
+        #####
+        ##### decompose_environment
+        #####
+        # (Find values of environmental variables by subtracting updraft values from grid mean values.
+        # Make sure the "bulkvalues" of the updraft variables are updated first)
+        # velocity (face indicies)
```
trontrytel, line 93: The comment about velocity (face indicies) can be deleted?

trontrytel (approved):

src/update_aux.jl
yairchn: I think entire comment could be deleted

yairchn:
Bors r+

PR #509, opened 2021-11-03
lets see if it helps CI

trontrytel:


charleskawczynski (commented):
Perhaps this identifies that it is related to qi (and the use of strings, here). I'll keep digging into this to see what's going on. Feel free to close

trontrytel:
> Perhaps this identifies that it is related to qi (and the use of strings, here). I'll keep digging into this to see what's going on. Feel free to close

Sounds good. But in case this issue with CI slows down other PRs, I'm ok with merging it for now. And coming back to plotting qi later.

charleskawczynski:
Fair point.

PR #513, opened 2021-11-04
move env_thermo precip arrays to aux

trontrytel:
Peel off from #506 - moves environment precipitation related data structures to aux. 

(does not affect RiCO)

### DIFF
--- a/src/EDMF_Environment.jl
+++ b/src/EDMF_Environment.jl
@@ -1,11 +1,13 @@
-function update_env_precip_tendencies(en_thermo::EnvironmentThermodynamics, state, k, qt_tendency, θ_liq_ice_tendency)
+function update_env_precip_tendencies(state, k, qt_tendency, θ_liq_ice_tendency, qr_tendency, qs_tendency)
 
     aux_en = center_aux_environment(state)
     tendencies_pr = center_tendencies_precipitation(state)
-    en_thermo.qt_tendency_precip_formation[k] = qt_tendency * aux_en.area[k]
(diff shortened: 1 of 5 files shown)


charleskawczynski (approved):

src/EDMF_Environment.jl
```diff
@@ -52,7 +54,14 @@ function sgs_mean(en_thermo::EnvironmentThermodynamics, grid, state, en, precip,
...
         )
         update_sat_unsat(en_thermo, state, k, ts)
-        update_env_precip_tendencies(en_thermo, state, k, mph.qt_tendency, mph.θ_liq_ice_tendency)
+        update_env_precip_tendencies(
+            state,
+            k,
+            mph.qt_tendency,
+            mph.θ_liq_ice_tendency,
+            mph.qr_tendency,
+            mph.qs_tendency,
```
charleskawczynski, line 63: Maybe in a later PR we can change this function to accept `mph`, that way we don't accidentally mix up the arg order.

charleskawczynski: Ahh, but one of the calls doesn't create an `mph` struct. This is fine.

trontrytel: Right. I actually started coding it that way yesterday and then reminded myself that it is also used with quadrature outputs as arguments

charleskawczynski:
Ahh, sorry! I thought these were touching two different parts of the code base! 😬

charleskawczynski:
Ahhh, it was the rain -> precip rename that did it, sorry!

trontrytel:
np - will resolve now

PR #515, opened 2021-11-04
Additional data structs for snow

trontrytel:
The last of the peel offs from #506 Hopefully no changes as well. 🤞


charleskawczynski (approved):
LGTM. Sorry the tests are so brittle!

trontrytel:
> LGTM. Sorry the tests are so brittle!

No worries! They are extremely useful!

I will regenerate and update the mse tables after I run it again with updated aux order. Then r+ 🌦️

trontrytel:
Actually, I think the differences were only caused by the mse table changes from updated aux order. 🔥

PR #517, opened 2021-11-04
use latent heat const with T

trontrytel:
As discussed today, to be consistent with the definition of `theta_liq_ice` we should use constant latent heat.


trontrytel:
Curious how this will affect the plots...

charleskawczynski:
Ah, I think we need to bump Thermodynamics.jl, too, so that we update to the new version

yairchn:
> Curious how this will affect the plots...

Not sure what scares me more, that it would effect the plots, or that it would not effect them ....

trontrytel:
> Ah, I think we need to bump Thermodynamics.jl, too, so that we update to the new version

Right!

ilopezgp (approved):

charleskawczynski (approved):

yairchn (approved):

charleskawczynski (commented):
We could add type casting for type stability, but we can leave it for now and do a type-stability sprint sometime in the future.

PR #520, opened 2021-11-05
Use ClimaCore operators in get_GMV_CoVar

charleskawczynski:
Use ClimaCore operators in `get_GMV_CoVar`. It looks like this results in fewer allocations, so hopefully we'll see some micro improvements as we implement these more widely.


charleskawczynski:
@yairchn, @ilopezgp, @trontrytel, please review this (as it's a good representative example of how we can transition to ClimaCore operators). If you plan to use our TC.jl operators, please consider going straight to ClimaCore operators if possible.

charleskawczynski:
Awesome, this works out of the box

PR #522, opened 2021-11-05
Use ClimaCore operators in covariance_interdomain_src

charleskawczynski:


src/Turbulence_PrognosticTKE.jl
```diff
@@ -651,37 +651,21 @@ function compute_covariance_interdomain_src(
...
-                Δψ = ψ_up[k] - ψ_en[k]
-                aux_covar.interdomain[k] += tke_factor * aux_up[i].area[k] * (1.0 - aux_up[i].area[k]) * Δϕ * Δψ
-            end
-        end
+    @. interdomain = 0
+    @inbounds for i in 1:(edmf.n_updrafts)
+        ϕ_up = getproperty(prog_up[i], ϕ_var)
+        ψ_up = getproperty(prog_up[i], ψ_var)
+        a_up = aux_up[i].area
+        @. interdomain += tke_factor * a_up * (1 - a_up) * (Ic(ϕ_up) - Ic(ϕ_en)) * (Ic(ψ_up) - Ic(ψ_en))
```
simonbyrne, line 668: what are the types of all these objects?

charleskawczynski: They're all scalar fields, except `tke_factor`, which is a `Float64`

charleskawczynski: @simonbyrne, this turned out to be fine. Surprisingly, `@. interdomain = 0` was the issue. I guess we need to use `parent(interdomain) .= 0` for now.

charleskawczynski: (basically ran into [ClimaCore's 273](https://github.com/CliMA/ClimaCore.jl/issues/273))

PR #525, opened 2021-11-05
feat: add storage options to `default_namelist`

haakon-e:
In some cases I may want to

1. generate a default namelist for a case, without needing to saving it to file, or
2. specifying another location than `pwd` to store it to.

This PR resolves both issues.


charleskawczynski (approved):
Looks fine to me. Since this is pretty high-level stuff, can we get a second reviewer. @ilopezgp, what do you think?

ilopezgp (approved):
LGTM

PR #534, opened 2021-11-06
Use ClimaCore operators

charleskawczynski:


charleskawczynski:
Results are virtually unchanged, but I had to update the mse tables.

PR #536, opened 2021-11-06
add microphysics sources

trontrytel:
This PR adds snow microphysics source terms. It uses the the CloudMicrophysics.jl package. It uses the 0-moment autoconversion rates for rain and snow (we can't use the 1-moment snow autoconversion rate because it assumes that supersaturation exists in the model). All the other processes are parameterised using the 1-moment rates. 

The simulation breaks if the autoconversion timescale is long (~500s for example). The reason is that we transport too much condensate and exceed the domain height as a result of a too deep convection. Maybe one option would be to revisit the detrainment parameterisation and see if we need to have a stronger detrainment high up even when the environment RH is high.


trontrytel:
https://user-images.githubusercontent.com/1196696/144521873-ac14ab0f-b395-4b7a-9520-4ea6d1d38d89.mp4

src/EDMF_Precipitation.jl
```diff
@@ -38,6 +43,7 @@ function compute_precipitation_advection_tendencies(edmf, grid, state, gm, TS::T
             CFL_in_snow = Δt / Δz * term_vel_snow[k + 1]
         end
         if max(CFL_in_rain, CFL_in_snow, CFL_out_rain, CFL_out_snow) > CFL_limit
+            println(CFL_in_rain, CFL_in_snow, CFL_out_rain, CFL_out_snow, CFL_limit)
```
ilopezgp, line 46: Do we want to start using `@info` instead of prints?

charleskawczynski (commented):
Code changes look fine to me, the only suggestion I have is to move the CFL computations outside of `src/` (see #611).

trontrytel:
> Code changes look fine to me, the only suggestion I have is to move the CFL computations outside of `src/` (see #611).

Yep - I'm waiting on us to finish with the CFL PR before merging this one. Maybe we can chat after the sprint?

src/EDMF_Precipitation.jl
trontrytel: I can change that, if necessary. In general this CFL check will soon be moved to a different callback that checks all CFLs (precipitation, as well as updraft velocities, etc)

src/TimeStepping.jl
```diff
@@ -6,6 +6,7 @@ mutable struct TimeStepping
     cfl_limit::Float64
     dt_max::Float64
     dt_io::Float64
+    adapt_dt::Bool
```
charleskawczynski, line 9: I think we can remove this now, correct?

src/EDMF_Precipitation.jl
```diff
@@ -47,27 +48,54 @@ function compute_precipitation_sink_tendencies(grid, state, gm, TS::TimeStepping
...
+        tendencies_pr.q_rai[k] += S_qr_evap - S_qs_melt
+        tendencies_pr.q_sno[k] += S_qs_sub_dep + S_qs_melt
 
-        tendencies_pr.q_rai[k] += rain_evap
+        aux_tc.qt_tendency_precip_sinks[k] = -S_qr_evap - S_qs_sub_dep
+        aux_tc.θ_liq_ice_tendency_precip_sinks[k] =
+            1.0 / Π_m / c_pm * (
+                S_qr_evap * (L_v - R_v * T_gm) * (1.0 + R_m / c_vm) +
+                S_qs_sub_dep * (L_s - R_v * T_gm) * (1.0 + R_m / c_vm) +
+                S_qs_melt * L_f * (1.0 + R_m / c_vm)
```
charleskawczynski, line 97:
```suggestion
            1 / Π_m / c_pm * (
                S_qr_evap * (L_v - R_v * T_gm) * (1 + R_m / c_vm) +
                S_qs_sub_dep * (L_s - R_v * T_gm) * (1 + R_m / c_vm) +
                S_qs_melt * L_f * (1.0 + R_m / c_vm)
```

src/microphysics_coupling.jl
```diff
@@ -1,42 +1,101 @@
...
+
         if precipitation_model == "cutoff"
             qsat = TD.q_vap_saturation(ts)
-            qr_tendency = min((q.liq + q.ice) / dt, -CM0.remove_precipitation(param_set, q, qsat))
+            λ = TD.liquid_fraction(ts)
+
+            S_qt = -min((q.liq + q.ice) / dt, -CM0.remove_precipitation(param_set, q, qsat))
+
+            qr_tendency -= S_qt * λ
+            qs_tendency -= S_qt * (1.0 - λ)
```
charleskawczynski, line 30:
```suggestion
            qs_tendency -= S_qt * (1 - λ)
```

src/microphysics_coupling.jl
```diff
@@ -1,42 +1,101 @@
...
 """
 Computes the tendencies to qt and θ_liq_ice due to precipitation formation
+(autoconversion + accretion)
 """
-function precipitation_formation(param_set::APS, precipitation_model, qr, qs, area, ρ0, dt, ts)
+function precipitation_formation(param_set::APS, precipitation_model, qr::FT, qs::FT, area, ρ0, dt, ts) where {FT}
 
+    # TODO - when using adaptive timestepping we are limiting the source terms
+    #        with the previous timestep dt
+    qt_tendency = 0.0
```
charleskawczynski, line 9:
```suggestion
    qt_tendency = FT(0)
```
And similar below

src/microphysics_coupling.jl
```diff
@@ -1,42 +1,101 @@
...
+            # Autoconversion of cloud ice to snow is done with a simplified rate.
+            # The saturation adjustment scheme prevents using the
+            # 1-moment snow autoconversion rate that assumes
+            # that the supersaturation is present in the domain.
+            S_qt_rain = -min(q.liq / dt, CM1.conv_q_liq_to_q_rai(param_set, q.liq))
+            S_qt_snow = -min(q.ice / dt, CM1.conv_q_ice_to_q_sno_no_supersat(param_set, q.ice))
+            qr_tendency -= S_qt_rain
+            qs_tendency -= S_qt_snow
+            qt_tendency += S_qt_rain + S_qt_snow
+            θ_liq_ice_tendency -= 1.0 / Π_m / c_pm * (L_v0 * S_qt_rain + L_s0 * S_qt_snow)
```
charleskawczynski, line 52:
```suggestion
            θ_liq_ice_tendency -= 1 / Π_m / c_pm * (L_v0 * S_qt_rain + L_s0 * S_qt_snow)
```

src/microphysics_coupling.jl
```diff
@@ -1,42 +1,101 @@
...
+            qs_tendency -= S_qt * (1.0 - λ)
+            qt_tendency += S_qt
+            θ_liq_ice_tendency -= S_qt / Π_m / c_pm * (L_v0 * λ + L_s0 * (1 - λ))
         end
-    end
 
-    qt_tendency = -qr_tendency
-    qs_tendency = 0.0
-    θ_liq_ice_tendency = θ_liq_ice_helper(ts, qt_tendency)
+        if precipitation_model == "clima_1m"
```
charleskawczynski, line 35: This is fine for now, but we should start thinking about how we'll change this to using dispatch so that the compiler can do branch elimination

charleskawczynski (approved):
Code changes look fine to me overall. There are some type instabilities that can be cleaned up. The TRMM results do look a bit different, that's alright I assume?

driver/generate_namelist.jl
```diff
@@ -279,14 +279,12 @@ function Rico(namelist_defaults)
     namelist["grid"]["nz"] = 80
     namelist["grid"]["dz"] = 50.0
 
-    # Rico cannot run with adaptive dt for now
-    namelist["time_stepping"]["adapt_dt"] = true
+    namelist["time_stepping"]["adapt_dt"] = false
     namelist["time_stepping"]["t_max"] = 86400.0
-    namelist["time_stepping"]["dt_max"] = 9.0
+    #namelist["time_stepping"]["dt_max"] = 5.0
```
yairchn, line 284: I don't think you need to comment this out if 'adapt_dt = false'

driver/generate_namelist.jl
```diff
@@ -298,14 +296,15 @@ function TRMM_LBA(namelist_defaults)
 
     namelist["meta"]["casename"] = "TRMM_LBA"
 
-    namelist["grid"]["nz"] = 320
-    namelist["grid"]["dz"] = 50.0
+    namelist["grid"]["nz"] = 80
+    namelist["grid"]["dz"] = 200
```
yairchn, line 300: do we wanna keep this low vertical resolution? did you test the higher resolution?

yairchn (approved):

driver/generate_namelist.jl
yairchn: I run a 50m resolution TRMM locally and the results are similar - so I think this is ok

driver/generate_namelist.jl
trontrytel: Right. That was just for me, because I was getting confused which dt is used if we are running without the adaptive dt. But I did not want to delete the whole line.

I'm hoping we can soon make Rico work with the adaptive dt. It's just that CFL callback. Not sure why it's reporting it as false

trontrytel:
> Code changes look fine to me overall. There are some type instabilities that can be cleaned up. The TRMM results do look a bit different, that's alright I assume?

Thank you. I applied your comments to the code. 

Yeah - we expect TRMM to be different, now that we have all the snow/ice related microphysics. As a next step we should update the LES dataset we are comparing with. And then look into actually calibrate some parameters :)

PR #540, opened 2021-11-08
Reduce allocations

charleskawczynski:


integration_tests/utils/main.jl
```diff
@@ -166,6 +166,7 @@ cent_aux_vars_edmf(FT, n_up) = (;
         ),
         term_vel_rain = FT(0),
         term_vel_snow = FT(0),
+        ∂θv∂z = CC.Geometry.Covariant3Vector(FT(0)),
```
charleskawczynski, line 169: @simonbyrne, is this the correct vector type to use?

The results using this configuration deviate quite a bit from our existing implementation, which I think should be an exact translation, AFAICT.

charleskawczynski:
This was kind of superseded by #541.

PR #544, opened 2021-11-09
Reduce allocations, use ClimaCore operators

charleskawczynski:
This took a lot of experimentation, but I think I figured out something that is roughly equivalent.


charleskawczynski:
Results are virtually unchanged 🎉

PR #548, opened 2021-11-09
add adaptive timestep callback

yairchn:
lets see how fast /slow is this PR


yairchn:
There is an issue with such PR that the dt that is used in 'compute_precipitation_sink_tendencies' I snot updated accordingly with the adaptive tilmestep - but rather taken from the nameless (effectively the maximum dt) 
@trontrytel @charleskawczynski  - not sure how bad it is / what could be done otherwise

charleskawczynski (commented):
The results look good, but it is still taking too long.

ilopezgp:
Thanks Yair! Ouch, 51 minutes. So, the results that we have right now are beyond the CFL -- or is the CFL computation that expensive?

trontrytel:
> There is an issue with such PR that the dt that is used in 'compute_precipitation_sink_tendencies' I snot updated accordingly with the adaptive tilmestep - but rather taken from the nameless (effectively the maximum dt) @trontrytel @charleskawczynski - not sure how bad it is / what could be done otherwise

It should be limited by the dt actually used. 

Do you know what timestep we end up with? Is it very similar to the one that was there already?

yairchn:
> Thanks Yair! Ouch, 51 minutes. So, the results that we have right now are beyond the CFL -- or is the CFL computation that expensive?

well, we are using OrdinaryDiffEQs here - which I think is still a slowdown in the simulation time + the dt is not always larger than the fix value. It is definitely the case in the beginning of each simulation when it is sometimes x10 the fix dt for the case (like TRMM) but later it can be much smaller than the fix dt (i.e. the fix dt was not meeting stability criteria but did not break the simulations after all). 
I am working on adding a minimum dt as well

yairchn:
> It should be limited by the dt actually used.

@charleskawczynski this would require the dt from DiffEQs to be available in the microphysics tendencies - is that  something we can (or would like) to do?

> Do you know what timestep we end up with? Is it very similar to the one that was there already?

In a simulation like TRMM dt is 10sec for the first 40% of the simulation but can be 0.9 sec in the last hour

charleskawczynski:
> @charleskawczynski this would require the dt from DiffEQs to be available in the microphysics tendencies - is that something we can (or would like) to do?

I'm not sure. Also, wouldn't that limit allowable timestepping method? What's the time-continuous equations that we're trying to model?

trontrytel:
> > @charleskawczynski this would require the dt from DiffEQs to be available in the microphysics tendencies - is that something we can (or would like) to do?
> 
> I'm not sure. Also, wouldn't that limit allowable timestepping method? What's the time-continuous equations that we're trying to model?

Yes, in principle this introduces additional limit on dt. In practice you don't really want to do it this way, as you will end up with very small time steps. 

A good mathematic overview of the problem is here: https://gmd.copernicus.org/articles/5/975/2012/

charleskawczynski:
> Yes, in principle this introduces additional limit on dt

I don't mean a limit on dt, I mean the algorithm (Explicit Euler, RK4, etc.). It seems like this is just a heuristic anyway, can't we use a different measure that doesn't depend on dt?

trontrytel:
> > Yes, in principle this introduces additional limit on dt
> 
> I don't mean a limit on dt, I mean the algorithm (Explicit Euler, RK4, etc.). It seems like this is just a heuristic anyway, can't we use a different measure that doesn't depend on dt?
> 
> Also, I'm fine with adding limiters / filters if we need

Let's talk about this on Thursday. The paper I mention above is a good intro to the problem.

yairchn:
superseded by #571

yairchn:
superseded by #571

PR #549, opened 2021-11-09
Move purely diagnostic fields outside of src

charleskawczynski:
This PR moves the purely diagnostic computations outside of `src/`. This will effectively prevent #468 from reemerging (since we won't have access to these fields inside `src/`).

This is going to cause a bunch of conflicts with the ODE branch, but I think that we can perhaps do that a bit more incrementally.

### DIFF
--- a/diagnostics/compute_diagnostics.jl
+++ b/diagnostics/compute_diagnostics.jl
@@ -1,0 +1,212 @@
+# TODO: should this live in its own module?
+
+""" Purely diagnostic fields for the host model """
+diagnostics(state, fl) = getproperty(state, TC.field_loc(fl))
+
+center_diagnostics_grid_mean(state) = diagnostics(state, TC.CentField())
+center_diagnostics_turbconv(state) = diagnostics(state, TC.CentField()).turbconv
+face_diagnostics_turbconv(state) = diagnostics(state, TC.FaceField()).turbconv
+
+#=
+    io_dictionary_diagnostics()
+
+These functions return a dictionary whose
+ - `keys` are the nc variable names
+ - `values` are NamedTuples corresponding to
+    - `dims` (`("z")`  or `("z", "t")`) and
+    - `group` (`"reference"` or `"profiles"`)
+
+This dictionary is for purely diagnostic quantities--which
+are not required to compute in order to run a simulation.
+=#
+
+#! format: off
+function io_dictionary_diagnostics(diagnostics)
+    DT = NamedTuple{(:dims, :group, :field), Tuple{Tuple{String, String}, String, Any}}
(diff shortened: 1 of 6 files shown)


charleskawczynski:
I don't _think_ this should cause problems for the pipeline, but since we're moving things outside of `src`, I'd consider these breaking changes for sure.

integration_tests/utils/compute_diagnostics.jl
```diff
@@ -0,0 +1,212 @@
+# TODO: should this live in its own module?
```
ilopezgp, line 1: I am ok with moving them out of src, but I think diagnostics should not be inside integration_tests. Maybe `diagnostics/` would be better.

ilopezgp: I think of everything in `integration_tests/utils` as utils that enable testing the SCM in specific configurations, and comparing them to LES. But pure SCM diagnostics should live somewhere else.

charleskawczynski: Yeah, good point.

integration_tests/utils/compute_diagnostics.jl
```diff
@@ -0,0 +1,212 @@
...
+
+    @inbounds for i in 1:(up.n_updrafts)
+        up.cloud_base[i] = TC.zc_toa(grid)
+        up.cloud_top[i] = 0.0
+        up.updraft_top[i] = 0.0
+        up.cloud_cover[i] = 0.0
+
+        @inbounds for k in TC.real_center_indices(grid)
+            if aux_up[i].area[k] > 1e-3
+                up.updraft_top[i] = max(up.updraft_top[i], grid.zc[k])
```
yairchn, line 120: 'updraft_top' is used in the pressure closure and should be part of /src

integration_tests/utils/compute_diagnostics.jl
charleskawczynski: I’ll start a diagnostics folder

integration_tests/utils/compute_diagnostics.jl
charleskawczynski: Ah, yeah. Good catch. I'll move this in another PR

charleskawczynski: Opened an issue in #550.

integration_tests/utils/compute_diagnostics.jl
charleskawczynski: Ok, moved this to `TC.jl/diagnostics/`

yairchn (approved):
LGTM

PR #557, opened 2021-11-10
Add adaptive dt for ODE.solve

yairchn:
This PR adds an adaptive dt for the time stepping. 
Adaptive dt for now is not applicable for cases with any precipitate, as dt is used in the precipitate tendencies. For that I added a  (Bool) nameless flag for "adapt_dt".
To do so it adds a 'dt_min' for the nameless arguments that are case specific.
In the current format the "dt" argument is served as an initial (maybe maximum) dt and the case specific "dt_min" are used both as the minimum allowed dt and as default value of timestep for the various cases if the flag "adapt_dt" is set to false.
Other values of "cfl_limit" were tested and the current one (0.5) was found to work across all cases.


yairchn:
superseded by #571

PR #558, opened 2021-11-10
Use ClimaCore operators, reduce allocs

charleskawczynski:


charleskawczynski:

PR #546, opened 2021-11-09
Reduce allocations

charleskawczynski:

PR #559, opened 2021-11-11
Reduce allocations in MinDisspLen

charleskawczynski:
I'm hoping that this will supersede #546.

### DIFF
--- a/src/update_aux.jl
+++ b/src/update_aux.jl
@@ -335,6 +335,24 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
     #####
     ##### compute_eddy_diffusivities_tke
     #####
+
+    # Subdomain exchange term
+    Ic = CCO.InterpolateF2C()
+    b_exch = center_aux_turbconv(state).b_exch
+    parent(b_exch) .= 0
+    a_en = aux_en.area
+    w_en = aux_en_f.w
+    tke_en = aux_en.tke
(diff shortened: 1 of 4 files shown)


yairchn:
there is some weird issue in dycoms - I will try to look better tomorrow

src/update_aux.jl
```diff
@@ -336,6 +336,22 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
     ##### compute_eddy_diffusivities_tke
     #####
 
+    # Subdomain exchange term
+    Ic = CCO.InterpolateF2C()
+    b_exch = center_aux_turbconv(state).b_exch
+    a_en = aux_en.area
+    w_en = aux_en_f.w
+    tke_en = aux_en.tke
+    for i in 1:(up.n_updrafts)
```
yairchn, line 345: shouldn't b_exch be set to zero before the loop over updrafts ?

yairchn: adding '@. b_exch .*= 0' in line 343 fixed the problem

charleskawczynski: Yeah, good catch!

yairchn (approved):
LGTM - some table updates are needed but the results are qualitatively unchanged

charleskawczynski:
Hmm, not sure what's going on here, this PR actually _hurt_ performance, despite resulting in far less allocations per tendencies evaluations. We may want to review this or look into it further.

PR #561, opened 2021-11-11
Use ClimaCore operators

charleskawczynski:
This is kind of a prep PR to help fix the type instabilities that lead up to creating `MinDisspLen`.


charleskawczynski:

PR #562, opened 2021-11-11
Make EDMF model more concrete

charleskawczynski:
This PR makes the `EDMF_PrognosticTKE` add a type parameter for `EnvBuoyGradClosure`, to fix the type instability. This was causing a cascade of type inference failures leading up to computing `MinDisspLen`.


yairchn (approved):
LGTM

charleskawczynski:
MSE tables were changed because, I think, we've auto-updated our deps with a new patch release which actually had behavioral changes. Results look virtually unchanged.

PR #563, opened 2021-11-12
Add masked operators, remove vanish sd grad

charleskawczynski:
This PR removes the `c∇_vanishing_subdomain` operators and adds `masked_interpolate!` which, together with ClimaCore operators, can serve the same purpose.

This was the quick and dirty way to use the existing infrastructure, but it's certainly not performant, and needs to be revisited.


charleskawczynski:
Superseded by #565.

PR #564, opened 2021-11-12
Make more variables climacore fields

charleskawczynski:
This was peeled off from #563. This should help to locally check to see the difference between the old/new operators.


charleskawczynski:
Ugh, bors crashed, need to delete the folder that was moved into main...

charleskawczynski:
bors ping

PR #565, opened 2021-11-12
Replace vanishing subdomain gradients with ClimaCore operators

charleskawczynski:
Supersedes #563. I'm keeping #563 open for reference, so that this branch can be merged as a single commit.


src/update_aux.jl
```diff
@@ -361,6 +360,24 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
...
     @. ∂θv∂z = ∇c(wvec(If(θ_virt_en)))
 
+    # Second order approximation: Use dry and cloudy environmental fields.
+    cf = aux_en.cloud_fraction
+    shm = copy(cf)
+    parent(shm) .= shrink_mask(vec(cf))
+
+    # Since NaN*0 ≠ 0, we need to conditionally replace
+    # our gradients by their default values.
+    @. ∂qt∂z_sat = ∇c(wvec(If(aux_en_sat.q_tot)))
```
ilopezgp, line 370: We only have to compute these if we are using `BuoyGradQuadratures()`, so we can move them within that clause.

charleskawczynski: I’m cleaning up this whole block of code to reduce allocations, so I’ll do that in a later PR

PR #568, opened 2021-11-13
Try to improve type inference

charleskawczynski:
This PR passes `n_updrafts` as a type parameter so that we can potentially unroll some updraft loops. In addition, we avoid calling `ntuple` where it's not necessarily needed.


yairchn (approved):
LGTM

PR #571, opened 2021-11-17
Yc/ode adaptive dt local

yairchn:
I am adding here a fast adaptive dt PR that should allow calibration


integration_tests/utils/generate_namelist.jl
```diff
@@ -203,7 +206,7 @@ function Soares(namelist_defaults)
     namelist["grid"]["dz"] = 50.0
 
     namelist["time_stepping"]["t_max"] = 8 * 3600.0
-    namelist["time_stepping"]["dt"] = 1.0
+    namelist["time_stepping"]["dt_min"] = 1.0
```
ilopezgp, line 209: Shouldn't all these by `dt_max`, if anything? Otherwise, you are never going slower than the current default.

yairchn: "dt_max" in the namelist is the minimum dt the code allows to prevent too small time steps - it is case specific. 
"dt_max" in the namelist is the maximum dt and is not case specific for now.
 
The actual time step 'TS.dt' is bounded from below by "max(edmf.dt_max, dt_min)" and from above by "TS.dt_max" in the adaptive teamster function in main:
" TS.dt = min(TS.dt_max, max(edmf.dt_max, dt_min)) ".

The lower bound: "max(edmf.dt_max, dt_min)" is the maximum of the dt given by maximum velocity scale in the model and the smallest dt allowed from the namelist.

ilopezgp: When you say dt_max in the namelist, I suppose you mean `dt_min`? There are no `dt_max` in the namelist right now. Anyway, what I wanted to say is that the current `dt_min` from the name lists are too large, they are the default we had before.

charleskawczynski (commented):
Code changes look fine to me. It does look like the TRMM build up is a bit less smooth, but the results overall look qualitatively fine.

Is there an easy way to switch this adaptive time stepping off and use a fixed timestep? Maybe it would be nice to add in the description how to do that. Also, feel free to remove (comment out) one of the cases from the allocation tests, [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/perf/allocs.jl#L10-L24), so that the overall CI time is reduced.

ilopezgp:
We could have `namelist_defaults["time_stepping"]["adapt_dt"]=False` as the default before merging, but let's leave it True until then to make sure simulations look ok.

ilopezgp:
Not that this is a practical concern now, but it would be good to enforce that the maximum time step for all simulations is at most the diagnostic output period.

integration_tests/utils/generate_namelist.jl
trontrytel: > "dt_max" in the namelist is the minimum dt the code allows to prevent too small time steps - it is case specific.

Thats supposed to be for `dt_min`? We will still error out if the desired `dt_min` is too large?

src/EDMF_Precipitation.jl
```diff
@@ -31,6 +31,8 @@ function compute_precipitation_advection_tendencies(grid, state, gm, TS::TimeSte
             CFL_in_rain = 0.0
             CFL_in_snow = 0.0
         else
+            vel_max = max(term_vel_rain[k+1], term_vel_snow[k+1])
```
trontrytel, line 34: We compute the CFL criterion based on the flux in and out. Shouldn't there be something similar done in lines 28-29 here?

src/update_aux.jl
```diff
@@ -192,6 +192,7 @@ function update_aux!(edmf, gm, grid, state, Case, param_set, TS)
                 anew_k = interpc2f(aux_up[i].area, grid, k; a_up_bcs...)
                 if anew_k >= edmf.minimum_area
                     aux_up_f[i].w[k] = max(prog_up_f[i].ρaw[k] / (ρ0_f[k] * anew_k), 0)
+                    edmf.dt_max = min(edmf.dt_max, TS.cfl_limit * grid.Δz / (abs(aux_up_f[i].w[k]) + 0.001))
```
trontrytel, line 195: Why do we add 0.001 to the denominator here? Is it to prevent divisions by zero?

trontrytel:
Another question is whether we should do something smarter with our output if we start using the adaptive timestep. I think the current assumption is that our model dt is reasonably big. If our dt gets very small we might start outputting more often than we shoudl?

`    condition_io(u, t, integrator) = mod(round(Int, t), round(Int, sim.Stats.frequency)) == 0
`

ilopezgp:
> Another question is whether we should do something smarter with our output if we start using the adaptive timestep. I think the current assumption is that our model dt is reasonably big. If our dt gets very small we might start outputting more often than we shoudl?
> 
> ` condition_io(u, t, integrator) = mod(round(Int, t), round(Int, sim.Stats.frequency)) == 0`

I think output frequency should be fixed and decoupled from the timestep. Otherwise, things will get messy when analyzing covariances, etc, of the diagnostics.

trontrytel:
> > Another question is whether we should do something smarter with our output if we start using the adaptive timestep. I think the current assumption is that our model dt is reasonably big. If our dt gets very small we might start outputting more often than we shoudl?
> > ` condition_io(u, t, integrator) = mod(round(Int, t), round(Int, sim.Stats.frequency)) == 0`
> 
> I think output frequency should be fixed and decoupled from the timestep. Otherwise, things will get messy when analyzing covariances, etc, of the diagnostics.

What I'm saying is that if out model timestep is small we will be outputting here many times instead of just once. (If I understand the condition right)

ilopezgp:
> > > Another question is whether we should do something smarter with our output if we start using the adaptive timestep. I think the current assumption is that our model dt is reasonably big. If our dt gets very small we might start outputting more often than we shoudl?
> > > ` condition_io(u, t, integrator) = mod(round(Int, t), round(Int, sim.Stats.frequency)) == 0`
> > 
> > 
> > I think output frequency should be fixed and decoupled from the timestep. Otherwise, things will get messy when analyzing covariances, etc, of the diagnostics.
> 
> What I'm saying is that if out model timestep is small we will be outputting here many times instead of just once. (If I understand the condition right)

Oh I see. You are right, this condition only works for time steps larger than 1s. Otherwise it prints every time, good catch! We need to change this.

trontrytel:
> Oh I see. You are right, this condition only works for time steps larger than 1s. Otherwise it prints every time, good catch! We need to change this.

Right. With time step that are non-integer and possibly small we should probably have a condition for "the first time we are greater than what the output frequency dictates". We won't be able to have exact match here

ilopezgp:
> > Oh I see. You are right, this condition only works for time steps larger than 1s. Otherwise it prints every time, good catch! We need to change this.
> 
> Right. With time step that are non-integer and possibly small we should probably have a condition for "the first time we are greater than what the output frequency dictates". We won't be able to have exact match here

Another option is to have the internal clock for diagnostics in ms or ns (just multiply everything by 10^3 or 10^6 in `condition_io`). That way we do not have to keep a counter in memory.

src/update_aux.jl
yairchn: yes

trontrytel: Could this be done with some standard epsilon instead? Looks like a hack. Plus, if the velocity is zero, then we don't really have to worry about CFL?

yairchn: it was added to prevent zero division, I replaced it by 'eps(FT)'

integration_tests/utils/generate_namelist.jl
yairchn: sorry I mean "dt_min" in the namelist is the minimum dt the code allows to prevent too small time steps

integration_tests/utils/generate_namelist.jl
```diff
@@ -401,8 +414,8 @@ function LES_driven_SCM(namelist_defaults)
     namelist["grid"]["nz"] = 200
 
     namelist["stats_io"]["frequency"] = 10.0
-    namelist["time_stepping"]["dt"] = 3.0
     namelist["time_stepping"]["t_max"] = 3600.0 * 6
+    namelist["time_stepping"]["dt_min"] = 3.0
```
ilopezgp, line 418: Can you reduce this to 1.0 s?

yairchn:
@trontrytel @ilopezgp I revise the condition_io to obey the frequency from nameless regardless of dt. 
I think the PR is ready

yairchn:
please notice the oscillations in BOMEX - they appear with the change in the condition_io, trying to find out why happy for suggestions

src/EDMF_Precipitation.jl
yairchn: the revised condition_io fixes the issue - thanks

ilopezgp (approved):
LGTM

ilopezgp:
Thanks @yairchn !

src/EDMF_Precipitation.jl
```diff
@@ -32,6 +32,8 @@ function compute_precipitation_advection_tendencies(grid, state, gm, TS::TimeSte
             CFL_in_rain = 0.0
             CFL_in_snow = 0.0
         else
+            vel_max = max(term_vel_rain[k + 1], term_vel_snow[k + 1])
+            edmf.dt_max = min(edmf.dt_max, CFL_limit * Δz / (vel_max + eps(Float32)))
```
charleskawczynski, line 36: Can we move these computations into one of the callbacks (or add a new callback)? I’m fine if you want to just do this in a subsequent PR, though

src/update_aux.jl
```diff
@@ -451,6 +453,7 @@ function update_aux!(edmf::EDMF_PrognosticTKE{N_up}, gm, grid, state, Case, para
 
         KM[k] = c_m * ml.mixing_length * sqrt(max(aux_en.tke[k], 0.0))
         KH[k] = KM[k] / aux_tc.prandtl_nvec[k]
+        edmf.dt_max = min(edmf.dt_max, TS.cfl_limit * grid.Δz^2 / (max(KH[k], KM[k]) + eps(Float32)))
```
charleskawczynski, line 456: These should be defined in the callback

src/EDMF_Precipitation.jl
yairchn: I think the computation cost would be large enough to make the adaptive time step inefficient

src/update_aux.jl
yairchn: I think the computation cost would be large enough to make the adaptive time step inefficient

src/EDMF_Precipitation.jl
charleskawczynski: We can always adjust the frequency of the callback. Isn’t this currently run at every time step?

yairchn: currently every timestep, I had issues running the original PR with adaptive dt every x timestep if x>2

yairchn: I think there is cost difference, here we are already in a k-loop and dt_max is not indexed (its a single value). In a callback I would be looping on k and indexing 5 different variables

charleskawczynski: I think the extra column traversals are fine for now.

charleskawczynski (approved):
Overall, LGTM

yairchn:
Bors r+

Project.toml
```diff
@@ -58,7 +59,6 @@ PrettyTables = "1.2"
 RootSolvers = "0.2"
 StaticArrays = "1.2"
 StatsBase = "0.33"
-SurfaceFluxes = "0.1"
```
charleskawczynski, line 61: I don't think we need to remove this..

PR #572, opened 2021-11-17
Replace more operators with ClimaCore operators

charleskawczynski:
This PR removes some old unused operators and replaces some operators with ClimaCore operators for the `perturbation_pressure` function, which has been split up into 3 methods: `nh_pressure_buoy`,`nh_pressure_adv`,`nh_pressure_drag`.


charleskawczynski:
Results look virtually unchanged.

PR #577, opened 2021-11-19
Use more ClimaCore operators

charleskawczynski:


src/Turbulence_PrognosticTKE.jl
```diff
@@ -867,18 +859,18 @@ function compute_en_tendencies!(
         entr_sc = aux_up[i].entr_sc
         w_up = aux_up_f[i].w
         a_up = aux_up[i].area
+        # TODO: using `Int(bool) *` means that NaNs can propagate
+        # into the solution. Could we somehow call `ifelse` instead?
```
jakebolewski, line 863: @charleskawczynski `ifelse` should broadcast correctly as it eagerly evaluates both arguments:

```
@. ifelse(a <= min_area, zero(FT), ρ0_c * a_up * Ic(w_up) * (entr_sc + turb_entr))
```

charleskawczynski: I'll do this in a separate PR since there's a couple other places where we can use this.

PR #588, opened 2021-11-22
update bc for precip, fix diagnostics

trontrytel:
This is a peel-off from my debugging of ice and snow in TRMM.

It changes the bottom bc for precipitation to zero gradient, adds ice to updraft and environment diagnostics and adds rain and snow to CI plots in TRMM


charleskawczynski (approved):
LGTM

PR #589, opened 2021-11-30
Use ClimaCore operators

charleskawczynski:
Use more ClimaCore operators, in `compute_covariance_entr`

### DIFF
--- a/src/Turbulence_PrognosticTKE.jl
+++ b/src/Turbulence_PrognosticTKE.jl
@@ -694,12 +694,12 @@ function compute_covariance_entr(
     grid,
     state,
     ::Val{covar_sym},
-    v1::Val{var1},
-    v2::Val{var2} = v1,
-) where {covar_sym, var1, var2, N_up}
+    ϕs::Val{ϕ_sym},
+    ψs::Val{ψ_sym} = ϕs,
+) where {covar_sym, ϕ_sym, ψ_sym, N_up}
 
     ρ0_c = center_ref_state(state).ρ0
-
+    FT = eltype(grid)
     is_tke = covar_sym == :tke
     tke_factor = is_tke ? 0.5 : 1
     aux_up = center_aux_updrafts(state)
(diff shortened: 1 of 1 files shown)


src/Turbulence_PrognosticTKE.jl
```diff
@@ -707,56 +707,51 @@ function compute_covariance_entr(
...
+                ρ0_c *
+                a_up *
+                abs(Ic(w_up)) *
+                eps_turb *
+                ((Idc(ϕ_en) - Idc(ϕ_gm)) * (Idc(ψ_up) - Idc(ψ_en)) + (Idc(ψ_en) - Idc(ψ_gm)) * (Idc(ϕ_up) - Idc(ϕ_en)))
+            ),
+            0,
+        )
+        @. detr_loss +=
+            ifelse(a_up > edmf.minimum_area, tke_factor * ρ0_c * a_up * abs(Ic(w_up)) * (entr_sc + eps_turb) * covar, 0)
```
jakebolewski, line 753: the return type needs to be the same so I would use `zero(...)`

src/Turbulence_PrognosticTKE.jl
```diff
@@ -707,56 +707,51 @@ function compute_covariance_entr(
...
-                    abs(w_u) *
-                    eps_turb *
-                    ((envvar1 - gmvvar1) * (updvar2 - envvar2) + (envvar2 - gmvvar2) * (updvar1 - envvar1))
-                entr_gain[k] += dynamic_entr + turbulent_entr
-                detr_loss[k] += tke_factor * ρ0_c[k] * a_up * abs(w_u) * (entr_sc[k] + eps_turb) * covar[k]
-            end
-        end
+        ϕ_up = getproperty(prog_up_i, ϕ_sym)
+        ψ_up = getproperty(prog_up_i, ψ_sym)
+        parent(entr_gain) .= 0
```
jakebolewski, line 735: we should fix this so `parent` doesn't have to be called, it would be good to leave a todo

PR #592, opened 2021-11-30
remove the stability check of LES_driven_SCM in the ode solver

yairchn:
It seems that 'unstable_check_kwarg(::Cases.LES_driven_SCM)' is not needed anymore

### DIFF
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -354,19 +354,12 @@ function solve_args(sim::Simulation1d)
         alg = ODE.Euler()
     end
 
-    # TODO: LES_driven_SCM is currently unstable w.r.t. higher order moments (HOM).
-    # So, we tell OrdinaryDiffEq.jl to not perform NaNs check on the solution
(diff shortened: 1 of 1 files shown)


charleskawczynski:
I think we really don't need the unstable kwarg check, since all this really does is error if our higher order moments become NaNs. Our MSE check already tests the tke, if the other HOM are important, then we should add them to the mse tables..

charleskawczynski (commented):
I think we really don't need the unstable kwarg check, since all this really does is error if our higher order moments become NaNs. Our MSE check already tests the tke, if the other HOM are important, then we should add them to the mse tables..

charleskawczynski:
I think that we can actually revive this now that we have these tests in place:
```julia
@testset "Simulation completion" begin
    # Test that the simulation has actually finished,
    # and not aborted early.
    @test !(return_code == :simulation_aborted)
    @test return_code == :success
end
```

charleskawczynski:
Let's try at least

PR #593, opened 2021-11-30
Try to improve inference

charleskawczynski:
This PR tries to improve inference by improving inference, and fixes some variable names to conform to our naming conventions.

### DIFF
--- a/src/Turbulence_PrognosticTKE.jl
+++ b/src/Turbulence_PrognosticTKE.jl
@@ -612,10 +612,10 @@ function compute_covariance_shear(
     grid,
     state,
     gm::GridMeanVariables,
-    covar_sym::Symbol,
-    en_var1_sym::Symbol,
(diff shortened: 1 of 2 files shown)


jakebolewski:
It would be better to test these with `JET`, Val can be useful but has a large cost.

charleskawczynski:
> It would be better to test these with `JET`, Val can be useful but has a large compile time cost. If you are using parameterized Val{symbol} for dispatch / specialization then it might be cleaner to refactor them into types.

Yeah, I agree that it'd be ideal to use JET / unit tests, I figured we'll wait until 1.7 is released and we can incorporate it into our CI.

I also agree about avoiding the use of Val, it's just the simplest thing for now (to see if these are in fact the source of allocations) since there's not a ton of combinations, and we're using these symbols with `getproperty`.

PR #599, opened 2021-12-01
Hoist getproperty outside of broadcast, move zero-out before up loop

charleskawczynski:


src/Turbulence_PrognosticTKE.jl
```diff
@@ -733,30 +736,22 @@ function compute_covariance_entr(
...
-                eps_turb *
-                ((Idc(ϕ_en) - Idc(ϕ_gm)) * (Idc(ψ_up) - Idc(ψ_en)) + (Idc(ψ_en) - Idc(ψ_gm)) * (Idc(ϕ_up) - Idc(ϕ_en)))
-            ),
-            zero(FT),
-        )
-        @. detr_loss += ifelse(
-            a_up > edmf.minimum_area,
-            tke_factor * ρ0_c * a_up * abs(Ic(w_up)) * (entr_sc + eps_turb) * covar,
-            zero(FT),
+        @. entr_gain += Int(a_up > edmf.minimum_area)
```
jakebolewski, line 742: I don't think you need to convert to Int, (true / false) * (some floating point expression value) should convert to Float 0.0 / 1.0

charleskawczynski: Yeah, it's not necessary. I'm going to leave it this way for now (for search-ability) since we're applying this same pattern in several places. I think using `ifelse` is the long term solution since it can work nicely with NaNs

charleskawczynski: Opened #603

PR #601, opened 2021-12-01
Time solve

charleskawczynski:
It would be nice to time the `solve` method so that, as we reduce allocations, we can get validation from `@time` that allocations are being reduced.


ilopezgp:
This would also be a great way to compare different solvers (e.g., Euler, RK4), since they have different CFL conditions and different evals-per-dt.

PR #602, opened 2021-12-01
add entr closure, specify entr type in namelist

costachris:
This PR: 
       1.) Adds an entrainment closure type (small neural network)
       2.) Makes the `entrainment` namelist argument functional, so that it's easy to switch between potential entrainment closures.
       3.) Refactors components of  `entry_detr.jl`


costachris:
The implemented NN is not yet calibrated since I added some inputs and changed the architecture, but it runs. To switch between entrainment types use the `entrainment` namelist arg.

Project.toml
```diff
@@ -16,6 +16,7 @@ DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 FastGaussQuadrature = "442a2c76-b920-505d-bb47-c5924d526838"
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
```
charleskawczynski, line 19: Let's add `Flux` to the `test/` environment, too.

costachris: It's added in ` test/Project.toml`, or did you mean something else?

charleskawczynski: Ah, yes, sorry, I didn't see it below

haakon-e (approved):
Great work!

PR #605, opened 2021-12-02
Switch cases for allocation monitoring

charleskawczynski:


charleskawczynski:
I think there's really no need for this, actually. Most allocations are quite small at this point. Time stepping is the primary bottleneck

PR #606, opened 2021-12-02
Use more ClimaCore operators

charleskawczynski:
This PR adds more use of ClimaCore operators, and removes `∇f2c`.


charleskawczynski:

PR #611, opened 2021-12-02
Move CFL computations to callback

charleskawczynski:


integration_tests/utils/main.jl
```diff
@@ -432,6 +437,52 @@ function adaptive_dt!(integrator)
     ODE.u_modified!(integrator, false)
 end
 
+function monitor_cfl!(integrator)
+    UnPack.@unpack gm, grid, edmf, aux, params, TS = integrator.p
+    state = TC.State(integrator.u, aux, integrator.du)
+    param_set = TC.parameter_set(gm)
+    FT = eltype(grid)
+    Δz = grid.Δz
+    Δt = TS.dt
```
trontrytel, line 446: This `TS.dt` would be equal to dt from the previous time step? If yes, then this check will not work.

integration_tests/utils/main.jl
```diff
@@ -432,6 +437,52 @@ function adaptive_dt!(integrator)
...
+    # TODO: assuming gm.W = 0
+    # TODO: verify translation
+    term_vel_rain = aux_tc.term_vel_rain
+    term_vel_snow = aux_tc.term_vel_snow
+
+    @inbounds for k in TC.real_center_indices(grid)
+        term_vel_rain[k] = CM1.terminal_velocity(param_set, CM1.RainType(), ρ0_c[k], prog_pr.q_rai[k])
+        term_vel_snow[k] = CM1.terminal_velocity(param_set, CM1.SnowType(), ρ0_c[k], prog_pr.q_sno[k])
+        vel_max = max(term_vel_rain[k], term_vel_snow[k])
+        edmf.dt_max = min(edmf.dt_max, CFL_limit * Δz / (vel_max + eps(Float32)))
```
trontrytel, line 464: I think we should keep the adaptive timestep computation outside of the `monitor_cfl` function

integration_tests/utils/main.jl
```diff
@@ -432,6 +437,52 @@ function adaptive_dt!(integrator)
...
+    end
+
+    @inbounds for k in TC.real_center_indices(grid)
+        # check stability criterion
```
trontrytel, line 468: What I was trying to solwe in my PR was that
1. In case of fixed timestep the delta t here would be the user defined timestep
2. In case of the adaptive timestep the delta t would be the timestep limited by the CFL criterion for precipitation. (Not necessarily the current timestep, as some processes further down could still limit the dt further)

src/EDMF_Precipitation.jl
```diff
@@ -19,29 +15,6 @@ function compute_precipitation_advection_tendencies(edmf, grid, state, gm, TS::T
     term_vel_rain = aux_tc.term_vel_rain
     term_vel_snow = aux_tc.term_vel_snow
 
-    @inbounds for k in real_center_indices(grid)
-        term_vel_rain[k] = CM1.terminal_velocity(param_set, rain_type, ρ0_c[k], prog_pr.q_rai[k])
-        term_vel_snow[k] = CM1.terminal_velocity(param_set, snow_type, ρ0_c[k], prog_pr.q_sno[k])
-        vel_max = max(term_vel_rain[k], term_vel_snow[k])
```
trontrytel, line 25: We could move the terminal velocity computation to update aux and delete it from here?

integration_tests/utils/main.jl
trontrytel: Or in general, is there a way for us to know if all dt limiting was already done? For example right now one of the limiting criterion is just below and it updates edmf.dt but not the TS.dt that is being used in the CFL check here

integration_tests/utils/main.jl
charleskawczynski: Fine by me, I just moved everything from the tendency function into one callback, but we can certainly shuffle things around

integration_tests/utils/main.jl
charleskawczynski: > This TS.dt would be equal to dt from the previous time step? If yes, then this check will not work.

It's the timestep at the next timestep but, the good news is, so is the state, so everything is in sync.

> Or in general, is there a way for us to know if all dt limiting was already done? For example right now one of the limiting criterion is just below and it updates edmf.dt but not the TS.dt that is being used in the CFL check here

The callbacks are called in order, after `step!` is called, and before the update. I think all we need to do is make sure the monitor CFL callback is called before the set dt callback (which sets dt for the next timestep). Here's a MWE to demo the order of the calls:

```julia
import OrdinaryDiffEq as ODE
condition_every_iter(u, t, integrator) = true;

step!(du, u, p, t) = (du[1] = 1; println("Step!, u = $(u)");)
cb1!(integrator) = (
    integrator.u[1]+=1;
    println("cb1, u = $(integrator.u)");
    ODE.u_modified!(integrator, false)
);
cb2!(integrator) = (
    integrator.u[1]+=1;
    println("cb2, u = $(integrator.u)");
    ODE.u_modified!(integrator, false)
);

cb1 = ODE.DiscreteCallback(condition_every_iter, cb1!; save_positions = (false, false));
cb2 = ODE.DiscreteCallback(condition_every_iter, cb2!; save_positions = (false, false));
params = (;);
FT = Float64;
u = FT[0];
Δt = FT(0.3);
nsteps = 3;
t_span = (FT(0), nsteps*Δt);
prob = ODE.ODEProblem(step!, u, t_span, params; dt = Δt);
sol = ODE.solve(
    prob,
    ODE.Euler();
    progress_steps = 1,
    save_start = false,
    saveat = last(t_span),
    callback = ODE.CallbackSet(cb1, cb2),
    progress = true,
    progress_message = (dt, u, p, t) -> t,
);

#=
Step!, u = [0.0] # for/during initializeation
Step!, u = [0.3]
cb1, u = [1.3]
cb2, u = [2.3]
Step!, u = [2.5999999999999996]
cb1, u = [3.5999999999999996]
cb2, u = [4.6]
Step!, u = [4.8999999999999995]
cb1, u = [5.8999999999999995]
cb2, u = [6.8999999999999995]
=#
```

src/EDMF_Precipitation.jl
charleskawczynski: Oh, yes, good catch, thanks!

PR #615, opened 2021-12-03
Reorganize folders

charleskawczynski:
We're preparing to move dycore code outside of `src/`, and decided it would be good to reorganize some of the folders


charleskawczynski:
This will probably cause several git conflicts, sorry! 😬 cc @ilopezgp, @trontrytel, @yairchn

PR #620, opened 2021-12-04
Move dt_max computations into callback

charleskawczynski:
This PR moves all of the `dt_max` computations into a single callback


driver/main.jl
```diff
@@ -442,29 +442,58 @@ function adaptive_dt!(integrator)
...
+        TC.is_surface_face(grid, k) && continue
+        @inbounds for i in 1:N_up
+            dt_max = min(dt_max, CFL_limit * Δz / (abs(aux_up_f[i].w[k]) + eps(Float32)))
+        end
+        dt_max = min(dt_max, CFL_limit * Δz / (abs(aux_en_f.w[k]) + eps(Float32)))
+    end
     @inbounds for k in TC.real_center_indices(grid)
         vel_max = max(term_vel_rain[k], term_vel_snow[k])
-        edmf.dt_max = min(edmf.dt_max, CFL_limit * Δz / (vel_max + eps(Float32)))
+        dt_max = min(dt_max, CFL_limit * Δz / (vel_max + eps(Float32)))
```
trontrytel, line 475: Maybe it would be good to leave a comment briefly describing what conditions we are checking for?

driver/main.jl
```diff
@@ -529,10 +558,13 @@ function run(sim::Simulation1d)
     callback_io = ODE.DiscreteCallback(condition_io, affect_io!; save_positions = (false, false))
     callback_cfl = ODE.DiscreteCallback(condition_every_iter, monitor_cfl!; save_positions = (false, false))
     callback_cfl = sim.Turb.Precip.precipitation_model == "clima_1m" ? (callback_cfl,) : ()
+    callback_dtmax = ODE.DiscreteCallback(condition_every_iter, dt_max!; save_positions = (false, false))
     callback_filters = ODE.DiscreteCallback(condition_every_iter, affect_filter!; save_positions = (false, false))
     callback_adapt_dt = ODE.DiscreteCallback(condition_every_iter, adaptive_dt!; save_positions = (false, false))
     callback_adapt_dt = sim.adapt_dt ? (callback_adapt_dt,) : ()
 
+    callbacks = ODE.CallbackSet(callback_adapt_dt..., callback_cfl..., callback_dtmax, callback_filters, callback_io)
+
```
trontrytel, line 567: I thought we agreed that the monitor_cfl callback should be done after we compute the dt?

trontrytel (changes requested):

driver/main.jl
charleskawczynski: Ah, right, we can change this

driver/main.jl
charleskawczynski: I've added a short comment, let me know if/how we can modify it

trontrytel: lgtm!

trontrytel (approved):

charleskawczynski:
Results are virtually unchanged. The DryBubble is a bit different, but we dropped tests on it, and so it may just be unstable and sensitive to small changes.

PR #622, opened 2021-12-04
Add changelog to TagBot GHA

charleskawczynski:
This PR adds a changelog to the tagbot, so that we don't need to manually go through the PRs for the release notes. I'm not sure if the strings need an exact match or not (e.g., `'bug' in in issue.labels`, where the label may be `bug :bug:`).


jakebolewski:
they need to be an exact match, I would stick to ascii labels to make things easy

otherwise LGTM but I didn't check over the bracket matching carefully

jakebolewski (approved):

PR #624, opened 2021-12-05
Use more ClimaCore operators

charleskawczynski:


charleskawczynski:

PR #625, opened 2021-12-06
Move updraft_top to update_aux!

### LINKED ISSUE #550: `up.updraft_top` should not be computed in `compute_diagnostics`, as simulations depend on it
Thanks, @yairchn, for pointing this out.

charleskawczynski:
Closes #550. This PR move `updraft_top`, which the model depends on, from `compute_diagnostics!` to `update_aux!`. This PR also moves the `GMV_third_m` computations into `compute_diagnostics!`, since they are purely diagnostic and the model does not depend on them.

cc @ilopezgp, @trontrytel


yairchn (approved):
LGTM, @trontrytel , @ilopezgp -   this bug fix will have behavioral change on the simualtions (especially TRMM) - but it is following a change to Julia.17

PR #629, opened 2021-12-07
Reorganize / split EDMF from dycore

charleskawczynski:
This PR:
 - Moves all of the callbacks from `driver/main.jl` into a separate file: `driver/callbacks.jl`
 - Moves all of the specified field variable names, specific to edmf, into `src/variables.jl`
 - Moves the `src/Variables.jl` `satadjust` method into `driver/dycore.jl`


driver/dycore.jl
```diff
@@ -0,0 +1,95 @@
...
+# Face only
+face_prognostic_vars(FT, n_up) = (; w = FT(0), TC.face_prognostic_vars_edmf(FT, n_up)...)
+# TC.face_prognostic_vars_edmf(FT, n_up) = (;) # could also use this for empty model
+
+
+#####
+##### Methods
+#####
+
+function satadjust(gm::TC.GridMeanVariables, grid, state)
```
yairchn, line 77: if this function is used only in initialization I believe we can / should just added it to"initial_conditions.jl" or "Cases.jl"

yairchn (commented):
the PR is not working yet - not sure why - but the general direction is great, left one comment

driver/dycore.jl
charleskawczynski: Ah, good point, I'm fine with that, thanks!

yairchn:
Bors r+

PR #636, opened 2021-12-09
move massflux src for cov to diagnostic

yairchn:
This PR moves massflux src for covariances as it is purely diagnostic
it also removes the extra computation of covariance detr that was in the original formulation when covariances were solved implicitly


charleskawczynski (approved):

yairchn:
Bors r+

PR #639, opened 2021-12-10
Hoist getproperty calls outside of broadcast expressions

charleskawczynski:


charleskawczynski:
This isn't actually too much of a gain, so let's leave this as is for now.

Issue #640, opened 2021-12-10
Search over sorted arrays are linear


ilopezgp:
Currently we are using linear search over sorted arrays. An example is finding the updraft top (https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/update_aux.jl#L511). This is a standard binary search problem:

```
up.updraft_top[I] = find_updraft_top(center_indices, kc_surface(grid).i, kc_top_of_atmos(grid).i)

function find_updraft_top(center_indices, bottom, top)
    mid = trunc(Int, (bottom+top)/2)
    if mid < top
        if aux_up[i].area[mid] > threshold and aux_up[i].area[mid+1] <= threshold
              return mid
        elseif aux_up[i].area[mid] > threshold
              return find_updraft_top(center_indices, mid+1, top)
        else
              return find_updraft_top(center_indices, bottom, mid)
        end
    return mid 
...
```

It would be good to be able to do this on the ClimaCore grid, passing a generic function to the search that returns either (-1, 0, 1) for go left, success, or go right.

jakebolewski:
how many items are you searching over?  in general, it would be good to rationalize that this would even make sense to implement with some micro-benchmarks

ilopezgp:
This is done once per time step and updraft, since it is needed to advance the prognostic state.

ilopezgp:
Another option is to start the linear search at the previous updraft top, since the evolution is smooth. In this case, we know whether to traverse down or up by evaluating the current updraft fraction. This heuristic would be even faster, but specific to this particular problem.

charleskawczynski:
This might improve things, but I think we should make profile-guided decisions so that we get the best bang for our buck. We can leave the issue open, but I don't think we should bother to take action until we profile to see how this measures up to other performance bottlenecks.

PR #641, opened 2021-12-10
End updraft top search when no updraft is found.

ilopezgp:
Since we might not want to put resources into changing grid traversal or search over the ClimaCore grid (#640), this should be an easy way to reduce unnecessary checks. We can end the search as soon as we hit a height with no updraft because updrafts are always defined from z_surface to some z, without reappearance above. This definition of updraft is consistent with our use of updraft height in the sub grid pressure closures.

### DIFF
--- a/src/update_aux.jl
+++ b/src/update_aux.jl
@@ -511,6 +511,8 @@ function update_aux!(edmf::EDMF_PrognosticTKE{N_up}, gm, grid, state, Case, para
         @inbounds for k in real_center_indices(grid)
             if aux_up[i].area[k] > 1e-3
                 up.updraft_top[i] = max(up.updraft_top[i], grid.zc[k])
+            else
+                break
             end
         end
     end


yairchn:
this changes lifecycle and bubble - as in this case you might have an updraft above non updraft

ilopezgp:
I see. In this case, do we need to compute the updraft base to get the updraft height?

ilopezgp:
What enters the pressure closure should be the height of the updraft, not the updraft top I guess.

ilopezgp:
@jiahe23

ilopezgp:
I am not sure how to handle this if we can have area fraction arbitrarily vanishing and reappearing at any height. Shouldn't in that case the height entering the closure be different for updrafts below and above?

ilopezgp:
Closing this PR and opening an issue discussing the problems with the current implementation of the pressure closure.

PR #649, opened 2021-12-14
Try different dispatch for generic functions

charleskawczynski:


charleskawczynski:
We've basically made things pretty static in favor of improving inference (which resulted in fewer inference triggers)

PR #650, opened 2021-12-14
Fix CI, update compat and CLIMAParameters call func call

trontrytel:
@charleskawczynski this should be the fix


trontrytel:
bors try

charleskawczynski (approved):

PR #658, opened 2021-12-16
use SurfaceFluxes.jl rm surface_functions.jl

yairchn:
This PR removes surface_functions.jl  and uses SurfaceFluxes.jl v0.2.0 to compute the Monin Obukhov surface layer


charleskawczynski (approved):
Code changes look fine to me. It looks like GABLS is a bit different now. I'm assuming this is okay?

yairchn:
> Code changes look fine to me. It looks like GABLS is a bit different now. I'm assuming this is okay?

yes these are acceptable changes. thanks

src/Surface.jl
```diff
@@ -1,42 +1,25 @@
...
 
 #####
 ##### SurfaceNone
 #####
 
 function update(surf::SurfaceBase{SurfaceNone}, grid, state, gm::GridMeanVariables, param_set)
     # JH: assigning small fluxed so that simulation won"t crash when computing mixing length
+    # UPDATE  - YC  - the windspeed, zrough, bflux values that were here amounted to a
+    # prescribed ustar 0.000540989171658113.
+    # ustar plays a role in the surface covarianve and TKE and expalin the changes seen when
```
ilopezgp, line 10:
```suggestion
    # ustar plays a role in the surface covariance and TKE and explains the changes seen when
```

src/Surface.jl
```diff
@@ -1,42 +1,25 @@
...
+    # switching from SCAMPy to TC.jl in the second moment profile in the buble case.
```
ilopezgp, line 11:
```suggestion
    # switching from SCAMPy to TC.jl in the second moment profile in the bubble case.
```

src/Surface.jl
```diff
@@ -1,42 +1,25 @@
...
-    free_convection_windspeed(surf, grid, state, gm, param_set, BaseCase())
 
 #####
 ##### SurfaceNone
 #####
 
 function update(surf::SurfaceBase{SurfaceNone}, grid, state, gm::GridMeanVariables, param_set)
     # JH: assigning small fluxed so that simulation won"t crash when computing mixing length
+    # UPDATE  - YC  - the windspeed, zrough, bflux values that were here amounted to a
+    # prescribed ustar 0.000540989171658113.
```
ilopezgp, line 9: Doesn't the value of ustar also depend on `grid.zc[kc_surf]`? Can you still leave this dependence?

src/Surface.jl
```diff
@@ -46,49 +29,56 @@ function update(surf::SurfaceBase{SurfaceFixedFlux}, grid, state, gm::GridMeanVa
     v_gm_surf = prog_gm.v[kc_surf]
     q_tot_gm_surf = prog_gm.q_tot[kc_surf]
     θ_liq_ice_gm_surf = prog_gm.θ_liq_ice[kc_surf]
-    T_gm_surf = aux_gm.T[kc_surf]
 
-    ts = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
-    rho_tflux = surf.shf / TD.cp_m(ts)
+    ts_sfc = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
+    ts_in = thermo_state_pθq(param_set, p0_f_surf, θ_liq_ice_gm_surf, q_tot_gm_surf)
```
ilopezgp, line 34: I think `p0_f_surf` is staggered wrt `θ_liq_ice_gm_surf, q_tot_gm_surf`.

src/Surface.jl
```diff
@@ -46,49 +29,56 @@ function update(surf::SurfaceBase{SurfaceFixedFlux}, grid, state, gm::GridMeanVa
...
-                print("α0_f_surf ==>", α0_f_surf)
-                error("Terminating execution.")
-            end
-        end
-
-        surf.ustar = compute_friction_velocity(param_set, surf.windspeed, surf.bflux, surf.zrough, zb)
+    scheme = SF.FVScheme()
+    surf.bflux = SF.compute_buoyancy_flux(param_set, surf.shf, surf.lhf, ts_in, ts_sfc, scheme)
+    zi = get_inversion(grid, state, param_set, surf.Ri_bulk_crit)
+    gustiness = get_wstar(surf.bflux, zi) # yair here zi in TRMM should be adjusted
```
ilopezgp, line 39: This is the Deardorff velocity or convective velocity scale, not the gustiness. We had parameterized the gustiness as proportional to this velocity. I am fine with setting the proportionality coefficient to 1, but I would like this to be stated explicitly.

ilopezgp: Finally, `get_wstar` is only meant to be applied in convective conditions, and the way it is coded it will return a negative value in stable conditions. Please, introduce a condition based on the sign of the stability (MO-based or bflux-based) either here or within the function itself. This should return 0 for stable conditions.

src/Surface.jl
```diff
@@ -46,49 +29,56 @@ function update(surf::SurfaceBase{SurfaceFixedFlux}, grid, state, gm::GridMeanVa
...
+        sc = SF.Fluxes{FT}(
+            state_in = vals_int,
+            state_sfc = vals_sfc,
+            shf = surf.shf,
+            lhf = surf.lhf,
+            z0m = surf.zrough,
+            z0b = surf.zrough,
+            gustiness = gustiness,
+        )
+        result = SF.surface_conditions(param_set, sc, UF.Businger(), scheme)
```
ilopezgp, line 67: Can you expose the `UF` choice as a kwarg?

src/Surface.jl
```diff
@@ -102,85 +92,87 @@ function update(surf::SurfaceBase{SurfaceFixedCoeffs}, grid, state, gm::GridMean
...
-
-    surf.ustar = friction_velocity_given_windspeed(surf.cm, surf.windspeed)
-    surf.obukhov_length = obukhov_length(param_set, surf.ustar, surf.bflux)
-
-    surf.rho_uflux = -ρ0_f_surf * surf.ustar^2 / surf.windspeed * u_gm_surf
-    surf.rho_vflux = -ρ0_f_surf * surf.ustar^2 / surf.windspeed * v_gm_surf
+    z_sfc = FT(0)
+    z_in = grid.zc[kc_surf].z
+    ts_sfc = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
+    ts_in = thermo_state_pθq(param_set, p0_f_surf, θ_liq_ice_gm_surf, q_tot_gm_surf)
```
ilopezgp, line 98: Here again, maybe p is staggered with θ and q.

src/Surface.jl
```diff
@@ -102,85 +92,87 @@ function update(surf::SurfaceBase{SurfaceFixedCoeffs}, grid, state, gm::GridMean
...
+    u_sfc = SA.SVector{2, FT}(0, 0)
+    u_in = SA.SVector{2, FT}(u_gm_surf, v_gm_surf)
+    vals_sfc = SF.SurfaceValues(z_sfc, u_sfc, ts_sfc)
+    vals_int = SF.InteriorValues(z_in, u_in, ts_in)
+    sc = SF.ValuesOnly{FT}(state_in = vals_int, state_sfc = vals_sfc, z0m = surf.zrough, z0b = surf.zrough)
+    result = SF.surface_conditions(param_set, sc, UF.Businger(), SF.FVScheme())
+    surf.cm = result.Cd
+    surf.ch = result.Ch
+    surf.obukhov_length = result.L_MO
+    surf.lhf = result.lhf * 0.0
```
ilopezgp, line 158: This should only be the case in the dry implementation.

src/Surface.jl
```diff
@@ -194,91 +186,74 @@ function update(surf::SurfaceBase{SurfaceMoninObukhovDry}, grid, state, gm::Grid
     T_gm_surf = aux_gm.T[kc_surf]
     Pg = surf.ref_params.Pg
 
-    pvg = TD.saturation_vapor_pressure(param_set, TD.PhaseEquil, surf.Tsurface)
-    surf.qsurface = TD.q_vap_saturation_from_density(param_set, surf.Tsurface, ρ0_f_surf, pvg)
-    lv = TD.latent_heat_vapor(param_set, T_gm_surf)
-
     phase_part = TD.PhasePartition(surf.qsurface, 0.0, 0.0)
     h_star = TD.liquid_ice_pottemp_given_pressure(param_set, surf.Tsurface, Pg, phase_part)
```
ilopezgp, line 190: Something is not consistent here. You are passing `h_star` to the `ts_sfc` constructor, but on the base constructor above you are passing `T_surface` instead. I think it might be more accurate to pass `h_star` everywhere, what do you think? In addition, it might be better to change `h` for `θ_liq_ice`.

yairchn: well -  I am just following what is already in main coming from SCAMPy. git blame in SCAMPy points to @ilopezgp ;-)
What I think is happening here is that we are trying to construct a thermodynamic state that has no condensate in a sub freezing temperature. So the Hola-hoop is to get the zero condensate 'liquid_ice_pottemp', and to setup a thermodynamic state from it. If we change to
'ts_sfc = thermo_state_pTq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)'
we would probably get a thermodynamic state  with ice.

yairchn: not sure what you mean by 
'I think it might be more accurate to pass h_star everywhere, what do you think'

yairchn: yes to 'change h for θ_liq_ice'

yairchn (commented):
@ilopezgp thanks for the review, see my replies below + pdf DM on slack

akshaysridhar:
@yairchn Can the `update` functions be collapsed into one instance with different `sc:AbstractSurfaceConditions` (from SurfaceFluxes.jl) types? It seems they all populate the same 11 variables in `surf`.

yairchn:
> @yairchn Can the `update` functions be collapsed into one instance with different `sc:AbstractSurfaceConditions` (from SurfaceFluxes.jl) types? It seems they all populate the same 11 variables in `surf`.

@charleskawczynski -- this is what I was suggesting a while back, what do you think? there is a lot of repetition in the code, maybe in a future PR we can make these changes?

driver/Cases.jl
```diff
@@ -120,7 +120,7 @@ get_surface_type(::Rico) = TC.SurfaceFixedCoeffs
 get_surface_type(::GATE_III) = TC.SurfaceFixedCoeffs
 get_surface_type(::GABLS) = TC.SurfaceMoninObukhovDry
 get_surface_type(::SP) = TC.SurfaceSullivanPatton
-get_surface_type(::DryBubble) = TC.SurfaceNone
+get_surface_type(::DryBubble) = TC.SurfaceFixedFlux
```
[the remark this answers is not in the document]
charleskawczynski: If we're going to make this change, then I'd prefer just deleting the line so that it falls back on the `::AbstractCaseType`

charleskawczynski:
Let's move forward with this and we can refactor `update` in a follow-up PR. We need to address the mutability issue, too. So there's several more iterations on this file remaining.

yairchn:
> I'm fine with the changes as they are. Let's move forward with this and we can refactor `update` in a follow-up PR. We need to address the mutability issue, too. So there's several more iterations on this file remaining.

make sure to wait for @ilopezgp's approval - he had several comments and I wanna make sure they were fully addressed

src/Surface.jl
ilopezgp: I am saying passing `h_star` is correct, this is correct. But in other parts of the code, you are passing `T_surface` instead to the constructor, and the constructor should read a liquid ice potential temperature. I will point to those sections of the code.

src/Surface.jl
```diff
@@ -1,284 +1,235 @@
...
-                print("surf.shf ==>", surf.shf)
-                print("surf.lhf ==>", surf.lhf)
-                print("u_gm_surf ==>", u_gm_surf)
-                print("v_gm_surf ==>", v_gm_surf)
-                print("q_tot_gm_surf ==>", q_tot_gm_surf)
-                print("α0_f_surf ==>", α0_f_surf)
-                error("Terminating execution.")
-            end
-        end
+    ts_sfc = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
```
ilopezgp, line 17: this is what I was referring to earlier, with the `h_star` comment. Here, we are passing `surf.Tsurface` as the potential temperature, but there are virtual loading effects that we are not taking into account. I think you should pass a potential temperature here as well.

src/Surface.jl
```diff
@@ -1,284 +1,235 @@
...
-    surf.ustar = friction_velocity_given_windspeed(surf.cm, surf.windspeed)
-    surf.obukhov_length = obukhov_length(param_set, surf.ustar, surf.bflux)
-
-    surf.rho_uflux = -ρ0_f_surf * surf.ustar^2 / surf.windspeed * u_gm_surf
-    surf.rho_vflux = -ρ0_f_surf * surf.ustar^2 / surf.windspeed * v_gm_surf
+    universal_func = UF.Businger()
+    scheme = SF.FVScheme()
+    z_sfc = FT(0)
+    z_in = grid.zc[kc_surf].z
+    ts_sfc = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
```
ilopezgp, line 82: Here again, we are passing `surf.Tsurface`, it should be the corresponding θ_liq_ice.

src/Surface.jl
```diff
@@ -1,284 +1,235 @@
...
-
-    ts = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
-    surf.shf = TD.cp_m(ts) * surf.rho_hflux
-
-    surf.bflux = buoyancy_flux(param_set, surf.shf, surf.lhf, surf.Tsurface, surf.qsurface, α0_f_surf, ts)
-    surf.ustar = friction_velocity_given_windspeed(surf.cm, surf.windspeed)
 
+    universal_func = UF.Businger()
+    scheme = SF.FVScheme()
+    ts_sfc = thermo_state_pTq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
```
ilopezgp, line 128: Where is the `thermo_state_pTq` method defined? Should this be `thermo_state_pθq`? If so, please update the `surf.Tsurface` to be the appropriate liq_ice_θ.

src/Surface.jl
```diff
@@ -1,284 +1,235 @@
...
-    surf.rho_qtflux = -surf.ch * surf.windspeed * (q_tot_gm_surf - surf.qsurface) * ρ0_f_surf
-    surf.lhf = lv * 0.0
-
-    ts = thermo_state_pθq(param_set, p0_f_surf, surf.Tsurface, surf.qsurface)
-    surf.shf = TD.cp_m(ts) * surf.rho_hflux
-
-    surf.bflux = buoyancy_flux(param_set, surf.shf, surf.lhf, surf.Tsurface, surf.qsurface, α0_f_surf, ts)
-    surf.ustar = friction_velocity_given_windspeed(surf.cm, surf.windspeed)
-
+    θ_star = TD.liquid_ice_pottemp_given_pressure(param_set, surf.Tsurface, p0_f_surf, phase_part)
```
ilopezgp, line 167: θ_star is typically interpreted as the surface potential temperature flux divided by the friction velocity u_star, see https://github.com/CliMA/SurfaceFluxes.jl/blob/main/src/SurfaceFluxes.jl#L559 for instance for the Monin-Obukhov relation to an actual temperature within the surface layer. Can you change the name of this variable?

src/Surface.jl
yairchn: sorry I meant 'TD.PhaseEquil_pTq'

ilopezgp (approved):
After the rename of theta_star and the bug in the pTq construct I am good with this. But let's not forget about the issue with surface constructors!

yairchn:
Bors r+

PR #662, opened 2021-12-18
Remove moisture comparison in GABLS to fix CI

charleskawczynski:
This PR removes `qt_mean` and `QTvar_mean` comparisons for GABLS because `qt` is zero in the GABLS paper.


ilopezgp (approved):
LGTM

PR #672, opened 2021-12-22
Try stretched grids

charleskawczynski:
In this PR we added an option for `GeneralizedExponentialStretching` with a grid realistic for a GCM. To make sure that the stretch grid in all cases uses similar points (to each case's vertical domain) the stretch grid is generated first for the GCM domain (45km) and we added a function `TCMeshFromGCMMesh` that takes a subset of that domain relevant for each case (i.e. 400m for GABLS and 17km for TRMM). This stretching effects the results of several cases and for now it is not the default, but appears in an addition Bomex base.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -95,6 +95,10 @@ steps:
         command: "julia --color=yes --project=integration_tests integration_tests/driver.jl --case Bomex --micro gaussian"
         artifact_paths: "Output.Bomex.01_gaussian/stats/comparison/*"
 
+      - label: ":partly_sunny: Bomex with stretched grid"
+        command: "julia --color=yes --project=integration_tests integration_tests/driver.jl --case Bomex --stretch_grid true"
+        artifact_paths: "Output.Bomex.01_stretch_grid_true/stats/comparison/*"
+
       - label: ":partly_sunny: Bomex with NN_nonlocal"
         command: "julia --color=yes --project=integration_tests integration_tests/driver.jl --case Bomex --entr NN_nonlocal"
         artifact_paths: "Output.Bomex.01_NN_nonlocal/stats/comparison/*"
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -142,6 +142,12 @@ function default_namelist(
     # From namelist
     namelist_defaults["grid"] = Dict()
     namelist_defaults["grid"]["dims"] = 1
+    namelist_defaults["grid"]["stretch"] = Dict()
+    namelist_defaults["grid"]["stretch"]["flag"] = false
+    namelist_defaults["grid"]["stretch"]["nz"] = 55
+    namelist_defaults["grid"]["stretch"]["dz_surf"] = 30.0
+    namelist_defaults["grid"]["stretch"]["dz_toa"] = 8000.0
+    namelist_defaults["grid"]["stretch"]["z_toa"] = 45000.0
 
     namelist_defaults["thermodynamics"] = Dict()
     namelist_defaults["thermodynamics"]["thermal_variable"] = "thetal"
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -72,18 +72,37 @@ function Simulation1d(namelist)
     cfl_limit = namelist["time_stepping"]["cfl_limit"]
     dt_min = namelist["time_stepping"]["dt_min"]
 
+    truncated_gcm_mesh = TC.parse_namelist(namelist, "grid", "stretch", "flag"; default = false)
+
     Δz = FT(namelist["grid"]["dz"])
     nz = namelist["grid"]["nz"]
     z₀, z₁ = FT(0), FT(nz * Δz)
-    domain = CC.Domains.IntervalDomain(
-        CC.Geometry.ZPoint{FT}(z₀),
-        CC.Geometry.ZPoint{FT}(z₁),
-        boundary_tags = (:bottom, :top),
-    )
-    mesh = CC.Meshes.IntervalMesh(domain, nelems = nz)
+    if truncated_gcm_mesh
+        nzₛ = namelist["grid"]["stretch"]["nz"]
+        Δzₛ_surf = FT(namelist["grid"]["stretch"]["dz_surf"])
+        Δzₛ_top = FT(namelist["grid"]["stretch"]["dz_toa"])
+        zₛ_toa = FT(namelist["grid"]["stretch"]["z_toa"])
+        stretch = CC.Meshes.GeneralizedExponentialStretching(Δzₛ_surf, Δzₛ_top)
+        domain = CC.Domains.IntervalDomain(
+            CC.Geometry.ZPoint{FT}(z₀),
+            CC.Geometry.ZPoint{FT}(zₛ_toa),
+            boundary_tags = (:bottom, :top),
+        )
+        gcm_mesh = CC.Meshes.IntervalMesh(domain, stretch; nelems = nzₛ)
+        mesh = TC.TCMeshFromGCMMesh(gcm_mesh; z_max = z₁)
+    else
+        CC.Meshes.Uniform()
+        domain = CC.Domains.IntervalDomain(
+            CC.Geometry.ZPoint{FT}(z₀),
+            CC.Geometry.ZPoint{FT}(z₁),
+            boundary_tags = (:bottom, :top),
+        )
+        mesh = CC.Meshes.IntervalMesh(domain, nelems = nz)
+    end
     grid = TC.Grid(mesh)
 
     Stats = skip_io ? nothing : NetCDFIO_Stats(namelist, grid)
+
     case_type = Cases.get_case(namelist)
     surf_ref_state = Cases.surface_ref_state(case_type, param_set, namelist)
 
--- a/integration_tests/driver.jl
+++ b/integration_tests/driver.jl
@@ -12,6 +12,7 @@ if run_from_command_line
         "--entr"           # Try other entr-detr models
         "--stoch_entr"     # Choose type of stochastic entr-detr model
         "--calibrate_io"   # Test that calibration IO passes regression tests
+        "--stretch_grid"   # Test stretched grid option
         "--skip_io"        # Test that skipping IO passes regression tests
     end
     parsed_args = ArgParse.parse_args(ARGS, s)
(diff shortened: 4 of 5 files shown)


szy21:
Is the generalized stretching function you need this one: https://github.com/CliMA/ClimaCore.jl/blob/8725010c313222d512d06a576cca8f7fa5e27f01/src/Meshes/interval.jl#L100? If so we can just change this line after the PR is merged

trontrytel:
Not sure why buildkite CI is marked all green. The results for TRMM look different. For example meant temperature and horizontal velocities are off near the top boundary of the domain

yairchn:
> Not sure why buildkite CI is marked all green. The results for TRMM look different. For example meant temperature and horizontal velocities are off near the top boundary of the domain

I extended the domain above what it was and so you see theta profile that more in line with PYCLES. I added a slight increase in temperature at the top to have a strong tropopause in case updrafts develop too high, so they don't hit the top boundary

yairchn:
The tables have been updated, so they do not reflect the changes in the results. Are the convection changes substantial to you? or only the mean profiles?

trontrytel:
> The tables have been updated, so they do not reflect the changes in the results. Are the convection changes substantial to you? or only the mean profiles?

Hard to compare the contours if the domain is changed. Maybe we could extend the domain in one PR and then add the streching in another? This way we will be more certain that it does not affect the simulation negatively?

yairchn:
> the contours if the domain is changed. Maybe we could extend the domain in one PR and then add the streching in another? This way we will be more certain that it does not affect the simulation negatively?

but the profiles are aligned so you can see there the perspective of how high convection developed and the values associated with it

trontrytel:
> > the contours if the domain is changed. Maybe we could extend the domain in one PR and then add the streching in another? This way we will be more certain that it does not affect the simulation negatively?
> 
> but the profiles are aligned so you can see there the perspective of how high convection developed and the values associated with it

Looking just at the profiles then: I can spot that the variance is increasing in some cases, for example ARM, TRMM and Rico.

yairchn:
> > > the contours if the domain is changed. Maybe we could extend the domain in one PR and then add the streching in another? This way we will be more certain that it does not affect the simulation negatively?
> > 
> > 
> > but the profiles are aligned so you can see there the perspective of how high convection developed and the values associated with it
> 
> Looking just at the profiles then: I can spot that the variance is increasing in some cases, for example ARM, TRMM and Rico.

True, but are these increases a problem? do you think this change should not be merged? Variances are noisy anyways.

yairchn (approved):

trontrytel:
> > Looking just at the profiles then: I can spot that the variance is increasing in some cases, for example ARM, TRMM and Rico.
> 
> True, but are these increases a problem? do you think this change should not be merged? Variances are noisy anyways.

For Rico the difference is small and looks like noise. For TRMM I'm not sure if the change comes from stretching or the change of profiles you made. For ARM I'm just confused because I would think that the simulation should not have been changed at all? Looks good otherwise.

Did you guys test what is the effect of using the stretched grid on the simulation quality, for example by making the stretching more and more coarse? Or in other words, how did you chose the default value that is there right now?

Also, would it be possible to have it switch on/off as a namelist argument? In case we need it for debugging later?

yairchn:
I think this PR is ready now @trontrytel

driver/generate_namelist.jl
```diff
@@ -298,7 +298,7 @@ function TRMM_LBA(namelist_defaults)
     namelist["meta"]["casename"] = "TRMM_LBA"
 
     namelist["grid"]["nz"] = 80
-    namelist["grid"]["dz"] = 200
+    namelist["grid"]["dz"] = 210
```
ilopezgp, line 301: Why are the `dz` values changed?

ilopezgp (changes requested):
I see that the exponential stretching would be hardcoded in this PR. I strongly prefer this to be optional, as @trontrytel suggested. Hardcoding the value of stretching is also not great.

driver/generate_namelist.jl
yairchn: Sorry I meant to increase the domain height using 'nz' not 'dz'. I fixed the mistake now. Basically since the addition of snow the domain height was "tight" for the convection and the slight changes in grid stretching caused convection to hit the top and the model breaks (instability is detected).

src/Grid.jl
```diff
@@ -120,41 +120,5 @@ Base.iterate(fi::Base.Iterators.Reverse{T}, state = Nstop) where {Nstart, Nstop,
 face_space(grid::Grid) = grid.fs
 center_space(grid::Grid) = grid.cs
 
-#=
-    findfirst_center
```
charleskawczynski, line 124: This should not be commented out... @yairchn

driver/generate_namelist.jl
```diff
@@ -213,8 +215,8 @@ function Soares(namelist_defaults)
     namelist = deepcopy(namelist_defaults)
     namelist["meta"]["casename"] = "Soares"
 
-    namelist["grid"]["nz"] = 75
-    namelist["grid"]["dz"] = 50.0
+    namelist["grid"]["nz"] = 60
+    namelist["grid"]["dz"] = 60.0
```
ilopezgp, line 219: How should I think about `dz` when the grid is stretched? Is it the minimum dz (first level)?

If so, why are we increasing the dz here? In the end, we probably want 50m near the surface, and coarsening above. I don't see why we have to coarsen the grid.

driver/generate_namelist.jl
```diff
@@ -229,11 +231,11 @@ function Nieuwstadt(namelist_defaults)
     namelist = deepcopy(namelist_defaults)
     namelist["meta"]["casename"] = "Nieuwstadt"
 
-    namelist["grid"]["nz"] = 75
-    namelist["grid"]["dz"] = 50.0
+    namelist["grid"]["nz"] = 60
+    namelist["grid"]["dz"] = 60.0
```
ilopezgp, line 235: Same here and in others.

driver/generate_namelist.jl
```diff
@@ -136,6 +136,8 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
     # From namelist
     namelist_defaults["grid"] = Dict()
     namelist_defaults["grid"]["dims"] = 1
+    namelist_defaults["grid"]["stretch"] = true
```
ilopezgp, line 139: I assume this is for testing this PR, but we will change it to False  before merging?

ilopezgp:
In general, I would prefer the stretching to be set to False by default for now, it makes calibration simpler. I mention this because I see that the mse tables have been updated to the stretched formulation.

yairchn:
> In general, I would prefer the stretching to be set to False by default for now, it makes calibration simpler. I mention this because I see that the mse tables have been updated to the stretched formulation.

First, Tapio prefers to have stretching as default - but that would be with the new stretching that is still in PR355 in CC. I could set it as false now but just heads up it will not stay this way.

The other issue is that if I set it to default now we will not be testing it ...

driver/callbacks.jl
```diff
@@ -127,9 +127,9 @@ function monitor_cfl!(integrator)
             CFL_in_rain = Δt / Δz[k] * term_vel_rain[k + 1]
             CFL_in_snow = Δt / Δz[k] * term_vel_snow[k + 1]
         end
-        if max(CFL_in_rain, CFL_in_snow, CFL_out_rain, CFL_out_snow) > CFL_limit
-            error("Time step is too large for rain fall velocity!")
-        end
+        # if max(CFL_in_rain, CFL_in_snow, CFL_out_rain, CFL_out_snow) > CFL_limit
+        #     error("Time step is too large for rain fall velocity!")
+        # end
```
charleskawczynski, line 132: We're going to uncomment these, correct?

driver/main.jl
```diff
@@ -61,18 +61,39 @@ function Simulation1d(namelist)
...
-    z₀, z₁ = FT(0), FT(nz * Δz)
-    domain = CC.Domains.IntervalDomain(
-        CC.Geometry.ZPoint{FT}(z₀),
-        CC.Geometry.ZPoint{FT}(z₁),
-        boundary_tags = (:bottom, :top),
-    )
-    mesh = CC.Meshes.IntervalMesh(domain, nelems = nz)
+    stretched_grid = TC.parse_namelist(namelist, "grid", "stretch"; default = false)
+
+    if stretched_grid
```
charleskawczynski, line 66: Should we maybe call this `truncated_gcm_mesh` or something? This is different from, and maybe somewhat unintuitive compared to, using a simple exponential stretching.

trontrytel (approved):
LGTM

Is there any link to ClimaCore docs you could add somewhere to explain how the mesh is constructed? I'm guessing that not all combinations of `nz`, `dz_surf`, `dz_toa` and `z_toa` 
are valid?

```
    namelist_defaults["grid"]["stretch"]["nz"] = 55
    namelist_defaults["grid"]["stretch"]["dz_surf"] = 30.0
    namelist_defaults["grid"]["stretch"]["dz_toa"] = 8000.0
    namelist_defaults["grid"]["stretch"]["z_toa"] = 45000.0
```

.buildkite/pipeline.yml
```diff
@@ -95,6 +95,10 @@ steps:
         command: "julia --color=yes --project=integration_tests integration_tests/driver.jl --case Bomex --micro gaussian"
         artifact_paths: "Output.Bomex.01_gaussian/stats/comparison/*"
 
+      - label: ":partly_sunny: Bomex with stretched grid"
+        command: "julia --color=yes --project=integration_tests integration_tests/driver.jl --case Bomex --stretch_grid true"
+        artifact_paths: "Output.Bomex.01_stretch_grid/stats/comparison/*"
```
charleskawczynski, line 100:
```suggestion
        artifact_paths: "Output.Bomex.01_stretch_grid_true/stats/comparison/*"
```

charleskawczynski: I think this should make the plots appear

yairchn:
Bors r+

ilopezgp:
> LGTM
> 
> Is there any link to ClimaCore docs you could add somewhere to explain how the mesh is constructed? I'm guessing that not all combinations of `nz`, `dz_surf`, `dz_toa` and `z_toa` are valid?
> 
> ```
>     namelist_defaults["grid"]["stretch"]["nz"] = 55
>     namelist_defaults["grid"]["stretch"]["dz_surf"] = 30.0
>     namelist_defaults["grid"]["stretch"]["dz_toa"] = 8000.0
>     namelist_defaults["grid"]["stretch"]["z_toa"] = 45000.0
> ```

+1 to this. Is this the resolution we aim for in the GCM?

Can we include as a comment above other combinations that could work? (e.g., coarser resolution for ensemble simulations, operational resolution for single simulations, high resolution for subseasonal NWP). These things WILL come up in the papers we want to write.

integration_tests/driver.jl
```diff
@@ -12,6 +12,7 @@ if run_from_command_line
         "--entr"           # Try other entr-detr models
         "--stoch_entr"     # Choose type of stochastic entr-detr model
         "--calibrate_io"   # Test that calibration IO passes regression tests
+        "--stretch_grid"   # Test stratched grid option
```
ilopezgp, line 15:
```suggestion
        "--stretch_grid"   # Test stretched grid option
```

ilopezgp (approved):

yairchn:
Bors r+

costachris (approved):

yairchn:
Bors r+

PR #673, opened 2021-12-22
Add invalidations script

charleskawczynski:
For now, this script grabs invalidations and plots the number (and percentage) of children per method invalidation.

It seems that [DiffEq.jl's `Base.mapreduce_empty`](https://github.com/SciML/DiffEqBase.jl/blob/master/src/common_defaults.jl#L4) is triggering ~1k invalidations (almost 50% of all invalidations). It is [recommended](https://timholy.github.io/SnoopCompile.jl/stable/#workflow) to tackle expensive invalidations before moving on to improving inferrability / profile-guided despecialization to improve latency.

Screen Shot 2021-12-22 at 11 15 54 AM


charleskawczynski:
cc @jakebolewski / @simonbyrne, this could be affecting ClimaCore, too, since we have many overlapping deps.

jakebolewski:
I'm not sure what we can do if it is caused by a dependency?

jakebolewski:
There are some patched diffeq support functions that are no longer necessary now that we support `FieldVector`'s, if we get rid of some of those and only support Diffeq operations on FieldVectors and not fields that might help.

charleskawczynski:
This was fixed by [DiiffEqBase.jl's 723](https://github.com/SciML/DiffEqBase.jl/pull/723) 🎉

ChrisRackauckas:
How did you get the number of invalidations to see the priority? I want to run that on all of DiffEq and just do a bunch of quick PRs like this.

charleskawczynski:
> How did you get the number of invalidations to see the priority? I want to run that on all of DiffEq and just do a bunch of quick PRs like this.

@ChrisRackauckas, Everything is actually in this PR, here's a short pseudo-code summary:

```julia
# From: https://timholy.github.io/SnoopCompile.jl/stable/snoopr/
using SnoopCompileCore
invalidations = @snoopr begin
    # do representative work
end;

import SnoopCompile
trees = SnoopCompile.invalidation_trees(invalidations);

@show length(SnoopCompile.uinvalidated(invalidations)) # show total invalidations

show(trees[end]) # show the most invalidated method

# Count number of children (number of invalidations per invalidated method)
n_invalidations = map(trees) do methinvs
    SnoopCompile.countchildren(methinvs)
end

import Plots
Plots.plot(
    1:length(trees),
    n_invalidations;
    markershape = :circle,
    xlabel = "i-th method invalidation",
    label = "Number of children per method invalidations",
)
```

PR #682, opened 2021-12-24
Make sde closure immutable

charleskawczynski:
Mutable structs are not allowed on the GPU, and this one doesn't seem to be used AFAICT. Is this okay, @haakon-e?

### DIFF
--- a/src/stochastic_closures.jl
+++ b/src/stochastic_closures.jl
@@ -52,7 +52,6 @@ function sde_closure(param_set::APS, sde_model::sde_struct{SDEClosureType}, term
     u0 = sde_model.u0
     dt = sde_model.dt
     u = sde(θ, σ, u0, dt)
-    sde_model.u0 = u
     return u
 end
 
--- a/src/types.jl
+++ b/src/types.jl
@@ -239,7 +239,7 @@ struct LogNormalClosureType end
 struct SDEClosureType end
 
 # Stochastic differential equation memory
-Base.@kwdef mutable struct sde_struct{T}
+Base.@kwdef struct sde_struct{T}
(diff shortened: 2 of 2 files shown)


haakon-e:
It is used on L52-55 to update the SDE state.

charleskawczynski:
> It is used on L52-55 to update the SDE state. But instead of L55, I could try to reconstruct the struct entirely.

Sorry, I think I wasn't specific enough in my initial comment. It looks like the updated value `sde_model.u0` is not used after it's mutated.

If the mutated `sde_model.u0` value may be used in general, then I would agree that it'd be best if we just construct a new `sde_model` and return it from this function and pass it back to wherever it's needed.

haakon-e:
I’ll have a closer look in the new year. I think I may be moving to a different struct in which the state is stored.
________________________________
Fra: Charles Kawczynski ***@***.***>
Sendt: søndag, desember 26, 2021 7:24 pm
Til: CliMA/TurbulenceConvection.jl
Kopi: Haakon Ludvig Langeland Ervik; Mention
Emne: Re: [CliMA/TurbulenceConvection.jl] Make sde closure immutable (PR #682)


It is used on L52-55 to update the SDE state. But instead of L55, I could try to reconstruct the struct entirely.

Sorry, I think I wasn't specific enough in my initial comment. It looks like the updated value sde_model.u0 is not used after it's mutated.

If the mutated sde_model.u0 value may be used in general, then I would agree that it'd be best if we just construct a new sde_model and return it from this function and pass it back to wherever it's needed.

charleskawczynski:
Going to merge this and we can adjust it later if needed.

PR #693, opened 2021-12-28
Remove several unused edmf variables

charleskawczynski:
This PR removes several unused EDMF variables-- some are
 - simply never assigned / used
 - only assigned then never used
 - never used, but also appear as InternalClimaParameters (search `ICP.`)

This PR gets us a reasonable step towards immutable types. We only have
 - `SurfaceBase` (may be a bit tough)
 - `CasesBase` (should be relatively easy once `SurfaceBase` is immutable)
 - `sde_struct` (being fixed in #682)
 - `EDMF_PrognosticTKE` (for which `dt_max` is the only remaining mutating field)

I imagine `edmf.zi = get_inversion(grid, state, param_set, surface.Ri_bulk_crit)` is the most controversial removal, but it's not used! I think `zi` is computed locally and used in other places on the fly, which is perfectly fine.

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -153,7 +153,6 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
 
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["updraft_number"] = 1
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"] = "moisture_deficit" # "moisture_deficit" or "NN"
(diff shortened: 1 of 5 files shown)


driver/generate_namelist.jl
```diff
@@ -83,6 +83,7 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["surface_area"] = 0.1
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["max_area"] = 0.9
     # mixing_length
+    # mixing_length
```
ilopezgp, line 86: Delete?

ilopezgp (approved):

PR #703, opened 2022-01-05
Make `SurfaceBase` immutable

### LINKED ISSUE #702: Remove mutable structs for all model components
Using immutable structs, where possible, is a much simpler design pattern to follow since it greatly simplifies computation and initialization. In addition, it's necessary to use immutable structs on the GPU.

charleskawczynski:
This PR makes `SurfaceBase` immutable. This is a step towards closing #702.

This PR is slightly behavior changing because we used to use `SurfaceBase` before it was computed (inside `update_aux!`, for example, LMO is used [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/update_aux.jl#L20) and computed [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/update_aux.jl#L236))


yairchn (approved):
I wonder why there are any behavioral change at all, but code changes and results look good

charleskawczynski:
It's probably due to using some surface variables at previous timesteps (before `update_surface` was called). Now, `surf = get_surface(...)` is pushed all the way to the top of `update_aux!` so that we're always using the latest data from the current timestep. That's the only thing I can think of, at least.

PR #705, opened 2022-01-06
Make CasesBase immutable

### LINKED ISSUE #702: Remove mutable structs for all model components

charleskawczynski:
This PR makes `CasesBase` immutable, and does so by replacing the hard-coded interpolated radiation profile with use of `Dierckx`'s 2D linear interpolation.

Closes #702


charleskawczynski:
Results look virtually identical

PR #706, opened 2022-01-06
remove the DryMoninObukhov surface option

yairchn:
It turns out that by setting 'qsurface=0' gabls can use MoninObukhov and obtain the same result (i.e. lhf=0) only by virtue of qsurface=0.


charleskawczynski (approved):

yairchn:
Bors r+

PR #709, opened 2022-01-06
Add precompilable scripts

charleskawczynski:
This PR adds some scripts / jobs for measuring time to first and second runs. I haven't tried doing this in a while, but I'm happy to find that our runs are significantly faster on the second run. The speedups we're seeing on the second run are:

### Including initialization and IO
``` ​
[ Info: Precompiling run: 200.599363256
[ Info: Precompiled run: 21.977667847
[ Info: precompiled/precompiling: 0.10956000801933073)
```
### Including initialization and excluding IO
``` ​
[ Info: Precompiling run: 141.141579852
[ Info: Precompiled run: 5.931822228
[ Info: precompiled/precompiling: 0.04202746089579034)
```
### Excluding initialization and including IO
``` ​
[ Info: Precompiling run: 118.238665175
[ Info: Precompiled run: 16.121032558
[ Info: precompiled/precompiling: 0.13634315419697904)
```
### Excluding initialization and excluding IO
``` ​
[ Info: Precompiling run: 92.694283328
[ Info: Precompiled run: 5.852333368
[ Info: precompiled/precompiling: 0.06313586078756807)
```

Breaking things down in this way is pretty revealing in terms of where time is spent, and where we should focus our attention.

@jakebolewski, is this an accurate measure of what is _precompilable_? What's ultimately important is whether this translates into a useful system image. I'll retry configuring a system image to double check 🤞🏻.


jakebolewski:
I don't know if it is a useful measure of what is precompilable, but it is a useful measure of once there is a compile cache the relative performance.  What subset of that compile cache makes sense to precompile is a bit different.  

What I suggest to do is generate precompile statements for a variety of useful test cases and then see where there is common intersection.  If you are compiling in values into the types (as Vals) you probably don't want to use those methods for the precompiled cache for instance.

Good to see that once there is a cache, performance is quite good?

charleskawczynski:
> I don't know if it is a useful measure of what is precompilable, but it is a useful measure of once there is a compile cache the relative performance.

Ah, ok, I think I didn't fully understand this distinction. I think this makes a bit more sense

> What I suggest to do is generate precompile statements for a variety of useful test cases and then see where there is common intersection.

Yeah, I figured this might be a good last step

> If you are compiling in values into the types (as Vals) you probably don't want to use those methods for the precompiled cache for instance unless the set of all possible values is small.

Yes, the set of all possible values is small (I believe it's ~4?)

> Good to see that once there is a cache, performance is quite good?

Agreed! Yes, performance is pretty good considering we're running fully explicitly.

PR #729, opened 2022-01-11
Delete deprecated aspect ratio, entrainment_sigma; clean up pressure and entrainment closures

ilopezgp:
Upon inspection of the code, I found two parameters that are not really independent and can be removed from the codebase. The parameters are `aspect_ratio` and `entrainment_sigma`; they both appear dividing other parameters and therefore can be absorbed into them.

Other changes include:

- Removal of the `compute_pressure_plume_spacing` deprecated function. We only really need the updraft height in the closures, this variable was kind of inherited from Zhihong's early closures.
- Fix a bug? in the computation of the diagnostic third order moment <w*w*w*>. This used a diffusivity, and we cannot really understand why it was there. We have substituted it by an isotropic turbulence formulation. This should not have any effect in the code since this was merely a diagnostic.


yairchn (approved):
LGTM

charleskawczynski (approved):

ilopezgp:
I have requested again review because I discovered that we can actually clean up the code quite a bit.

src/closures/perturbation_pressure.jl
```diff
@@ -10,59 +10,58 @@ for updraft i following [He2020](@cite), given:
  - `asp_ratio`: the specific aspect ratio of the updraft
  - `bcs`: a `NamedTuple` of BCs with updraft area fraction and buoyancy entries
 """
-function nh_pressure_buoy(param_set::APS, a_up, b_up, ρ0, asp_ratio::FT, bcs) where {FT <: Real}
+function nh_pressure_buoy(param_set::APS, a_up, b_up, ρ0, bcs)
```
charleskawczynski, line 13:
```suggestion
function nh_pressure_buoy(::Type{FT}, param_set::APS, a_up, b_up, ρ0, bcs) where {FT <: Real}
```
(and fix calls appropriately)

We kind of need to change some of these functions anyway, but let's keep this type-stable. Grabbing `FT` from `CPEDMF.α_b(param_set)` will be a run-time lookup because it's (in the long run) inherently type-unstable.

src/closures/perturbation_pressure.jl
```diff
@@ -10,59 +10,58 @@ for updraft i following [He2020](@cite), given:
...
     Ifb = CCO.InterpolateC2F(; bcs.b_up...)
     Ifa = CCO.InterpolateC2F(; bcs.a_up...)
 
-    α_b::FT = CPEDMF.α_b(param_set)
-    α₂ = FT(0)
+    α_b = CPEDMF.α_b(param_set)
+    FT = typeof(α_b)
```
charleskawczynski, line 18:
```suggestion
```

src/Turbulence_PrognosticTKE.jl
```diff
@@ -925,17 +918,8 @@ function GMV_third_m(
...
-            horiz_K_eddy = aux_up_c[i].horiz_K_eddy
-            a_up = aux_up_c[i].area
-            a_bulk = aux_bulk.area
-            @. ϕ_en_cov += -horiz_K_eddy * ∇c(wvec(w_en)) * a_up / a_bulk
-        end
-    else
-        @. ϕ_en_cov = covar_en
-    end
+    # w'w' ≈ 2/3 TKE (isotropic turbulence assumption)
+    @. ϕ_en_cov = is_tke ? 2.0 / 3.0 * covar_en : covar_en
```
charleskawczynski, line 922:
```suggestion
    @. ϕ_en_cov = is_tke ? FT(2 / 3) * covar_en : covar_en
```
and call `FT = eltype(grid)` above. I'm a bit impressed this works. Maybe take a look at the allocations analysis to make sure this doesn't pop up as a major contributor. If it does, I'd just change this to
```julia
    @. ϕ_en_cov = is_tke ? FT(2 / 3) * covar_en : covar_en
    if is_tke
        @. ϕ_en_cov = FT(2 / 3) * covar_en
    else
        @. ϕ_en_cov = covar_en
    end
```

charleskawczynski (commented):
Overall the changes look great. I added a few notes regarding type stability / allocations

jiahe23 (approved):

PR #731, opened 2022-01-12
Use ClimaParameters c_gamma

ilopezgp:


charleskawczynski (commented):
LGTM. We need to update the `test/Project.toml` compat, too.

ilopezgp:
It seems we also need to update the ClimaParams compatibility in Thermodynamics

ilopezgp:
Need to fix SurfaceFluxes version as well...

Issue #638, opened 2021-12-10
Prep to inter-operate with dycore


charleskawczynski:
As we've checked off many items from our todo list / Project, it's clear that preparing to inter-operate with a dycore is becoming important. We need to make several structural code changes to provide a simple API to work with. Here's a WIP sketch of the 10k foot view:

```julia

# types
turbconv = TC.TurbConvModel(...)
physics = AtmosPhysics(; turbconv)
atmos = AtmosModel(; physics)

# methods
TC.compute_turbconv_tendencies!(tendencies, prog, atmos.turbconv, args...) = ...  # to compute tc tendencies, determined by tc
TC.compute_sgs_flux!(field, atmos.turbconv, args...) = ... # to compute gm tendencies, determined by tc
```

charleskawczynski:
# Functions to move outside of src
 - ∑tendencies!
 - compute_gm_tendencies!
 - Contents of `src/Surface.jl`
 - Contents of `src/Radiation.jl`
 - Contents of `src/ReferenceState.jl`
 - Contents of `src/Forcing.jl`


# Names
We should improve the names, so that our API is consistent (`turbconv` vs `edmf`, vs `EDMF_PrognosticTKE`)

# Documentation
We should add documentation specifically for the dycore API so that using this library involves a clear set of steps / requirements.

yairchn:
Following the recent changes that move NTECDF_IO out of src I am wondering about these two TC. calls in main.jl

https://github.com/CliMA/TurbulenceConvection.jl/blob/dc9c06599027846c2a4254bdf4061aef59f4c08b/driver/main.jl#L101
https://github.com/CliMA/TurbulenceConvection.jl/blob/dc9c06599027846c2a4254bdf4061aef59f4c08b/driver/main.jl#L193

The first one related to the io of the reference state - but both IO and reference state are moved out of src. 
The second one is the assignment of microphysics scheme - I think MP should be defined in driver and TC will inherit this definition - @Anna Jaruga?

charleskawczynski:
Sure, we can move `io_dictionary_ref_state` out, that should be trivial. Regarding `Clima1M`, we're dispatching on it in TC.jl

yairchn:
> Sure, we can move `io_dictionary_ref_state` out, that should be trivial. Regarding `Clima1M`, we're dispatching on it in TC.jl

[...]

PR #733, opened 2022-01-12
Move Radiation and Forcing to driver/

charleskawczynski:
Another contribution to #638.


driver/Radiation.jl
```diff
@@ -1,26 +1,32 @@
-update(self::RadiationBase, grid, state, gm::GridMeanVariables, param_set) = nothing
-initialize(self::RadiationBase{RadiationNone}, grid, state) = nothing
+update_radiation(self::TC.RadiationBase, grid, state, gm::GridMeanVariables, param_set) = nothing
```
yairchn, line 1: why do we need 'TC.RadiationBase' ?

charleskawczynski: We could move it out, but then we'll need to move out the related IO. Also, it looks like `radiation::RadiationBase` is an arg to `compute_sgs_tendencies!`, but it doesn't seem used. I'm fine with making these changes if that's the way to go.

yairchn, line 1: ODE needs to be available in this function

driver/Radiation.jl
```diff
@@ -65,22 +71,19 @@ function calculate_radiation(self::RadiationBase{RadiationDYCOMS_RF01}, grid, st
...
-    aux_gm = center_aux_grid_mean(state)
+    aux_gm = TC.center_aux_grid_mean(state)
     dTdt = NC.Dataset(LESDat.les_filename, "r") do data
         imin = LESDat.imin
         imax = LESDat.imax
 
         # interpolate here
         zc_les = Array(get_nc_data(data, "zc"))
-        meandata = mean_nc_data(data, "profiles", "dtdt_rad", imin, imax)
+        meandata = TC.mean_nc_data(data, "profiles", "dtdt_rad", imin, imax)
```
yairchn, line 83: needs to be 'TC.get_nc_data'

charleskawczynski: That's not what is currently called..

driver/Radiation.jl
charleskawczynski: ODE is a package, so I'm not sure what you mean

driver/Radiation.jl
charleskawczynski: Ah, I see, it's the call above that's missing. Thanks!

driver/Radiation.jl
charleskawczynski: I see, we needed to include OrdinaryDiffEq in Cases.jl

charleskawczynski: Let's tackle this on the next iteration

PR #735, opened 2022-01-13
 move stochastic_closures.jl to src/closures/

yairchn:
I think that it should be in closures folder


haakon-e:
I'll clean this up in a separate PR. I'm not using that closure anymore.

PR #736, opened 2022-01-13
Expose parameters, enforce consistency.

ilopezgp:
This PR:

- Fixes an inconsistency in the critical Richardson number, which was hardcoded in one of the functions.
- Exposes Ri_c and Pr_t for calibration by moving them to the EDMF_Turbulence_... dictionary. These are after all empirical values.
- Gets Ri_c and a few other parameters from ClimaParameters.
- Exposes kappa_star^2, which was previously hardcoded.


charleskawczynski (approved):
Changes look fine to me.

Issue #628, opened 2021-12-07
Slim dependencies


charleskawczynski:
We should really work on sliming down our dependencies. Of our dependencies, I think we really only need the following ones in our default project environment:

 - [ ] ArgParse
 - [ ] ArtifactWrappers
 - [x] CLIMAParameters
 - [x] ClimaCore
 - [x] CloudMicrophysics
 - [ ] Dates
 - [ ] Debugger
 - [ ] Dierckx
 - [ ] DifferentialEquations
 - [x] Distributions
 - [x] DocStringExtensions
 - [x] FastGaussQuadrature
 - [ ] Flux
 - [ ] Glob
 - [ ] JSON
 - [x] LambertW
 - [x] LinearAlgebra
 - [ ] Logging
 - [ ] NCDatasets
 - [ ] OrderedCollections
 - [x] OrdinaryDiffEq
 - [x] PoissonRandom
 - [ ] PrettyTables
 - [x] RootSolvers
 - [ ] SciMLBase
 - [x] StaticArrays
 - [x] Statistics
 - [x] StatsBase
 - [x] SurfaceFluxes
 - [ ] TerminalLoggers
 - [x] Thermodynamics
 - [x] UnPack

It would be nice if we could remove the unchecked deps, and add separate experiment environments for specific cases. This will improve time to first `using TurbulenceConvection`.

charleskawczynski:
One option here would be to add a new library package: `TurbulenceConvection.jl/lib/TurbulenceConvectionClosures.jl`, which imports TC and implements, for example, the closure(s) that use Flux. This would help make TC significantly more lightweight.

PR #738, opened 2022-01-16
Move `NetCDFIO_Stats` outside of src

charleskawczynski:
This PR moves `NetCDFIO_Stats` outside of `src/`, and removes TC.jl's dependence on `NCDatasets` (helping towards #628)


yairchn (approved):
LGTM great work!

PR #740, opened 2022-01-16
Remove Flux dep

charleskawczynski:
Here's an example of how we could remove our dependence on Flux.jl


charleskawczynski:
I'll retry this at some later point.

PR #741, opened 2022-01-18
revamp stochastic model design

haakon-e:
spring cleaning on stochastic model design for entrainment closures.

* add noisy relaxation process that is
compatible with any deterministic
mean-model
* add lognormal scaling process that
scales any deterministic mean-model

This PR retains time step in the tendencies to update the stochastic forcing


yairchn (approved):
LGTM

driver/generate_namelist.jl
```diff
@@ -154,7 +149,8 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
     namelist_defaults["turbulence"]["scheme"] = "EDMF_PrognosticTKE"
 
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["updraft_number"] = 1
-    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"] = "moisture_deficit" #{"moisture_deficit", "NN", "Linear"}
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"] = "moisture_deficit"  # {"moisture_deficit", "NN", "Linear"}
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["stochastic_entrainment"] = "deterministic"  # {"deterministic", "noisy_relaxation_process", "lognormal_scaling"}
```
charleskawczynski, line 153: Should this just be `namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"] = "stochastic_entr_detr"`?

haakon-e: the idea is that I can combine a stochastic entrainment model with a deterministic entrainment model.
So in the case where 
```
namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["stochastic_entrainment"] = "noisy_relaxation_process"
```
then the value of
```
namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"]
```
will determine the mean-behavior of the stochastic process. This way, I can code the stochastic closure once, and experiment with any current or future deterministic closures as the mean-behavior of the stochastic process.

charleskawczynski: Ah, ok

PR #746, opened 2022-01-19
Add DYCOMS_RF02 case

trontrytel:
This adds a drizzling Sc test case. Currently the LES data is generated with SB microphysics scheme. Once I'm done with implementing the CliMA 1-moment scheme we should switch the dataset

### DIFF
--- a/integration_tests/DYCOMS_RF02.jl
+++ b/integration_tests/DYCOMS_RF02.jl
@@ -1,0 +1,40 @@
+import Pkg
+Pkg.develop(path = ".")
+import TurbulenceConvection
+using Test
+
+const tc_dir = dirname(dirname(pathof(TurbulenceConvection)))
+include(joinpath(tc_dir, "driver", "main.jl"))
+include(joinpath(tc_dir, "driver", "generate_namelist.jl"))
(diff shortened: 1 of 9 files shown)


integration_tests/DYCOMS_RF02.jl
```diff
@@ -0,0 +1,41 @@
+if !("." in LOAD_PATH) # for easier local testing
+    push!(LOAD_PATH, ".")
+end
```
charleskawczynski, line 3:
```suggestion
import Pkg
Pkg.develop(path = ".")
```

charleskawczynski (approved):
LGTM.

This would be a great PR to document, somehow, since it cleanly adds a new experiment, and all of the separate pieces that are needed.

trontrytel:
> LGTM.
> 
> This would be a great PR to document, somehow, since it cleanly adds a new experiment, and all of the separate pieces that are needed.

Should I write down the commentary as a comment to this PR?

charleskawczynski:
> Should I write down the commentary as a comment to this PR?

I think it's fine, we can easily grab the commit if/when we add such docs

yairchn (approved):
LGTM

PR #748, opened 2022-01-19
Add mwe Rico

charleskawczynski:
This PR adds a minimum working example (MWE) for Rico as a separate case, to be used for documentation purposes. The "MWE" is still 1500+ lines, so we've got some work to do on further trimming things down so that this looks like a sane driver example.

### DIFF
--- a/integration_tests/mwe_Rico.jl
+++ b/integration_tests/mwe_Rico.jl
@@ -1,0 +1,1616 @@
+if !haskey(ENV, "BUILDKITE")
+    import Pkg
+    Pkg.develop(Pkg.PackageSpec(; path = dirname(@__DIR__)))
+end
+import TurbulenceConvection
+const TC = TurbulenceConvection
+const tc_dir = dirname(dirname(pathof(TurbulenceConvection)))
+
+import UnPack
+
+import StaticArrays
+import OrderedCollections
+const SA = StaticArrays
+
+import SurfaceFluxes
+const SF = SurfaceFluxes
+const UF = SF.UniversalFunctions
+
+import CLIMAParameters
+const CP = CLIMAParameters
+const CPP = CP.Planet
+const APS = CP.AbstractEarthParameterSet
+
+import StatsBase
+
+import ClimaCore
+const CC = ClimaCore
+const CCO = CC.Operators
+
+import Thermodynamics
+const TD = Thermodynamics
+
+using Test
+
+import NCDatasets
+const NC = NCDatasets
+
+import OrdinaryDiffEq
+const ODE = OrdinaryDiffEq
+
+import Dierckx
+import Statistics
+import Random
+
+import Logging
+import TerminalLoggers
+Logging.global_logger(TerminalLoggers.TerminalLogger())
+
+import JSON
+import ArgParse
+
+import CloudMicrophysics
+const CM = CloudMicrophysics
+const CM0 = CloudMicrophysics.Microphysics_0M
+const CM1 = CloudMicrophysics.Microphysics_1M
+
+import SciMLBase
+import JSON
+
+mutable struct NetCDFIO_Stats
+    root_grp::NC.NCDataset{Nothing}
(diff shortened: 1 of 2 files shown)


integration_tests/mwe_Rico.jl
```diff
@@ -0,0 +1,1570 @@
...
+    #####
+    ##### Fluxes
+    #####
+    Ic = CCO.InterpolateF2C()
+    parent(diag_tc.massflux) .= 0
+    @inbounds for i in 1:N_up
+        @. diag_tc.massflux += Ic(aux_up_f[i].massflux)
+    end
+
+    @inbounds for k in TC.real_center_indices(grid)
```
yairchn, line 436: I think this (lines 436 - 470) should be in src

integration_tests/mwe_Rico.jl
```diff
@@ -0,0 +1,1570 @@
...
+        @show param_set
+        error("The parameter set MUST be isbits in order to be stack-allocated.")
+    end
+    return param_set
+end
+#! format: on
+
+struct Rico end
+
+function ForcingBase(param_set::APS)
```
yairchn, line 617: do we need 'ForcingBase' here?

integration_tests/mwe_Rico.jl
```diff
@@ -0,0 +1,1570 @@
...
+    cfl_limit = namelist["time_stepping"]["cfl_limit"]
+    dt_min = namelist["time_stepping"]["dt_min"]
+
+    grid = TC.Grid(FT(namelist["grid"]["dz"]), namelist["grid"]["nz"])
+    Stats = skip_io ? nothing : NetCDFIO_Stats(namelist, grid)
+    ref_params = reference_params(grid, param_set, namelist)
+
+    gm = TC.GridMeanVariables(param_set)
+    Fo = ForcingBase(param_set)
+    Rad = TC.RadiationBase{TC.RadiationNone}()
```
yairchn, line 1020: Radiation should be outside src

integration_tests/mwe_Rico.jl
```diff
@@ -0,0 +1,1570 @@
...
+        tendencies_gm.u[k] -= force.coriolis_param * (aux_gm.vg[k] - prog_gm.v[k])
+        tendencies_gm.v[k] += force.coriolis_param * (aux_gm.ug[k] - prog_gm.u[k])
+
+        tendencies_gm.θ_liq_ice[k] -= ∇θ_liq_ice_gm[k] * aux_gm.subsidence[k]
+        tendencies_gm.q_tot[k] -= ∇q_tot_gm[k] * aux_gm.subsidence[k]
+
+        tendencies_gm.θ_liq_ice[k] += aux_gm.dTdt[k] / Π
+        tendencies_gm.q_tot[k] += aux_gm.dqtdt[k]
+
+        tendencies_gm.q_tot[k] +=
```
yairchn, line 893: I think totes should be out of src, the driver should be exposed to the updraft and env, maybe we need a microphysics coupling function to handle this, @trontrytel ?

integration_tests/mwe_Rico.jl
```diff
@@ -0,0 +1,1570 @@
...
+            aux_bulk.qt_tendency_precip_formation[k] +
+            aux_en.qt_tendency_precip_formation[k] +
+            aux_tc.qt_tendency_precip_sinks[k]
+        tendencies_gm.θ_liq_ice[k] +=
+            aux_bulk.θ_liq_ice_tendency_precip_formation[k] +
+            aux_en.θ_liq_ice_tendency_precip_formation[k] +
+            aux_tc.θ_liq_ice_tendency_precip_sinks[k]
+    end
+    TC.compute_sgs_tendencies!(edmf, grid, state, surf, radiation, force, gm)
```
yairchn, line 902: we need to think of a manner in which this function can be informed of a TurbConvModel which could be EDMF (in that case it would added the SGS fluxes) or None (in which case it adds nothing) or another TurbConvModel. The TurbConvModel should be passed via 'TC.compute_gm_tendencies' and via '∑tendencies'

PR #749, opened 2022-01-19
Allow for non-local entr-detr models

charleskawczynski:
This PR preps some of the entr-detr modeling.

We are duplicating some code here, we could alternatively change the original code to accept the pi groups, if that would help simplify the second part of the entr-detr closure. Regardless, this should help encapsulate / overload the existing entr-detr closures

### DIFF
--- a/src/update_aux.jl
+++ b/src/update_aux.jl
@@ -45,8 +45,6 @@ function update_aux!(
     aux_en_sat = aux_en.sat
     m_entr_detr = aux_tc.ϕ_temporary
     ∇m_entr_detr = aux_tc.ψ_temporary
-    w_up_c = aux_tc.w_up_c
-    w_en_c = aux_tc.w_en_c
     wvec = CC.Geometry.WVector
     max_area = edmf.max_area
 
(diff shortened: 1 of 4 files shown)


costachris:
It may be more straightforward to have a single column-wise `compute_entr_detr` function so that it's easier to expose the full vertical profile to other closures. Then we can dispatch on `entr_detr` or `non_dimensional_function`, giving the full profile to the FNO function but applying the others point-wise.

src/update_aux.jl
```diff
@@ -244,84 +242,18 @@ function update_aux!(
...
-                    nondim_detr_sc = aux_up[i].nondim_detr_sc[k],
-                )
-
-                er = entr_detr(param_set, εδ_model_vars, edmf.entr_closure)
-                aux_up[i].entr_sc[k] = er.ε_dyn
-                aux_up[i].detr_sc[k] = er.δ_dyn
-                aux_up[i].frac_turb_entr[k] = er.ε_turb
-                # nondim entr/detr
-                aux_up[i].nondim_entr_sc[k] = er.ε_dyn_nondim
-                aux_up[i].nondim_detr_sc[k] = er.δ_dyn_nondim
```
haakon-e, line 306: It seems like this PR removed this update, which, until we potentially get prognostic entr/detr will be needed for me/

charleskawczynski: Ah, whoops, sorry, I must have messed up rebasing after it was added.

PR #764, opened 2022-01-26
compute sgs flux in src and tendency in dycore

yairchn:
In this PR I am adding an sgs_flux variable per dycore prognostic variable. This flux is to be update by a SGS scheme, while the flux divergence (tendency) is computed by the dycore with other dycore tendencies.


driver/dycore.jl
```diff
@@ -63,7 +63,16 @@ cent_aux_vars_gm(FT) = (;
 cent_aux_vars(FT, n_up) = (; aux_vars_ref_state(FT)..., cent_aux_vars_gm(FT)..., TC.cent_aux_vars_edmf(FT, n_up)...)
 
 # Face only
-face_aux_vars_gm(FT) = (; massflux_s = FT(0), diffusive_flux_s = FT(0), total_flux_s = FT(0), f_rad = FT(0))
+face_aux_vars_gm(FT) = (;
+    massflux_s = FT(0),
```
ilopezgp, line 67: what is the `_s` at the end? sgs?

driver/dycore.jl
```diff
@@ -340,6 +352,24 @@ function compute_gm_tendencies!(
...
+    aeKMu_bc = surf.ρu_flux / a_en[kc_surf]
+    aeKMv_bc = surf.ρv_flux / a_en[kc_surf]
+
+    tbc = (; top = CCO.SetValue(wvec(FT(0))))
+    ∇aeKH_q_tot = CCO.DivergenceF2C(; bottom = CCO.SetValue(wvec(aeKHq_tot_bc)), tbc...)
+    ∇aeKH_θ_liq_ice = CCO.DivergenceF2C(; bottom = CCO.SetValue(wvec(aeKHθ_liq_ice_bc)), tbc...)
+    ∇aeKM_u = CCO.DivergenceF2C(; bottom = CCO.SetValue(wvec(aeKMu_bc)), tbc...)
+    ∇aeKM_v = CCO.DivergenceF2C(; bottom = CCO.SetValue(wvec(aeKMv_bc)), tbc...)
+
+    @. tendencies_gm.θ_liq_ice += -α0_c * ∇aeKH_θ_liq_ice(wvec(aux_gm_f.sgs_flux_θ_liq_ice))
```
ilopezgp, line 368: This says ∇aeKH_θ_liq_ice, but we are passing aeKH_θ_liq_ice + mass flux term. It is misleading, I think.

driver/dycore.jl
```diff
@@ -340,6 +352,24 @@ function compute_gm_tendencies!(
             aux_en.θ_liq_ice_tendency_precip_formation[k] +
             aux_tc.θ_liq_ice_tendency_precip_sinks[k]
     end
+
+    a_en = aux_en.area
```
ilopezgp, line 356: So, the Dycore needs access to the environmental area? Doesn't seem right.

driver/dycore.jl
```diff
@@ -340,6 +352,24 @@ function compute_gm_tendencies!(
...
+    aeKHq_tot_bc = surf.ρq_tot_flux / a_en[kc_surf]
+    aeKHθ_liq_ice_bc = surf.ρθ_liq_ice_flux / a_en[kc_surf]
+    aeKMu_bc = surf.ρu_flux / a_en[kc_surf]
+    aeKMv_bc = surf.ρv_flux / a_en[kc_surf]
```
ilopezgp, line 360: Now that I think of it. Does this make sense? The way it is written, we have

ρ(w'u')_surf = ae^2 * K_m * u_bc

Which should be

ρ(w'u')_surf = ρ * ae * K_m * ∂(u)/∂z_surf

Even if we are defining u_bc = ∂(u)/∂z_surf (which I am not sure of), what about the density and the double area fraction?

yairchn: this is a name problem not more. 
it should be called 'sgs_q_tot_bc'

yairchn:
@ilopezgp thanks for your comments , the PR is updated - see if you can take a look

src/Turbulence_PrognosticTKE.jl
```diff
@@ -90,42 +91,30 @@ function compute_sgs_tendencies!(
...
-    @. tends_v += -α0_c * ∇aeKM_v(wvec(diffusive_flux_v))
+    @. aux_gm_f.sgs_flux_θ_liq_ice = diffusive_flux_h + massflux_h
+    @. aux_gm_f.sgs_flux_q_tot = diffusive_flux_qt + massflux_qt
+    @. aux_gm_f.sgs_flux_u = diffusive_flux_u # + massflux_u
+    @. aux_gm_f.sgs_flux_v = diffusive_flux_v # + massflux_v
+    # apply surface BC
+    aux_gm_f.sgs_flux_θ_liq_ice[kf_surf] = surf.ρθ_liq_ice_flux / a_en[kc_surf]
+    aux_gm_f.sgs_flux_q_tot[kf_surf] = surf.ρq_tot_flux / a_en[kc_surf]
+    aux_gm_f.sgs_flux_u[kf_surf] = surf.ρu_flux / a_en[kc_surf]
+    aux_gm_f.sgs_flux_v[kf_surf] = surf.ρv_flux / a_en[kc_surf]
```
ilopezgp, line 117: With the updated notation, this illustrates perfectly the issue. We are inputting an energy flux at the surface (e.g., surf.ρθ_liq_ice_flux), which is irrespective of the SGS model. Then, we are dividing it by a_en before putting that energy into the GCM! We are actually multiplying the input energy flux by (1/a_en) > 1.

The `a_en` values should be deleted from here. They should only be applied to the **environmental flux**, which is **the flux within the environment**.

driver/dycore.jl
ilopezgp: I am fine with it if @charleskawczynski is.

src/Turbulence_PrognosticTKE.jl
ilopezgp: Since we do not have environmental equations for θ_liq_ice, q_tot, u or v, this means it should not appear. It should, and will, when we create flux BCs for TKE (coming soon).

driver/dycore.jl
charleskawczynski: I'm fine with leaving it for a later PR. There's probably a decent effort required to move more things out of aux into diagnostics.

src/Turbulence_PrognosticTKE.jl
yairchn: in that case I moved the surface flux to the dycore, it is the more correct code implementation. The surface layer flux is the SGS flux at the lowest level after all

ilopezgp (approved):

yairchn:
Bors r+

yairchn:
Bors r+

charleskawczynski (approved):

driver/dycore.jl
```diff
@@ -340,6 +352,28 @@ function compute_gm_tendencies!(
...
+
+    ∇θ_liq_ice_sgs = CCO.DivergenceF2C()
+    ∇q_tot_sgs = CCO.DivergenceF2C()
+    ∇u_sgs = CCO.DivergenceF2C()
+    ∇v_sgs = CCO.DivergenceF2C()
+
+    @. tends_θ_liq_ice += -α0_c * ∇θ_liq_ice_sgs(wvec(aux_gm_f.sgs_flux_θ_liq_ice))
+    @. tends_q_tot += -α0_c * ∇q_tot_sgs(wvec(aux_gm_f.sgs_flux_q_tot))
+    @. tends_u += -α0_c * ∇u_sgs(wvec(aux_gm_f.sgs_flux_u))
+    @. tends_v += -α0_c * ∇v_sgs(wvec(aux_gm_f.sgs_flux_v))
```
charleskawczynski, line 377: Let's hoist the `getproperty` calls to avoid adding inference triggers. Please see this [ClimaCore issue](https://github.com/CliMA/ClimaCore.jl/issues/341).

src/Turbulence_PrognosticTKE.jl
```diff
@@ -90,42 +91,23 @@ function compute_sgs_tendencies!(
...
     diffusive_flux_v = aux_tc_f.diffusive_flux_v
 
-    @. tends_q_tot += -α0_c * ∇aeKH_q_tot(wvec(diffusive_flux_qt))
-    @. tends_θ_liq_ice += -α0_c * ∇aeKH_θ_liq_ice(wvec(diffusive_flux_h))
-    @. tends_u += -α0_c * ∇aeKM_u(wvec(diffusive_flux_u))
-    @. tends_v += -α0_c * ∇aeKM_v(wvec(diffusive_flux_v))
+    @. aux_gm_f.sgs_flux_θ_liq_ice = diffusive_flux_h + massflux_h
+    @. aux_gm_f.sgs_flux_q_tot = diffusive_flux_qt + massflux_qt
+    @. aux_gm_f.sgs_flux_u = diffusive_flux_u # + massflux_u
+    @. aux_gm_f.sgs_flux_v = diffusive_flux_v # + massflux_v
```
charleskawczynski, line 110: Let's hoist the `getproperty` calls to avoid adding inference triggers. Please see this [ClimaCore issue](https://github.com/CliMA/ClimaCore.jl/issues/341).

yairchn:
Bors r+

yairchn:
Bors r-

yairchn:
Bors r+

yairchn:
Bors r+

yairchn:
Bors r+

PR #766, opened 2022-01-28
add FNO entr detr function

yairchn:
This PR merges the working version of FNO entr closure into main.

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -105,11 +105,9 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
                    0.0, 0.0, 0.843,-0.340,
                    0.655, 0.113, 0.0, 0.0)
     # For FNO add here
-    # namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_FNO_params"] =
-    #     SA.SVector(0.3038, 0.719,-0.910,-0.483,
-    #                0.739, 0.0755, 0.178, 0.521,
-    #                0.0, 0.0, 0.843,-0.340,
-    #                0.655, 0.113, 0.0, 0.0)
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["fno_ent_params"] =
+        SA.SVector{74}(rand(74))
+
     # General stochastic entrainment/detrainment parameters
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_stochastic_ent_params"] =
     SA.SVector(
--- a/src/closures/nondimensional_exchange_functions.jl
+++ b/src/closures/nondimensional_exchange_functions.jl
@@ -75,17 +75,36 @@ function non_dimensional_function!(
     Π₄::AbstractArray{FT}, # input
     εδ_model::FNOEntr,
 ) where {FT <: Real}
-    # c_gen_fno = ICP.c_gen_fno(param_set)
 
-    # Π = (Π₁ ,Π₂ ,Π₃ ,Π₄)
-    # trafo = OF.FourierTransform(modes = (8, 8))
-    # model = Chain(
-    #     SpectralKernelOperator(trafo, 4 => 4, gelu),
-    #     SpectralKernelOperator(trafo, 4 => 2, gelu),
-    # )
-    # output = model(Π, c_gen_fno)
-    # nondim_ε .= output[1,:]
-    # nondim_δ .= output[2,:]
+    # define the model
+    Π = hcat(Π₁, Π₂, Π₃, Π₄)'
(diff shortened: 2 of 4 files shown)


costachris (approved):

driver/generate_namelist.jl
```diff
@@ -105,11 +105,9 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
                    0.0, 0.0, 0.843,-0.340,
                    0.655, 0.113, 0.0, 0.0)
     # For FNO add here
-    # namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_FNO_params"] =
-    #     SA.SVector(0.3038, 0.719,-0.910,-0.483,
-    #                0.739, 0.0755, 0.178, 0.521,
-    #                0.0, 0.0, 0.843,-0.340,
-    #                0.655, 0.113, 0.0, 0.0)
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["fno_ent_params"] =
+        SA.SVector{74}(rand(74))
```
charleskawczynski, line 109: cc @simonbyrne, @odunbar 😂

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -75,17 +75,32 @@ function non_dimensional_function!(
...
+    for p in Flux.params(model)
+        if eltype(p[:]) == Float32
+            p[:] = c_fno[index:(index + length(p) - 1)]
+            index = index + length(p)
+        end
+        if eltype(p[:]) == ComplexF32
+            p[:] = c_fno[index:(index + length(p) - 1)] + c_fno[(index + length(p)):(index + length(p) * 2 - 1)] * im
+            index = index + length(p) * 2
+        end
+    end
```
charleskawczynski, line 97: It looks like `p` is unused

charleskawczynski, line 97:
```suggestion
    c_fno = ICP.c_fno(param_set)
    index = 1
    for p in Flux.params(model)
        len_p = length(p)
        p_slice = p[:]
        c_fno_slice = c_fno[index:(index + len_p - 1)]
        if eltype(p_slice) <: Real
            p_slice .= c_fno_slice
            index += len_p
        end
        if eltype(p[:]) <: Complex
            p_slice .= c_fno_slice + c_fno[(index + len_p):(index + len_p * 2 - 1)] * im
            index += len_p * 2
        end
    end
```

driver/generate_namelist.jl
simonbyrne: Do these parameters have any meaning, or are they just NN weights?

yairchn: just NN weights

simonbyrne: How big are these going to want to get? We may want to reconsider how we handle these.

zongyi-li: I highly suggest to load/save from a text file or something. In future, it can be thousands and millions of parameters. We may want to first pre-train the network parameters to the physics closure offline, and then load them into the solver.

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -75,17 +75,36 @@ function non_dimensional_function!(
...
+    trafo = OF.FourierTransform(modes = (2,))
+    model = OF.Chain(OF.SpectralKernelOperator(trafo, 4 => 2, Flux.relu),)
+
+    # set the parameters
+    c_fno = ICP.c_fno(param_set)
+    index = 1
+    for p in Flux.params(model)
+        len_p = length(p)
+        p_slice = p[:]
+        c_fno_slice = c_fno[index:(index + len_p - 1)]
```
zongyi-li, line 91: Maybe we can move line 91 into the if-else argument? Ideally we can treat the real and complex parameter in the same manner. Especially line 96 makes it look like we don't need to define c_fno_slice.

charleskawczynski: `c_fno_slice` is used in both the `if` and `else` block. Ah, but I guess we're slicing the same once in both cases. That's fine with me if you'd like.

zongyi-li: yes, maybe for the real part, `c_fno_slice = c_fno[index:(index + len_p - 1)]`, and for the complex part `c_fno_slice = c_fno[index:(index + 2*len_p - 1)]`

driver/generate_namelist.jl
yairchn: That will have to wait for a future PR , I doubt it if there is a point an assigning a ML closure with thousands / millions or parameters to replace a physics one that has 4. But that is another discussion

PR #767, opened 2022-01-28
Widen allocation monitoring

charleskawczynski:
This PR widens the allocation monitoring to the entire `step!` function. To see allocations outside of TC.jl, we'll need to send additional directories to Coverage.jl, and it's a bit tricky due to how environments are configured, so I'll leave that for later.


charleskawczynski:
I had to update the tolerances for the precompilable scripts, since we're now excluding any initialization inside `solve`, which seems reasonable.

PR #774, opened 2022-01-31
Add perf benchmark for ODE.step

charleskawczynski:


charleskawczynski:
Something weird is going on with the logging, it's printing way too much. Going to fix that first.

charleskawczynski:
Very interesting.

PR #776, opened 2022-02-01
Remove some TC. calls in driver

yairchn:
In this PR I have moved the initialization of EDMF variables into src while concentrating them in a 'initialize_turbconv' for a single call in driver. In the future 'initialize_turbconv' could be dispatched for EDMF or another model

### DIFF
--- a/driver/initial_conditions.jl
+++ b/driver/initial_conditions.jl
@@ -3,7 +3,7 @@ const TC = TurbulenceConvection
 import Thermodynamics
 const TD = Thermodynamics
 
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Why do the mse tables change?

yairchn:
> Why do the mse tables change?

now they are not changed

charleskawczynski:
I guess this is fine, I still don't really understand why we want to put ICs / case-specific code inside src/

ilopezgp:
I don't think this should be in source

PR #778, opened 2022-02-01
fix unit bug in initialize_covariance

yairchn:
There is a bug in initialize_covariance that the initial profiles of thermodynamics covariances has a factor TKE in the units


driver/initial_conditions.jl
```diff
@@ -42,14 +43,7 @@ function initialize_covariance(edmf::TC.EDMF_PrognosticTKE, grid::TC.Grid, state
     aux_bulk = TC.center_aux_bulk(state)
     ae = 1 .- aux_bulk.area # area of environment
 
-    aux_en.tke .= aux_gm.tke
     prog_en.ρatke .= aux_en.tke .* ρ0_c .* ae
-
```
yairchn, line 47: the grid mean should be initialized in Cases like all other grid mean profiles. The env con should be here

haakon-e (approved):
I requested buildkite to rerun Soares for good measure, but LGTM

costachris (approved):

PR #779, opened 2022-02-01
call TD.PhaseEquil_pθq outside of src

yairchn:
Outside of src we should not use 'TC.thermo_state_pθq' but rather directly use 'TD.PhaseEquil_pθq'.


charleskawczynski (approved):

yairchn:
Bors r+

PR #780, opened 2022-02-01
Add nonlocal NN entr closure, fix entr bugs

costachris:
This PR: 
         1.) Adds a nonlocal NN entr closure type
         2.) Adds outputs to stats file needed to compute Pi groups offline and validate new closures
         3.) Addresses division by 0 in one of the Pi groups 
         4.) Fixes a bug affecting several entrainment-related outputs.

### DIFF
--- a/driver/compute_diagnostics.jl
+++ b/driver/compute_diagnostics.jl
@@ -36,7 +36,9 @@ function io_dictionary_diagnostics()
         "nh_pressure_b" => (; dims = ("zf", "t"), group = "profiles", field = state -> face_diagnostics_turbconv(state).nh_pressure_b,),
         "turbulent_entrainment" => (; dims = ("zc", "t"), group = "profiles", field = state -> center_diagnostics_turbconv(state).frac_turb_entr,),
         "entrainment_sc" => (; dims = ("zc", "t"), group = "profiles", field = state -> center_diagnostics_turbconv(state).entr_sc),
+        "nondim_entrainment_sc" => (; dims = ("zc", "t"), group = "profiles", field = state -> center_diagnostics_turbconv(state).nondim_entr_sc),
         "detrainment_sc" => (; dims = ("zc", "t"), group = "profiles", field = state -> center_diagnostics_turbconv(state).detr_sc),
+        "nondim_detrainment_sc" => (; dims = ("zc", "t"), group = "profiles", field = state -> center_diagnostics_turbconv(state).nondim_detr_sc),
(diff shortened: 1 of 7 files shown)


driver/compute_diagnostics.jl
```diff
@@ -235,11 +237,19 @@ function compute_diagnostics!(
 
     @inbounds for k in TC.real_center_indices(grid)
         a_up_bulk_k = a_up_bulk[k]
+        diag_tc.entr_sc[k] = 0.0
```
costachris, line 240: I zeroed the fields in `diag_tc` here because the `+=` logic was adding to values in previous time steps. `diag_tc` retains memory of previous timesteps, so perhaps other fields that are computed in a similar manner will also need to be zeroed (ie `nh_pressure` fields around L274)? or maybe there's a more graceful way to handle this than what I implemented.

src/closures/entr_detr.jl
```diff
@@ -151,12 +150,6 @@ function compute_entr_detr!(
                 aux_up[i].frac_turb_entr[k] = er.ε_turb
                 aux_up[i].nondim_entr_sc[k] = er.ε_dyn_nondim
                 aux_up[i].nondim_detr_sc[k] = er.δ_dyn_nondim
-            else
-                aux_up[i].entr_sc[k] = 0.0
```
costachris, line 155: These `aux_up` fields are already initialized with zeros, so this isn't needed

driver/compute_diagnostics.jl
charleskawczynski: Yes, can you please fix `nh_pressure` too? Simulations have still been stable because this is just for diagnostics.

yairchn: great catch! luckily 'diag_tc.entr_sc[k]' is only for diagnostics, it is not used in any equation :-). only the updraft specific variables enter the updraft equations.

yairchn (approved):
LGTM

src/closures/entr_detr.jl
costachris: I'll try to see where it's coming from

charleskawczynski (approved):

driver/compute_diagnostics.jl
costachris: @charleskawczynski is there a problem with switching the `k`, `i` loops for the `nh_pressure` fields?

costachris:
I don't have a good sense for what `nh_pressure` should look like, but this is how the diagnostic changes with the proposed bug fix. 
Screen Shot 2022-02-01 at 3 35 53 PM

PR #781, opened 2022-02-01
use the businger Pr_0 instead as they are identical

yairchn:
Going over the parameter list I noticed that Pr_n in EDMF is Pr_0 of the Businger type universal function. I am changing to use the Businger one and would ask to remove Pr_n from ClimaParameters


yairchn:
in that case I will close this PR and we will have to parameters with a similar name and same value (for now)

PR #784, opened 2022-02-02
Make dim scale for entrainment a namelist option, add 1/z scale

costachris:
Make dim scale for entrainment a namelist option, add 1/z scale

### DIFF
--- a/src/closures/entr_detr.jl
+++ b/src/closures/entr_detr.jl
@@ -35,10 +35,32 @@ function get_MdMdz(M::FT, dMdz::FT) where {FT}
     return MdMdz_ε, MdMdz_δ
 end
 
-function entrainment_length_scale(param_set, b_up::FT, b_en::FT, w_up::FT, w_en::FT, tke::FT) where {FT}
+function entrainment_length_scale(
+    param_set,
+    b_up::FT,
+    b_en::FT,
+    w_up::FT,
+    w_en::FT,
+    tke::FT,
+    zc_i::FT,
+    ::BuoyVelEntrDimScale,
+) where {FT}
     Δw = get_Δw(param_set, w_up, w_en)
     λ = compute_inverse_timescale(param_set, b_up, b_en, w_up, w_en, tke)
     return (λ / Δw)
+end
+
+function entrainment_length_scale(
+    param_set,
+    b_up::FT,
+    b_en::FT,
+    w_up::FT,
+    w_en::FT,
+    tke::FT,
+    zc_i::FT,
+    ::InvZEntrDimScale,
+) where {FT}
+    return (1 / zc_i)
 end
 
 """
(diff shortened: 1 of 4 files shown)


src/closures/entr_detr.jl
```diff
@@ -268,6 +299,8 @@ function compute_entr_detr!(
                 w_up_c[k],
                 w_en_c[k],
                 aux_en.tke[k],
+                grid.zc[k].z,
+                edmf.entr_dim_scale,
             )
             area_limiter = max_area_limiter(param_set, max_area, aux_up[i].area[k])
```
costachris, line 305: Regarding the cutoffs discussed this morning, do we need to remove the `area_limiter`? That'd be behavior changing for all the test cases. Otherwise there's 2 more cutoffs in for the default inverse length scale: `lamb_smooth_minimum(l, FT(0.1), FT(0.0005))` and the minimum updraft velocity. (in `entr_detr.jl` L23 and L28).

yairchn (approved):

src/closures/entr_detr.jl
```diff
@@ -35,12 +35,34 @@ function get_MdMdz(M::FT, dMdz::FT) where {FT}
...
+    param_set,
+    b_up::FT,
+    b_en::FT,
+    w_up::FT,
+    w_en::FT,
+    tke::FT,
+    zc_i::FT,
+    ::InvZEntrDimScale,
+) where {FT}
+    return (1.0 / zc_i)
```
charleskawczynski, line 63:
```suggestion
    return (1 / zc_i)
```

src/closures/entr_detr.jl
```diff
@@ -51,9 +73,18 @@ Cohen et al. (JAMES, 2020), given:
...
+    dim_scale = entrainment_length_scale(
+        param_set,
+        εδ_vars.b_up,
+        εδ_vars.b_en,
+        εδ_vars.w_up,
+        εδ_vars.w_en,
+        εδ_vars.tke,
+        εδ_vars.zc_i,
+        entr_dim_scale,
+    )
```
charleskawczynski, line 87: Hmm, I suppose that, if we're not going to decompose `GeneralizedEntr`, we might as well just pass `εδ_vars` through here. @yairchn, what do you think?

src/closures/entr_detr.jl
costachris: the area limiter term in detrainment can effectively be turned off by setting the `area_limiter_power ` namelist arg to `0`. So I'll leave that as is and we can turn it on/off as needed when calibrating.

src/closures/entr_detr.jl
costachris: We can do that, although `εδ_model_vars` would have to be constructed again [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/17c5037fe0635e716a63966cf9c053a883d58b99/src/closures/entr_detr.jl#L295) before calling `entrainment_length_scale` in that case

charleskawczynski: Oh, I see. Hmm, ok, maybe better to just leave it then

src/closures/entr_detr.jl
```diff
@@ -35,12 +35,34 @@ function get_MdMdz(M::FT, dMdz::FT) where {FT}
...
+    b_up::FT,
+    b_en::FT,
+    w_up::FT,
+    w_en::FT,
+    tke::FT,
+    zc_i::FT,
+    ::InvZEntrDimScale,
+) where {FT}
+    return (1 / zc_i)
+end
```
charleskawczynski, line 64: @costachris, it just occurred to me that we could have done this slightly differently. Not a big deal, but we could have made `InvZEntrDimScale` / `BuoyVelEntrDimScale` the first argument and then wrote this method as

```julia
entrainment_length_scale(::InvZEntrDimScale, args...) = (1 / zc_i)
```

The one above would still need to list all the args, but you know, this one would look nicer

PR #794, opened 2022-02-08
Random Feature Entr Detr Local Closure

nickhnelsen:
## In this PR

- A simple Random Fourier Feature approximation to a scalar-valued GP with squared exponential kernel and four inverse lengthscales (one for each Pi group input)
- Separates optimizable (expansion coefficients and hyperparameters) from fixed (realizations of random variables) parameters in the namelist

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -36,6 +36,8 @@ const SA = StaticArrays
 import ArtifactWrappers
 const AW = ArtifactWrappers
 
+import Random
+
 function les_driven_scm_data_folder()
     #! format: off
     LESDrivenSCM_output_dataset = AW.ArtifactWrapper(
@@ -72,7 +74,11 @@ function default_namelist(::Nothing)
     return default_namelist(case_name)
 end
 
-function default_namelist(case_name::String; root::String = ".", write::Bool = true)
+function default_namelist(case_name::String; root::String = ".", write::Bool = true, set_seed::Bool = true)
+
+    if set_seed
+        Random.seed!(2022)
+    end
 
     namelist_defaults = Dict()
     namelist_defaults["meta"] = Dict()
@@ -109,6 +115,17 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
     # For FNO add here
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["fno_ent_params"] =
         SA.SVector{74}(rand(74))
+
+    # m=100 random features, d=4 input Pi groups
+    # RF: parameters to optimize, 2 x (m + 1 + d)
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["rf_opt_ent_params"] =
+        vec(cat(sqrt(100) * randn(2,100),
+                    ones(2,5), dims=2))
+
+    # RF: fixed realizations of random variables, 2 x m x (1 + d)
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["rf_fix_ent_params"] =
+        vec(cat(2*pi*rand(2,100,1),
+                    randn(2,100,4), dims=3))
 
     # General stochastic entrainment/detrainment parameters
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_stochastic_ent_params"] =
@@ -155,7 +172,7 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
     namelist_defaults["turbulence"]["scheme"] = "EDMF_PrognosticTKE"
 
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["updraft_number"] = 1
-    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"] = "moisture_deficit"  # {"moisture_deficit", "NN", "NN_nonlocal", "Linear", "FNO"}
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"] = "moisture_deficit"  # {"moisture_deficit", "NN", "NN_nonlocal", "Linear", "FNO", "RF"}
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entr_dim_scale"] = "buoy_vel" # {"buoy_vel", "inv_z"}
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["stochastic_entrainment"] = "deterministic"  # {"deterministic", "noisy_relaxation_process", "lognormal_scaling"}
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["use_local_micro"] = true
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -274,7 +274,13 @@ nc_results_file(::Nothing) = @info "The simulation was run without IO, so no nc 
 
 function main1d(namelist; time_run = true)
     # TODO: generalize conversion of arrays from namelist to `SVector`s.
-    for param_name in ["general_ent_params", "general_stochastic_ent_params", "fno_ent_params"]
+    for param_name in [
+        "general_ent_params",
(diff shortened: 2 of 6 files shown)


src/closures/nondimensional_exchange_functions.jl
```diff
@@ -193,6 +193,41 @@ function non_dimensional_function(param_set, εδ_model_vars, ::LinearEntr)
...
+    scale_x_entr = abs.(c_rf_opt[1,m+2:m+d+1]).*nondim_groups
+    scale_x_detr = abs.(c_rf_opt[2,m+2:m+d+1]).*nondim_groups
+
+    # Features
+    f_entr = c_rf_opt[1,m+1]*sqrt(2)*cos.(c_rf_fix[1,:,2:d+1]*scale_x_entr + c_rf_fix[1,:,1])
+    f_detr = c_rf_opt[2,m+1]*sqrt(2)*cos.(c_rf_fix[2,:,2:d+1]*scale_x_detr + c_rf_fix[2,:,1])
+
+    # Predict and square output for nonnegativity
+    nondim_ε = sum(c_rf_opt[1,1:m].*f_entr)/m
+    nondim_δ = sum(c_rf_opt[2,1:m].*f_detr)/m
```
charleskawczynski, line 227: We could write this in a non-allocating way, but I think this is fine for now as we're experimenting with NN parameterizations.

charleskawczynski (approved):
LGTM.

I'm not 100% sure, but I think you may need to extend [this](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/driver/main.jl#L257-L260) (if we're going to run calibration on this.). cc @haakon-e for input

nickhnelsen:
> LGTM.
> 
> I'm not 100% sure, but I think you may need to extend [this](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/driver/main.jl#L257-L260) (if we're going to run calibration on this.). cc @haakon-e for input
> 
> Also, let's squash / rebase before merging.

Thanks for catching that, I'll defer to @haakon-e (in particular, would I only include the parameters here that need to be calibrated?).

Also, I have never squashed/rebased before so any advice on how this is usually done would be appreciated!

haakon-e:
> LGTM.
> 
> I'm not 100% sure, but I think you may need to extend [this](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/driver/main.jl#L257-L260) (if we're going to run calibration on this.). cc @haakon-e for input
> 
> Also, let's squash / rebase before merging.

yes, that needs to be extended.

I'm not super happy about creating new sets of parameter names for every closure we experiment with (that's why we created `c_gen`), but others (@yairchn, @costachris) may have had more thorough discussions on this lately to conclude that we can keep expanding the namelist file.

driver/generate_namelist.jl
```diff
@@ -110,6 +110,24 @@ function default_namelist(case_name::String; root::String = ".", write::Bool = t
...
+                    ones(2,1)/1e2,
+                    ones(2,2), dims=2))
+        )
+
+    # RF: fixed realizations of random variables, 2 x m x (1 + d)
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["rf_fix_ent_params"] =
+        SA.SVector{1000}(
+        vec(cat(2*pi*rand(2,100,1),
+                    randn(2,100,4), dims=3))
+        )
```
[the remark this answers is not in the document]
charleskawczynski: +1 for fixing the seed!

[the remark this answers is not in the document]
charleskawczynski: I believe that must be done at the top level, though, so we'll probably want to add that to `driver/main.jl`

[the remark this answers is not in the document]
haakon-e: Could we add a random seed as an optional entry to the `namelist`? This way, you can choose whether to fix or not fix the seed e.g. for every EKI iteration.

costachris:
> > LGTM.
> > I'm not 100% sure, but I think you may need to extend [this](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/driver/main.jl#L257-L260) (if we're going to run calibration on this.). cc @haakon-e for input
> > Also, let's squash / rebase before merging.
> 
> yes, that needs to be extended.
> 
> I'm not super happy about creating new sets of parameter names for every closure we experiment with (that's why we created `c_gen`), but others (@yairchn, @costachris) may have had more thorough discussions on this lately to conclude that we can keep expanding the namelist file.

We discussed it but never fully settled on a solution. I agree it's not ideal to have a different namelist vector for every closure type. It seems okay for now since there's only a few closure types and this makes it easy to switch between them, although one solution is to nest vectors under `c_gen` so we only have to pass around `c_gen` through the source code (ie.  `namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["c_gen"]["rf_fix_ent_params"]`... and similarly for others

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -193,6 +193,41 @@ function non_dimensional_function(param_set, εδ_model_vars, ::LinearEntr)
...
+    non_dimensional_function(param_set, εδ_model_vars, ::RFEntr)
+
+Uses a Random Feature model to predict the non-dimensional components of dynamical entrainment/detrainment.
+ - `param_set`      :: parameter set
+ - `εδ_model_vars`  :: structure containing variables
+ - `εδ_model_type`  :: RFEntr - basic RF entrainment closure
+"""
+function non_dimensional_function(param_set, εδ_model_vars, ::RFEntr)
+    # d=4 inputs, p=2 outputs, m=100 random features
+    d = 4
```
yairchn, line 207: do we wanna make  `d` and `m` flexible (i.e. move them to namelist arguments)?

src/closures/nondimensional_exchange_functions.jl
yairchn: if the non-allocating way simple - maybe it canoe done now and not worry about it in the future ?

yairchn (approved):
I left a few comments but overall looks good to me

src/closures/nondimensional_exchange_functions.jl
charleskawczynski: I think it's probably fine for now while we're experimenting. It's important that people using these untested (via the test suite) experimental features report that they're slow (if they are) though.

driver/generate_namelist.jl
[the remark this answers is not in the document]
nickhnelsen: Thanks, I'll add this

src/closures/nondimensional_exchange_functions.jl
nickhnelsen: Yes, m should be flexible. I'll add this

charleskawczynski:
bors try

yairchn:
> There are some performance issues we should figure out how to manage before merging. I think it may have to due with the matrices in the parameter set.

@charleskawczynski is there something we can do to help them resolve these issue and move forward with the PR?

charleskawczynski:
Can we try using regular vectors instead of SVectors in the namelist and lift the `isbits` limitation? (just add a separate commit and see if CI is faster)

driver/main.jl
```diff
@@ -254,7 +254,13 @@ nc_results_file(::Nothing) = @info "The simulation was run without IO, so no nc
 
 function main1d(namelist; time_run = true)
     # TODO: generalize conversion of arrays from namelist to `SVector`s.
-    for param_name in ["general_ent_params", "general_stochastic_ent_params", "fno_ent_params"]
+    for param_name in [
+        "general_ent_params",
+        "general_stochastic_ent_params",
+        "fno_ent_params",
+        "rf_opt_ent_params",
+        "rf_fix_ent_params",
```
nickhnelsen, line 262: comma should not be on line 262, will remove after performance test without SVec

yairchn:
@charleskawczynski can you take one more look at the buildkite before we merge?

charleskawczynski (approved):

PR #796, opened 2022-02-09
Fix dry/moist thermo state diagnostics

### LINKED ISSUE #795: Incorrect diagnostic of conditional variables in quadratures
The current implementation of the environmental quadratures is used to diagnose dry and moist variables by sampling the dry/cloudy quadrature points. The information is then used in the buoyancy gradient diagnostics, and will be coupled with radiation in the future.

At the moment we were computing conditional averages as

x_sat_contrib = ∫ x(q,θ)*H(ql(q,θ)>0) ds,

which is the contribution of the saturated parcel to the environment (over which we integrate). This equation is the correct quadrature for environmental properties with exclusive contributions from saturated parcels, like the cloud fraction or the liquid water specific humidity. However, if we are interested in the value of the variables within the saturated volume, we need to divide by the cloud fraction,

x_sat = [ ∫ x(q,θ)*H(ql(q,θ)>0) ds ]  / [ ∫ H(ql(q,θ)>0) ds ]

PD: Error found by @trontrytel

ilopezgp:
Closes #795 .

### DIFF
--- a/src/EDMF_Environment.jl
+++ b/src/EDMF_Environment.jl
@@ -70,7 +70,7 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
     quad_order = quadrature_order(en_thermo)
     tendencies_pr = center_tendencies_precipitation(state)
 
-    #TODO - remember you output source terms multipierd by Δt (bec. of instanteneous autoconcv)
+    #TODO - remember you output source terms multiplied by Δt (bec. of instantaneous autoconv)
     #TODO - add tendencies for gm H, QT and QR due to rain
     #TODO - if we start using eos_smpl for the updrafts calculations
     #       we can get rid of the two categories for outer and inner quad. points
@@ -85,7 +85,7 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
     sqpi_inv = 1 / sqrt(π)
     sqrt2 = sqrt(2)
 
-    epsilon = 10e-14 #np.finfo(np.float).eps
+    epsilon = 10e-14 # eps(float)
 
     # initialize the quadrature points and their labels
     inner_env = zeros(env_len)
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
Code changes look fine to me.

src/EDMF_Environment.jl
```diff
@@ -231,21 +231,30 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
...
-            aux_en_sat.T[k] = outer_env[i_T_sat]
-            aux_en_sat.q_vap[k] = outer_env[i_qt_sat] - outer_env[i_ql]
-            aux_en_sat.q_tot[k] = outer_env[i_qt_sat]
-            ts_sat = TD.PhaseEquil_pTq(param_set, p0_c[k], aux_en_sat.T[k], aux_en_sat.q_tot[k])
-            aux_en_sat.θ_dry[k] = TD.dry_pottemp(ts_sat)
-            aux_en_sat.θ_liq_ice[k] = TD.liquid_ice_pottemp(ts_sat)
+            if aux_en.cloud_fraction[k] > 0
+                aux_en_sat.T[k] = outer_env[i_T_sat] / aux_en.cloud_fraction[k]
+                aux_en_sat.q_tot[k] = outer_env[i_qt_sat] / aux_en.cloud_fraction[k]
+                aux_en_sat.q_vap[k] = (outer_env[i_qt_sat] - outer_env[i_ql]) / aux_en.cloud_fraction[k]
```
trontrytel, line 247:
```suggestion
                aux_en_sat.q_vap[k] = (outer_env[i_qt_sat] - outer_env[i_ql] - outer_env[i_qi]) / aux_en.cloud_fraction[k]
```

trontrytel (commented):
Should we also change the weights in lines 190 - 196?

src/EDMF_Environment.jl
trontrytel: Or we could also just use the TD function: https://github.com/CliMA/Thermodynamics.jl/blob/5945e684e36d7759088097d00fe4562ceb9db0cf/src/relations.jl#L210

yairchn (approved):
LGTM

ilopezgp:
> Should we also change the weights in lines 190 - 196?

I prefer not to change the weights, which are the weights for Gauss Hermite quadratures over the environment. When we diagnose conditional variables, we just have to divide by the cloud fraction, as discussed in #795 .

trontrytel:
> > Should we also change the weights in lines 190 - 196?
> 
> I prefer not to change the weights, which are the weights for Gauss Hermite quadratures over the environment. When we diagnose conditional variables, we just have to divide by the cloud fraction, as discussed in #795 .

What I meant is that in those lines we also compute the conditional saturated vs unsaturated statistics:
https://github.com/CliMA/TurbulenceConvection.jl/blob/80947ac1f4c75c98d4f476a4ae1a9c72a68a6641/src/EDMF_Environment.jl#L190-L196

ilopezgp:
> > > Should we also change the weights in lines 190 - 196?
> > 
> > 
> > I prefer not to change the weights, which are the weights for Gauss Hermite quadratures over the environment. When we diagnose conditional variables, we just have to divide by the cloud fraction, as discussed in #795 .
> 
> What I meant is that in those lines we also compute the conditional saturated vs unsaturated statistics:
> 
> https://github.com/CliMA/TurbulenceConvection.jl/blob/80947ac1f4c75c98d4f476a4ae1a9c72a68a6641/src/EDMF_Environment.jl#L190-L196

The division only has to be done once, I have done it after aggregation, which is when we have the cloud fraction available. It could be done here if we looped before just to get the cloud fraction, but this is more efficient.

PR #797, opened 2022-02-09
Let CI pass

charleskawczynski:
Our CI is currently broken because the cluster has been cleaned out, which is fine as the file system was building up quite a bit. So, let's restart things by ensuring that all of our tests pass and we can move the current main results onto the cluster.

**This pr short circuits our regression tests and must be reversed to avoid allowing all prs to automatically pass ci.** No other PRS should be allowed to merge in the mean time.


charleskawczynski:
bors try

PR #799, opened 2022-02-10
Refactor `src/EDMF_Environment.jl`

charleskawczynski:
~In efforts of reducing dependence on `Cent`, we're trying to replace our `real_center_indices`/`real_face_indices` loops with broadcast expressions. One place where we have such a loop is in `src/EDMF_Environment.jl`, which happens to also be a part of the code that we'd like to try improving w.r.t. performance (it costs  about 20-30% of `∑tendencies!`).~

I spoke with @trontrytel, and we decided that renaming some of the other variables would be better in a separate PR since it affects many other places in the code. So this PR now only has the some small non-controversial name changes.

### DIFF
--- a/src/EDMF_Environment.jl
+++ b/src/EDMF_Environment.jl
@@ -112,9 +112,8 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
                 corr = max(min(aux_en.HQTcov[k] / sqrt(aux_en.Hvar[k] * aux_en.QTvar[k]), 1), -1)
                 sd2_hq = log(corr * sqrt(aux_en.Hvar[k] * aux_en.QTvar[k]) / aux_en.θ_liq_ice[k] / aux_en.q_tot[k] + 1)
                 sd_cond_h_q = sqrt(max(sd_h * sd_h - sd2_hq * sd2_hq / sd_q / sd_q, 0))
-                mu_q =
-                    log(aux_en.q_tot[k] * aux_en.q_tot[k] / sqrt(aux_en.q_tot[k] * aux_en.q_tot[k] + aux_en.QTvar[k]))
-                mu_h = log(
+                μ_q = log(aux_en.q_tot[k] * aux_en.q_tot[k] / sqrt(aux_en.q_tot[k] * aux_en.q_tot[k] + aux_en.QTvar[k]))
+                μ_h = log(
                     aux_en.θ_liq_ice[k] * aux_en.θ_liq_ice[k] /
                     sqrt(aux_en.θ_liq_ice[k] * aux_en.θ_liq_ice[k] + aux_en.Hvar[k]),
                 )
@@ -130,7 +129,7 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
                 # TODO - change 1e-13 and 1e-10 to some epislon
                 sd_q = min(sd_q, sd_q_lim)
                 qt_var = sd_q * sd_q
-                sigma_h_star = sqrt(max(1 - corr * corr, 0)) * sd_h
+                σ_h_star = sqrt(max(1 - corr * corr, 0)) * sd_h
             end
 
             # zero outer quadrature points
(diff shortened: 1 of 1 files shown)


src/EDMF_Environment.jl
```diff
@@ -93,44 +93,53 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
     inner_src = zeros(src_len)
     outer_src = zeros(src_len)
     i_ql, i_qi, i_T, i_cf, i_qt_sat, i_qt_unsat, i_T_sat, i_T_unsat = 1:env_len
-    i_SH_qt, i_Sqt_H, i_SH_H, i_Sqt_qt, i_Sqt, i_SH, i_Sqr, i_Sqs = 1:src_len
+    i_SH_qt, i_Sqt_θ, i_SH_θ, i_Sqt_qt, i_Sqt, i_SH, i_Sqr, i_Sqs = 1:src_len
+
```
trontrytel, line 97:
```suggestion
i_Sθ_qt, i_Sqt_θ, i_Sθ_θ, i_Sqt_qt, i_Sqt, i_Sθ, i_Sqr, i_Sqs = 1:src_len
```
We should replace h with θ everywhere.

trontrytel: To be consistent with the rest of the code it should actually be  `θ_liq_ice` and not just `θ`. Same for `qt` it should be `q_tot`. But that would make some of the lines here quite involved with underscores as we go through `inner` and `outer`, `sat` and `unsat` options. So maybe it makes sense to drop `_liq_ice` notation inside the quadratures loop.

src/EDMF_Environment.jl
```diff
@@ -93,44 +93,53 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
...
+    QTvar_en = aux_en.QTvar
+    Hvar_en = aux_en.Hvar
+    HQTcov_en = aux_en.HQTcov
+    q_tot_en = aux_en.q_tot
```
trontrytel, line 101: We could also replace those with 
```
qt_var_en = aux_en.qt_var
θ_var_en = aux_en.θ_var
θ_qt_cov_en = aux_en.θ_qt_cov
```

src/EDMF_Environment.jl
```diff
@@ -93,44 +93,53 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
...
+    θ_liq_ice_en = aux_en.θ_liq_ice
+    area_en = aux_en.area
+    qt_tendency_precip_formation_en = aux_en.qt_tendency_precip_formation
+    θ_liq_ice_tendency_precip_formation_en = aux_en.θ_liq_ice_tendency_precip_formation
+    cloud_fraction_en = aux_en.cloud_fraction
+    Hvar_rain_dt_en = aux_en.Hvar_rain_dt
+    QTvar_rain_dt_en = aux_en.QTvar_rain_dt
+    HQTcov_rain_dt_en = aux_en.HQTcov_rain_dt
```
trontrytel, line 110: θ_var_rain_dt_en = aux_en.θ_var_rain_dt
    qt_var_rain_dt_en = aux_en.qt_var_rain_dt
    θ_qt_cov_rain_dt_en = aux_en.θ_qt_cov_rain_dt

src/EDMF_Environment.jl
```diff
@@ -93,44 +93,53 @@ function microphysics(en_thermo::SGSQuadrature, grid::Grid, state::State, precip
...
-                sd_q = sqrt(aux_en.QTvar[k])
-                sd_h = sqrt(aux_en.Hvar[k])
-                corr = max(min(aux_en.HQTcov[k] / max(sd_h * sd_q, 1e-13), 1), -1)
+                sd_q = sqrt(QTvar_en[k])
+                sd_θ = sqrt(Hvar_en[k])
+                corr = max(min(HQTcov_en[k] / max(sd_θ * sd_q, 1e-13), 1), -1)
 
-                # limit sd_q to prevent negative qt_hat
-                sd_q_lim = (1e-10 - aux_en.q_tot[k]) / (sqrt2 * abscissas[1])
+                # limit sd_q to prevent negative q̂_tot
```
trontrytel, line 135: Not a big fan of the hat symbol here. I can barely see it, tbh.

trontrytel (commented):
Left some comments. Looks good otherwise. After we have the paper notation about the quadratures more settled I'll go over the file again and make sure we match the symbols we use in the paper.

PR #801, opened 2022-02-10
Refactor quadrature

charleskawczynski:
This PR moves the quadrature into a separate function. It's still a WIP, but I got Bomex running and wanted to see how the performance looks before continuing.


src/EDMF_Environment.jl
```diff
@@ -57,43 +57,170 @@ function microphysics(
...
+    inner_src = zeros(src_len)
+    outer_src = zeros(src_len)
+
+    abscissas = a
+    weights = w
+
+    sqpi_inv = 1 / sqrt(π)
+    sqrt2 = sqrt(2)
+
+    if quadrature_type isa LogNormalQuad
```
yairchn, line 85: I think we can write a function `get_stand_div` that dispatches on `LogNormalQuad` or `GaussianQuad` and return the standard devisions here.

src/EDMF_Environment.jl
```diff
@@ -57,43 +57,170 @@ function microphysics(
...
+
+    # zero outer quadrature points
+    @inbounds for idx in 1:env_len
+        outer_env[idx] = 0.0
+    end
+    @inbounds for idx in 1:src_len
+        outer_src[idx] = 0.0
+    end
+
+    @inbounds for m_q in 1:quad_order
```
yairchn, line 118: I think we can write a function `sample_quad_point` that that dispatches on `LogNormalQuad` or `GaussianQuad`,  receives the inputs and return the `qt_hat` and `μ_h_star`

src/EDMF_Environment.jl
```diff
@@ -57,43 +57,170 @@ function microphysics(
...
+        # zero inner quadrature points
+        for idx in 1:env_len
+            inner_env[idx] = 0.0
+        end
+        for idx in 1:src_len
+            inner_src[idx] = 0.0
+        end
+
+        for m_h in 1:quad_order
+            if quadrature_type isa LogNormalQuad
```
yairchn, line 136: same idea here

PR #810, opened 2022-02-14
Add 3d sphere baroclinic wave example

charleskawczynski:
This PR adds a 3d baroclinic wave example on a sphere, and calls TC.jl. Right now, we only run for 2 timesteps, and quite a few hacks are used to allow it to run.

Specifically, right now the state vector is the union of the original example and the one we use in TC.jl, and they live on different spaces. I think the next step is fixing this and verifying that it's doing the correct thing.

Right now, this example includes a bunch of code that is not strictly needed for the driver to run, so some names have been changed to avoid name collisions (and over-writing methods).


charleskawczynski:
This example, which takes only 2 timesteps, takes 2 min 40 seconds and allocates almost 60 GB. Once we see that it's working, we'll probably need to look at performance / memory.

PR #819, opened 2022-02-18
Try MVector in quad_loop

charleskawczynski:


charleskawczynski:
This fixes the allocations in the quad loop 🎉

PR #824, opened 2022-02-21
Yc/add constant diff model

yairchn:
testing our ability to use a different turn_conv model. I still have a few issues left


driver/callbacks.jl
```diff
@@ -37,11 +37,13 @@ function affect_io!(integrator)
...
 
 function affect_filter!(integrator)
-    UnPack.@unpack edmf, grid, param_set, aux, case = integrator.p
+    UnPack.@unpack turb_conv, grid, param_set, aux, case = integrator.p
     t = integrator.t
-    state = TC.State(integrator.u, aux, integrator.du)
+    state = TC.State(integrator.u, aux, ODE.get_du(integrator))
     surf = get_surface(case.surf_params, grid, state, t, param_set)
-    TC.affect_filter!(edmf, grid, state, param_set, surf, case.casename, t)
+    if turb_conv isa TC.EDMFModel
```
yairchn, line 44: @charleskawczynski  is there a better way to do this ?

driver/callbacks.jl
```diff
@@ -100,8 +102,43 @@ function dt_max!(integrator)
     ODE.u_modified!(integrator, false)
 end
 
+function diffusivity_dt_max!(integrator)
```
yairchn, line 105: this requires some planning

driver/dycore.jl
```diff
@@ -197,75 +204,190 @@ function satadjust(param_set::APS, grid, state)
...
-    UnPack.@unpack edmf, precip_model, grid, param_set, case, aux, TS = params
+    UnPack.@unpack turb_conv, precip_model, grid, param_set, case, aux, TS = params
 
     state = TC.State(prog, aux, tendencies)
 
     Δt = TS.dt
     surf = get_surface(case.surf_params, grid, state, t, param_set)
     force = case.Fo
     radiation = case.Rad
+    if turb_conv isa TC.EDMFModel
```
yairchn, line 218: is there a better way to do this ?

driver/dycore.jl
```diff
@@ -197,75 +204,190 @@ function satadjust(param_set::APS, grid, state)
...
+end
+
+function update_aux!(turb_conv::TC.DiffusivityModel, grid, state, surf, param_set, t, Δt)
+    kc_surf = TC.kc_surface(grid)
+    FT = eltype(grid)
+    aux_gm_f = TC.face_aux_grid_mean(state)
+    prog_gm = TC.center_prog_grid_mean(state)
+    wvec = CC.Geometry.WVector
+    ∇ = CCO.DivergenceC2F(; bottom = CCO.SetDivergence(FT(0)), top = CCO.SetDivergence(FT(0)))
+    # ∇u = ∇(wvec(prog_gm.u))
```
yairchn, line 262: I still need to compute the shear here

driver/callbacks.jl
charleskawczynski: Yes, rename the function to `affect_filter!` edmf and only add this to the callback set when we're using EDMF. We probably shouldn't be running through this function at all if we don't have edmf.

charleskawczynski:
@yairchn, okay to close?

yairchn:
yes

PR #830, opened 2022-02-23
Convert Namelist arrays to tuples

charleskawczynski:
Can people test out if this works in general?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3,6 +3,7 @@ if !haskey(ENV, "BUILDKITE")
     Pkg.develop(Pkg.PackageSpec(; path = dirname(@__DIR__)))
 end
 using Test
+using JSON
 import TurbulenceConvection
 const TC = TurbulenceConvection
 const tc_dir = dirname(dirname(pathof(TC)))
@@ -35,6 +36,13 @@ end
         t_stop = 6 * 3600,
     )
     nothing
+
+    # Test running from parsed namelist
+    namelist_filename = "Output.Bomex.01/namelist_Bomex.in"
+    read_namelist = JSON.parsefile(namelist_filename)
+    read_namelist["meta"]["uuid"] = "02"
+    read_namelist["time_stepping"]["t_max"] = read_namelist["time_stepping"]["dt_max"]
+    main(read_namelist)
 end
 
 @testset "ClimaCore extensions" begin
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -284,19 +284,15 @@ main(namelist; kwargs...) = @timev main1d(namelist; kwargs...)
 nc_results_file(stats::NetCDFIO_Stats) = stats.path_plus_file
 nc_results_file(::Nothing) = @info "The simulation was run without IO, so no nc files were exported"
 
+to_svec(x::AbstractArray) = SA.SVector{length(x)}(x)
+to_svec(x::Tuple) = SA.SVector{length(x)}(x)
+
 function main1d(namelist; time_run = true)
-    # TODO: generalize conversion of arrays from namelist to `SVector`s.
-    for param_name in [
-        "general_ent_params",
-        "general_stochastic_ent_params",
-        "fno_ent_params",
-        "rf_opt_ent_params",
-        "rf_fix_ent_params",
-        "pi_norm_consts",
-    ]
-        if haskey(namelist["turbulence"]["EDMF_PrognosticTKE"], param_name)
-            _p = namelist["turbulence"]["EDMF_PrognosticTKE"][param_name]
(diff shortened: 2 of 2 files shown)


haakon-e:
Why are we changing this to tuples?

charleskawczynski:
> Why are we changing this to tuples?

It doesn't really matter if they're `SVector`s or tuples, what I'd like to ensure is that they're immutable / `isbits`

haakon-e:
> > Why are we changing this to tuples?
> 
> It shouldn't really matter if they're `SVector`s or tuples, what I'd like to ensure is that they're immutable / `isbits`

so, from a performance point of view, it should be the same? -- even for large vectors

charleskawczynski:
> so, from a performance point of view, it should be the same? -- even for large vectors

`SVector`s are backed by tuples, so performance should be the same.

costachris:
Converting everything to `ntuple` leads to some typing issues for some of the closure functions. ie. here
https://github.com/CliMA/TurbulenceConvection.jl/blob/084de533b24ecb89f5c3f936ed2699bc3f386881/src/closures/nondimensional_exchange_functions.jl#L137

since
```
typeof(StaticArrays.SVector(1,2,3,4,5,6)) <: AbstractArray{FT} where {FT <: Real} = true
typeof(ntuple(i -> i, 6)) <: AbstractArray{FT} where {FT <: Real} = false
```

So I can change the types of those to be consistent with `ntuples` instead of SVectors if we're happy with this

ilopezgp:
Should we then go with the homogeneous conversion to `SVector`s?

charleskawczynski:
Ok, I've changed this to convert to `SVectors`. Maybe someone can re-test? Or, better yet, perhaps we could add a test in TC to ensure that this will work upstream?

@ilopezgp, would it be difficult to add a test to `test/runtests.jl` with this?

ilopezgp:
> Ok, I've changed this to convert to `SVectors`. Maybe someone can re-test? Or, better yet, perhaps we could add a test in TC to ensure that this will work upstream?
> 
> @ilopezgp, would it be difficult to add a test to `test/runtests.jl` with this?

It should be easy. We just need to write to file the namelist, read it back, and run any integration test.

ilopezgp:
> > Ok, I've changed this to convert to `SVectors`. Maybe someone can re-test? Or, better yet, perhaps we could add a test in TC to ensure that this will work upstream?
> > @ilopezgp, would it be difficult to add a test to `test/runtests.jl` with this?
> 
> It should be easy. We just need to write to file the namelist, read it back, and run an integration test with that namelist.

Want me to do it or are you on it?

charleskawczynski:
> Want me to do it or are you on it?

That'd be great! Feel free to push to this branch

costachris:
With the most recent commit `integration_test/Bomex.jl` runs as expected with ("moisture_deficit", "NN", "NN_nonlocal", "FNO", "RF") closures. CEDMF is also running with both the default and NN closure.

test/runtests.jl
```diff
@@ -35,6 +36,12 @@ end
         t_stop = 6 * 3600,
     )
     nothing
+
+    # Test running from parsed namelist
+    namelist_path = "Output.Bomex.01/namelist_Bomex.in"
+    read_namelist = JSON.parsefile(namelist_path)
+    read_namelist["meta"]["uuid"] = "02"
+    main(read_namelist)
```
charleskawczynski, line 44:
```suggestion
    # Perform single step
    read_namelist["time_stepping"]["t_max"] = read_namelist["time_stepping"]["dt_max"]
    sim = Simulation1d(read_namelist)
    initialize(sim)
    (prob, alg, kwargs) = solve_args(sim)
    integrator = ODE.init(prob, alg; kwargs...)
    ODE.step!(integrator) # perform single step
```
Since this is done in serial, maybe we can just do a single timestep instead. We might need to add `import OrdinaryDiffEq; const ODE = OrdinaryDiffEq` above, too.

test/runtests.jl
```diff
@@ -35,6 +36,12 @@ end
         t_stop = 6 * 3600,
     )
     nothing
+
+    # Test running from parsed namelist
+    namelist_path = "Output.Bomex.01/namelist_Bomex.in"
+    read_namelist = JSON.parsefile(namelist_path)
```
charleskawczynski, line 42:
```suggestion
    namelist_filename = "Output.Bomex.01/namelist_Bomex.in"
    read_namelist = JSON.parsefile(namelist_filename)
```
I think `path` usually refers to folders (e.g., `mkpath`, `ispath`)

test/runtests.jl
ilopezgp: I think we can directly call main after modifying get the `t_max`, that's what we do in CEDMF for creating the references. I'll try.

ilopezgp (approved):

PR #835, opened 2022-02-25
Ensure NC files always close

### LINKED ISSUE #55: Improve IO management
Sometimes, when running interactively, simulations crash and re-trying the simulation gives an NC permissions error. #54 is a bit of a patch fix, and we can't [easily](https://github.com/Alexander-Barth/NCDatasets.jl/issues/135) open-close files on-the-fly. We could alternatively wrap the whole solve loop in a `do`-block that opens the nc file, but this may take a bit of work.

charleskawczynski:
I think this is a more careful way to handle our nc file open/closing. This should ensure that our nc files always close-- which I've been thinking about since looking more at the calibration pipeline / failures.

IIRC, I took this pattern from NCDatasets (looking at how the `do`-block syntax works).

Closes #55

cc @haakon-e

### DIFF
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -265,13 +265,21 @@ function run(sim::Simulation1d; time_run = true)
     sim.skip_io || open_files(sim.Stats) # #removeVarsHack
     (prob, alg, kwargs) = solve_args(sim)
     integrator = ODE.init(prob, alg; kwargs...)
-    if time_run
-        sol = @timev ODE.solve!(integrator)
-    else
-        sol = ODE.solve!(integrator)
+
+    local sol
+    try
+        if time_run
+            sol = @timev ODE.solve!(integrator)
+        else
+            sol = ODE.solve!(integrator)
+        end
+    catch e
+        @warn "TurbulenceConvection simulation crashed."
+        return :simulation_crashed
+    finally
+        sim.skip_io || close_files(sim.Stats) # #removeVarsHack
     end
 
-    sim.skip_io || close_files(sim.Stats) # #removeVarsHack
     if first(sol.t) == sim.TS.t_max
         return :success
     else


haakon-e (approved):
LGTM

driver/main.jl
```diff
@@ -265,13 +265,20 @@ function run(sim::Simulation1d; time_run = true)
...
+
+    local sol
+    try
+        if time_run
+            sol = @timev ODE.solve!(integrator)
+        else
+            sol = ODE.solve!(integrator)
+        end
+    catch e
+        return :simulation_crashed
```
ilopezgp, line 277: Is the `finally` reached if you return here?

jakebolewski: finally will be called before return.  it would be more natural to throw an exception though (and wrap the original exception)?

charleskawczynski: Yep:
```julia
julia> function foo()
           try
               error("oops")
               return 2
           catch
               return 1
           finally
               println("finally")
           end
       end
foo (generic function with 1 method)

julia> foo()
finally
1
```

charleskawczynski: > it would be more natural to throw an exception though (and wrap the original exception)?

Maybe you're right, but one advantage I've realized about working with the return code is that, instead of error-ing out, we can continue to the plotting scripts and (try) to plot results of the crashed solution and test which variables are blowing up.

jakebolewski: you don't have to error out though, you could capture the exception at a higher level (log it?) and then continue with the control flow.

charleskawczynski: Ahhh, I see, yeah, we can do that

charleskawczynski: I was thinking about this a bit more, if we returned an exception, wouldn't that make `main` type-unstable? I think, for now, just adding a warning to the log and returning a symbol is a significant memory management improvement that we should get merged, so let's move forward and we can open an issue about better exception handling.

PR #858, opened 2022-03-04
Expand tests

charleskawczynski:
This PR expands our test suite to include a "Vanilla + deviations" group, which basically all run bomex with different model schemes and namelist flags:

```
--micro gaussian
--entr NN_nonlocal
--entr NN
--entr FNO
--entr RF
--stoch_entr noisy_relaxation_process
--stoch_entr prognostic_noisy_relaxation_process
--calibrate_io true
--skip_io true
```

It's really helpful that we are able to scale up the number of tests because all sorts of things can diverge as different users / developers are experimenting with different models-- some functions break, performance tanks for certain cases. Since most of these models are being developed for upstream use in CalibrateEDMF, it's useful to run tests here to avoid laborious investigations into where things break / diverge. It's also helpful for us to easily identify where performance improvement efforts may be useful.

One downside is that it is making our driver a bit more complicated. Improving our regression tests has been a bucket list item for a while and I suspect it may be time to give it some attention, this PR illuminates some of the needs / requirements. I'll open an issue to expand on this.

Please comment if there are other flags / tests that we should add into this.


ilopezgp:
I am surprised all tests pass with the entrainment closure default parameters !

ilopezgp:
For `calibrate_io`, we want to test not only that the simulation does not break, but that the fields in the calibrate_dictionary are updated.

charleskawczynski:
> I am surprised all tests pass with the entrainment closure default parameters !

We're skipping lots of tests at the end, but it's nice that everything runs and we're able to plot the results. I'd like to

> For `calibrate_io`, we want to test not only that the simulation does not break, but that the fields in the calibrate_dictionary are updated.

Yeah, I was thinking that it'd be nice to ensure that the state is exactly the same. I think this will require more work, though.

Also, I just realized that I need to add the stochastic entrainment model

costachris:
> > I am surprised all tests pass with the entrainment closure default parameters !
> 
> We're skipping lots of tests at the end, but it's nice that everything runs and we're able to plot the results. I'd like to
> 
> > For `calibrate_io`, we want to test not only that the simulation does not break, but that the fields in the calibrate_dictionary are updated.
> 
> Yeah, I was thinking that it'd be nice to ensure that the state is exactly the same. I think this will require more work, though.
> 
> Also, I just realized that I need to add the stochastic entrainment model. Done.

What tests are skipped?

costachris (approved):

charleskawczynski:
> What tests are skipped?

If you look at the jobs, you'll see that a handful of added runs result in errors. We still see plots because we use a `try-catch` around `main`. We also bail out (if we deviate from the vanilla experiments) before we run any tests, including
 - Regression against scampy/les/main
 - Test that `code_ret == :success` (meaning that `OrdinaryDiffEq.solve!` hasn't bailed out due to detected numerical instability)
 - Test that 
```julia
        @test !any(isnan_or_inf.(Array(profile["qt_mean"])))
        @test !any(isnan_or_inf.(Array(profile["updraft_area"])))
        @test !any(isnan_or_inf.(Array(profile["updraft_w"])))
        @test !any(isnan_or_inf.(Array(profile["updraft_qt"])))
        @test !any(isnan_or_inf.(Array(profile["updraft_thetal"])))
        @test !any(isnan_or_inf.(Array(profile["u_mean"])))
        @test !any(isnan_or_inf.(Array(profile["tke_mean"])))
        @test !any(isnan_or_inf.(Array(profile["temperature_mean"])))
        @test !any(isnan_or_inf.(Array(profile["ql_mean"])))
        @test !any(isnan_or_inf.(Array(profile["qi_mean"])))
        @test !any(isnan_or_inf.(Array(profile["thetal_mean"])))
        @test !any(isnan_or_inf.(Array(profile["Hvar_mean"])))
        @test !any(isnan_or_inf.(Array(profile["QTvar_mean"])))
        @test !any(isnan_or_inf.(Array(profile["v_mean"])))
        @test !any(isnan_or_inf.(Array(profile["qr_mean"])))
```
These are basically all the tests that we run on the vanilla cases.

The first one is difficult for the Vanilla + deviation-- because we aren't (yet) storing results in main. The latter two are likely sooner in reach, but we need good values (cc @nickhnelsen, @zongyi-li) for the cases that are breaking, namely:
 - Random features
 - FNO
 - Local Neural network actually finishes, but it looks a bit wonky
 - Local Neural network actually finishes, but it looks a bit wonky

Also, @trontrytel, the Bomex with `gaussian` looks like it's identical, do we have to run a rain case to make sure we expand coverage?

costachris:
I'll open a separate PR using the parameters calibrated for the SCT1 benchmark that will give more reasonable results for the NN closures.

charleskawczynski:
> got it, I'll open a separate PR using the parameters calibrated for the SCT1 benchmark that will give more reasonable results for the NN closures.

Thanks!

charleskawczynski:
I'm realizing that it's a bit difficult to generalize this for cases that are failing. So, let's merge this and try to get the other cases working / running to completion first, this will at least test that the namelist can be read.

PR #861, opened 2022-03-04
use NN parameters calibrated for SCT1 Benchmark

costachris:


charleskawczynski (commented):
I'll rebase+merge this after #858 merges!

charleskawczynski:
Hmm, I may have messed up the rebase? It looks like the vector is 70 elements and we need 69.

costachris:
> Hmm, I may have messed up the rebase? It looks like the vector is 70 elements and we need 69.

Ah no, that was my mistake. I had the `turbulent_entrainment_factor` in there. TC guards against mistakes like this by checking the length of `general_ent_params` so this will error - but it's fixed now.

costachris:
It's also important to note this was trained with `"area_limiter_power" = 0.0` and `"entr_dim_scale" = "inv_z"`(like all the other new closures), so the comparison plots still look funky. If we want another version that's trained with the default area limiter and entr_dim_scale, I can use that instead. Otherwise I can merge this.

ilopezgp:
> It's also important to note this was trained with `"area_limiter_power" = 0.0` and `"entr_dim_scale" = "inv_z"`(like all the other new closures), so the comparison plots still look funky. If we want another version that's trained with the default area limiter and entr_dim_scale, I can use that instead. Otherwise I can merge this.

Can we add these changes to the vanilla + deviations setup? If it is difficult to do, it might not be worth it.

charleskawczynski:
Yeah, we can do that. For now we can just add it, but I think it's possible to somehow generalize-- I'm thinking about this. I'll update the PR, thanks for the update!

costachris:
> > It's also important to note this was trained with `"area_limiter_power" = 0.0` and `"entr_dim_scale" = "inv_z"`(like all the other new closures), so the comparison plots still look funky. If we want another version that's trained with the default area limiter and entr_dim_scale, I can use that instead. Otherwise I can merge this.
> 
> Can we add these changes to the vanilla + deviations setup? If it is difficult to do, it might not be worth it.

Is there a vanilla + deviations setup? Or are you suggesting adding one.

charleskawczynski:
> Is there a vanilla + deviations setup? Or are you suggesting we add one

Yeah, it's already been merged, we just need to somehow add more cli options to pass these additional flags to update the namelist

charleskawczynski:
Ok, I've rebased (so you'll need to delete local and re-pull/checkout) because there were some fixes that we needed to incorporate. Also, I've added cli options for both `area_limiter_power` and  `entr_dim_scale`. I tried running this locally and things looked a bit wonky. Maybe I messed something up (although, the rebase had no conflicts), the cli options were added for a separate commit for review.

PR #862, opened 2022-03-04
print full stacktrace for failed sim

haakon-e:


charleskawczynski (approved):
This looks great! I'll try to add a test with #858.

PR #866, opened 2022-03-05
Fix scope of calibrate_io, add more fields

charleskawczynski:
@costachris pointed out that we actually need some computations in `compute_diagnostics!`. Arguably, the variables in those aux fields should be moved into a diagnostics array instead, so that we can make better guarantees about not having simulation results depend on diagnostics. For now, I think this is a quick fix to enable using `calibrate_io`.

### DIFF
--- a/driver/callbacks.jl
+++ b/driver/callbacks.jl
@@ -20,9 +20,7 @@ function affect_io!(integrator)
     state = TC.State(integrator.u, aux, ODE.get_du(integrator))
 
     # TODO: is this the best location to call diagnostics?
-    if !calibrate_io
-        compute_diagnostics!(edmf, precip_model, param_set, grid, state, diagnostics, Stats, case, t)
-    end
+    compute_diagnostics!(edmf, precip_model, param_set, grid, state, diagnostics, Stats, case, t, calibrate_io)
 
     # TODO: remove `vars` hack that avoids
     # https://github.com/Alexander-Barth/NCDatasets.jl/issues/135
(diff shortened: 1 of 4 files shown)


ilopezgp (commented):
I have a few comments about calibrate_io,

- Right now we avoid writing to file the `io_dictionary_ref_state()`, but isn't this only written to file at initialization? I don't think it hits the performance that much, and we may need density to do output transformation between specific and total variables.

- Thinking ahead, I think we want to add the following to the calibration diagnostics for microphysics: `"qr_mean", "qs_mean", "cloud_fraction", "RH_mean", "qi_mean", "temperature_mean"`

charleskawczynski:
> I have a few comments about calibrate_io,
> 
> * Right now we avoid writing to file the `io_dictionary_ref_state()`, but isn't this only written to file at initialization? I don't think it hits the performance that much, and we may need density to do output transformation between specific and total variables.
> * Thinking ahead, I think we want to add the following to the calibration diagnostics for microphysics: `"qr_mean", "qs_mean", "cloud_fraction", "RH_mean", "qi_mean", "temperature_mean"`

Fine by me, we can add in these other pieces

charleskawczynski:
We could even, down the road if we wanted to, pass in a dict of which parameters to include / exclude in `io_dictionary_aux_calibrate()`.

ilopezgp:
> We could even, down the road if we wanted to, pass in a dict of which parameters to include / exclude in `io_dictionary_aux_calibrate()`.

Yes, I thought about that. But for now having the dict you created with the proposed additions should do the trick. We want to avoid writing all the environment-specific diagnostics which are not really observable from LES.

charleskawczynski:
Updated.

ilopezgp (approved):
LGTM

PR #869, opened 2022-03-05
bump TC version in integration_tests env

costachris:
Bump TC version in integration_tests env


ilopezgp (approved):

charleskawczynski (changes requested):
I don't think this is an issue since there's a `path` entry-- we should use `] up` when we want to update packages

PR #883, opened 2022-03-10
Add flame graph script for io

charleskawczynski:
This PR adds a flame graph script to analyze io performance. Here's what we're starting how it currently looks:

Screen Shot 2022-03-10 at 9 16 26 AM

@Alexander-Barth, why is `size(::NCDataset)` so expensive? This is happening in the call [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/0b7af151696b717bd5c82c81b68a02400704e3d1/driver/NetCDFIO.jl#L149)

### DIFF
--- a/perf/flame_io.jl
+++ b/perf/flame_io.jl
@@ -1,0 +1,34 @@
+include(joinpath(@__DIR__, "common.jl"))
+import Profile
+
+case_name = "Bomex"
+namelist = NameList.default_namelist(case_name)
+namelist["time_stepping"]["t_max"] = 15 * namelist["time_stepping"]["dt_max"]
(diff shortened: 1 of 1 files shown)


Alexander-Barth:
> why is size(::NCDataset) so expensive?
I don't know, but as you see it is all spent in the C library. `size` is just wrapping `nc_inq_dimlen`.

charleskawczynski:
Yeah, I see. I'm not too familiar with the C library. Maybe there's no way around this. It seems like something that we could track (the size of the unlimited dimension) from our side, but I've been trying to reduce the amount of book keeping. Thanks for taking a look!

PR #885, opened 2022-03-11
Refactor io a bit

charleskawczynski:
This PR:
 - Renames `path_plus_file` to `nc_filename`
 - Makes the inner constructor for `NetCDFIO_Stats` an outer constructor, because I'd like to decouple this data structure from the namelist so that we can later improve io management
 - Deletes some `NetCDFIO_Stats` properties that seem to be unused:
   - `last_output_time`
   - `uuid`
   - `stats_path`


charleskawczynski:
closing in favor of #887.

PR #890, opened 2022-03-14
Add single value per column fields

charleskawczynski:
This PR adds single value per column fields to the `diagnostics` `FieldVector`.


charleskawczynski:
bors try

ilopezgp (approved):

PR #891, opened 2022-03-15
Further refactor IO

charleskawczynski:
This PR makes some of the IO functions, e.g., `write_field` and `add_write_field` etc. accept `group`, instead of a `NetCDFIO_Stats`, and the group is now extracted outside. This makes these io function more low-level and alleviates our dependence on `NetCDFIO_Stats`. I think that we can apply this to other functions, but I'd like to do this incrementally.

### DIFF
--- a/driver/NetCDFIO.jl
+++ b/driver/NetCDFIO.jl
@@ -107,7 +107,7 @@ end
 
 function add_field(ds, var_name::String, dims, group)
     profile_grp = ds.group[group]
-    new_var = NC.defVar(profile_grp, var_name, Float64, dims)
+    NC.defVar(profile_grp, var_name, Float64, dims)
     return nothing
 end
 
@@ -117,7 +117,7 @@ end
 
 function add_ts(ds, var_name::String)
     ts_grp = ds.group["timeseries"]
-    new_var = NC.defVar(ts_grp, var_name, Float64, ("t",))
+    NC.defVar(ts_grp, var_name, Float64, ("t",))
     return nothing
 end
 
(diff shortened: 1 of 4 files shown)


driver/NetCDFIO.jl
```diff
@@ -117,45 +117,34 @@ end
...
 #####
 
-function write_field(self::NetCDFIO_Stats, var_name::String, data::T, group) where {T <: AbstractArray{Float64, 1}}
-    # Hack to avoid https://github.com/Alexander-Barth/NCDatasets.jl/issues/135
-    @inbounds self.vars[group][var_name][:, end] = data
-    # Ideally, we remove self.vars and use:
-    # var = self.profiles_grp[var_name]
+function write_field(group, var_name::String, data::T) where {T <: AbstractArray{Float64, 1}}
+    # @inbounds self.vars[group][var_name][:, end] = data
+    @inbounds group[var_name][:, end] = data
```
ilopezgp, line 130: Does this work with NCDatasets versions < 0.12?

driver/NetCDFIO.jl
```diff
@@ -117,45 +117,34 @@ end
...
-    # # Write to profiles group
-    profile_t = self.profiles_grp["t"]
-    @inbounds profile_t[end + 1] = t::Float64
-
-    # # Write to timeseries group
-    ts_t = self.ts_grp["t"]
-    @inbounds ts_t[end + 1] = t::Float64
+function write_simulation_time(group, t::Float64)
+    # Write to profiles group
+    @inbounds group[end + 1] = t::Float64
```
ilopezgp, line 149: Shouldn't you specify group["t"] here? I see that you are then passing `Stats.profiles_grp["t"]` directly, but this is less clear than indexing ["t"] within the function and just passing the group.

My understanding is that when we use `write_simulation_time`, we move the `end` to the right in all fields that use the time dimension, and that's why we use `end` when updating fields.

driver/callbacks.jl
```diff
@@ -26,7 +26,8 @@ function affect_io!(integrator)
     # https://github.com/Alexander-Barth/NCDatasets.jl/issues/135
     # opening/closing files every step should be okay. #removeVarsHack
     # TurbulenceConvection.io(sim) # #removeVarsHack
-    write_simulation_time(Stats, t) # #removeVarsHack
+    write_simulation_time(Stats.profiles_grp["t"], t)
+    write_simulation_time(Stats.ts_grp["t"], t)
```
ilopezgp, line 30: here it would be better to pass `Stats.profiles_grp`, `Stats.ts_grp` as the first arg.

charleskawczynski: Yeah, I think we can pass that and `var_name` (`"t"`)

ilopezgp:
bors try

PR #892, opened 2022-03-17
notation cleanup in quadratures

trontrytel:
I wanted to make the quadratures notation closer to the paper notation. I did not change any variable names outside of the quadrature loop.

Let me know what you think!


src/EDMF_Environment.jl
```diff
@@ -58,62 +58,61 @@ function microphysics(
...
+    eps_θ = eps(FT) * θi_mean
 
     if quadrature_type isa LogNormalQuad
-        # Lognormal parameters (mu, sd) from mean and variance
-        mu_q = log(q_tot_en * q_tot_en / max(sqrt(q_tot_en * q_tot_en + QTvar_en), eps_q))
-        mu_h = log(θ_liq_ice_en * θ_liq_ice_en / sqrt(θ_liq_ice_en * θ_liq_ice_en + Hvar_en))
-        sd_q = sqrt(log(QTvar_en / max(q_tot_en, eps_q) / max(q_tot_en, eps_q) + 1))
-        sd_h = sqrt(log(Hvar_en / θ_liq_ice_en / θ_liq_ice_en + 1))
+        # Lognormal parameters (ν, s) from mean and variance
+        ν_q = log(qt_mean^2 / max(sqrt(qt_mean^2 + qt′qt′), eps_q))
```
charleskawczynski, line 91:
```suggestion
        μ_q = log(qt_mean^2 / max(sqrt(qt_mean^2 + qt′qt′), eps_q))
```
? and alike below

trontrytel: In the paper I wanted to distinguish between the different means. So I have the:
- environmental mean and (co)variance denoted by bar and primes
- Gaussian mean and standard deviation (\mu and \sigma) 
- lognormal mean and standard deviation (\nu and s). 

It helps me to reason around the code and equations. Plus it ensures that we are not using the Gaussian mean in the lognormal quadrature loop and vice-versa.

charleskawczynski: Ah, ok, sounds good!

charleskawczynski (approved):

trontrytel:
bors try

src/EDMF_Environment.jl
```diff
@@ -58,62 +58,60 @@ function microphysics(
...
+    i_SH_qt, i_Sqt_H, i_SH_H, i_Sqt_qt, i_Sqt, i_SH, i_Sqr, i_Sqs = 1:src_len
+
     quadrature_type = en_thermo.quadrature_type
     quad_order = quadrature_order(en_thermo)
+    χ = en_thermo.a
+    weights = en_thermo.w
 
-    UnPack.@unpack QTvar_en, q_tot_en, Hvar_en, θ_liq_ice_en, HQTcov_en, area_en, q_rai, q_sno, area_en, ρ0_c, p0_c =
-        vars
+    UnPack.@unpack qt′qt′, qt_mean, θi′θi′, θi_mean, θi′qt′, area_en, q_rai, q_sno, ρ0_c, p0_c = vars
```
ilopezgp, line 72: What does the `I` stand for in `θi`?

ilopezgp: I would also keep `en` somehow, e.g., `en_qt′qt′`

src/EDMF_Environment.jl
```diff
@@ -58,62 +58,60 @@ function microphysics(
...
-    weights = w
-
     sqpi_inv = 1 / sqrt(π)
     sqrt2 = sqrt(2)
 
     # Epsilon defined per typical variable fluctuation
-    eps_q = q_tot_en ≈ FT(0) ? eps(FT) : eps(FT) * q_tot_en
-    eps_θ = eps(FT)
+    eps_q = qt_mean ≈ FT(0) ? eps(FT) : eps(FT) * qt_mean
+    eps_θ = eps(FT) #* θi_mean - TODO - testing
```
ilopezgp, line 86: This one should not be multiplied by θ. The product is against the order of fluctuations, which for temperature are O(1).

trontrytel: Then I think I would make both `eps_q` and `eps_theta` equal to eps(FT). `qt` is smaller than 1 having an `eps` smaller than your numerics epsilon doesn't make sense either?

src/EDMF_Environment.jl
trontrytel: Re `I`: I wanted to make the variable names shorter but still have two letters for easier searching. So `qt` for total water and `θ_i` for theta liquid ice. Again the changes are just for the inside quadrature loop, not suggesting to change the variable names outside of it.

Re `en`: Then for consistency we would need `en_qt′qt′` and `en_θi_mean` which makes them long again.

trontrytel:
Aside from the bikeshedding discussions. @charleskawczynski - is it normal for the vanilla tests to randomly require different epsilons? I'm seeing that Rico and TRMM have slightly different mse errors. (The simulations look the same).

charleskawczynski:
> Aside from the bikeshedding discussions. @charleskawczynski - is it normal for the vanilla tests to randomly require different epsilons? I'm seeing that Rico and TRMM have slightly different mse errors. (The simulations look the same).

Yes, our regression tests are basically sensitive to machine precision. Changing the order of operations can result in MSE changes. We just need to update them when this happens, that's all

src/EDMF_Environment.jl
ilopezgp: I prefer here long and understandable... the quadratures are not trivial (we have three variable levels: GMV, env-upd and dry/moist) and this could lead to error in interpretation. I personally find `θi` very misleading as θ_li. I thought it was the potential temperature indexed on a subdomain.

src/EDMF_Environment.jl
ilopezgp: I think the numerics factor out a scale or offset when performing arithmetic operations. Therefore, what's important is to not have terms outside the range scale*(eps(FT), 1/eps(FT)) in arithmetic operations.

For example, 1e-22/1e-10 is still accurate within machine precision, and its result can be used in other operations. `eps_q` appears in equations of the order of `qt`, so we should scale it by this magnitude.

src/EDMF_Environment.jl
yairchn: I support the longer names, but if we are taking shorter names we should make them as standard as possible. In that sense `θl` is better than `θi`.
About the `en`, `up` or `gm`: is it correct that `quad_loop` is a general algorithm for preforming Quadrature on whatever domain / subdomain? In that case there is no need to state the domain / subdomain it is what the user inputs. If we had covariances in the updrafts we could have call quadrature on the updrafts and if we had a reason to do so we could have call quadrature on the grid mean as well, right?

yairchn: @trontrytel I think for `θi_mean` you could use `θ̄i` unless it is not a unicode feature we would like to use in clima

ilopezgp: I would be fine with `θl`, as Yair suggests. I also agree with Yair's point that quadratures are general. Maybe now it's a good time to make an abstract function and just call it here with the environmental components?

trontrytel: Thank you for all the bikeshedding input :)

1) I will keep the `_mean` in the variable names. I can barely see the bar over theta both in terminal and in the web browser.
2) I'll use theta_l and q_t inside the quadrature loop and leave a comment there about what they are. We might have to change the notation anyway, once we switch to energy as state variable.
3) I agree  that the `quad_loop` should in principle work on any sub-domain. But I will keep the implementation as it is for now. I think it's good to add new features when we actually plan to use them and have a meaningful way to test them.

trontrytel:
bors try

PR #899, opened 2022-03-25
separate weights and biases

costachris:
Separates weight and bias parameter vectors for easy zeroing of biases and makes neural network architecture a namelist argument.

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -96,10 +96,16 @@ function default_namelist(
     # entrainment
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment_factor"] = 0.13
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["detrainment_factor"] = 0.51
+
     # nn parameters
     #! format: off
-    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_ent_params"] =
-        SA.SVector{69}(rand(69))
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_arc"] = (6, 5, 4, 2) # [#inputs, #neurons in L1, #neurons in L2, ...., #outputs]
+
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_ent_params"] =
+        SA.SVector{58}(rand(58))
+
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_ent_biases"] = false
+
     # For FNO add here
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["fno_ent_params"] =
         SA.SVector{74}(rand(74))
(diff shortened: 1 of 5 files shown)


driver/generate_namelist.jl
```diff
@@ -96,10 +96,16 @@ function default_namelist(
...
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["detrainment_factor"] = 0.51
+
     # nn parameters
     #! format: off
-    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_ent_params"] =
-        SA.SVector{69}(rand(69))
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_arc"] = (6, 5, 4, 2) # [#inputs, #neurons in L1, #neurons in L2, ...., #outputs]
+
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_ent_weights"] =
+        SA.SVector{58}(rand(58))
```
ilopezgp, line 105: Actually, can we build these through `num_weights_from_arc` and `num_biases_from_arc`? We could have convenience functions `initialize_weights(nn_arc)` and `initialize_biases(nn_arc)`, setting a kwarg for the constructor type (e.g., "uniform" for `rand()`)

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -114,61 +114,84 @@ function non_dimensional_function!(
...
+    weight_param_i = 1
+    bias_param_i = 1
+    # unpack weights and biases into network
     for layer_i in 1:(length(nn_arc) - 1)
         if layer_i == length(nn_arc) - 1
             activation_function = output_layer_activation_function
         end
         layer_num_weights = nn_arc[layer_i] * nn_arc[layer_i + 1]
+        layer_num_biases = nn_arc[layer_i + 1]
         layer = Flux.Dense(
```
ilopezgp, line 177: From the documentation, there is a Boolean kwarg in the Flux.Dense constructor called `bias` (https://fluxml.ai/Flux.jl/stable/models/layers/). Can we just set that to true or false from the namelist? This constructor seems fairly complex.

driver/generate_namelist.jl
costachris: Do you mean calling `num_weights_from_arc` and `num_biases_from_arc` from within the namelist? We can do that now but it'd be replaced with the calibrated parameters pretty soon.

src/closures/nondimensional_exchange_functions.jl
costachris: That makes sense if we choose to always have 0 biases, but would remove the capability to calibrate biases altogether.

driver/generate_namelist.jl
ilopezgp: Yes, I meant that. If you think we are going to replace these by values soon, then probably there is no need.

src/closures/nondimensional_exchange_functions.jl
ilopezgp: can't we have the choice of biases/no biases as a boolean in the namelist, and then we can choose whether to calibrate them or not simply by changing that?

costachris: ahh okay I follow. So you're suggesting a single nn parameter vector like before. When `biases` is `false` its size is equal to the number of weights and when  `biases` = `true` it's equal to the number of weights + biases?

ilopezgp: Yes! I think that's the minimal working solution for what we need

costachris: I made `nn_ent_biases` a flag instead and took it out of the local climaparameters

ilopezgp (approved):
LGTM

PR #901, opened 2022-03-26
update deps

trontrytel:


trontrytel:
Looks like it downgrades Julia to 1.7.0 from 1.7.2

Damn. I just want to be able to use my new parameter from CLIMAParameters v0.4.2 :(

charleskawczynski (approved):

Issue #903, opened 2022-03-27
CI broke with clearing depot


charleskawczynski:
We recently needed to clear the depot to upgrade some deps. This broke CI, and it looks like things break [here](https://github.com/SciML/DiffEqNoiseProcess.jl/blob/04090fdbc30547be4ad46ad4ced7e2a1f3e0fcc7/src/wiener.jl#L15), (which gets called in the prognostic stochastic closure examples).

I don't think there's a bullet proof way around this, unfortunately. We could stop using the depot, but then CI times will climb back up to ~30 min, which isn't terrible. Idk which is the worse of the two evils 🤷🏻‍♂️

PR #902, opened 2022-03-26
add microphysics params for calibration

trontrytel:

PR #904, opened 2022-03-27
Temporary fix for CI

charleskawczynski:
This PR adds `DiffEqNoiseProcess` as a dependency (which should be removed eventually), and overloads DiffEqNoiseProcess.jl's method `DiffEqNoiseProcess.wiener_randn!(rng::Random.AbstractRNG, rand_vec)` for ClimaCore `FieldVector`s. This fixes CI, but we should probably find a better way to do this, or submit a PR to DiffEqNoiseProcess that will allow this to work for us. The main constraint, from what I can tell, is that they're call calls `axes`, which is not a single axis for FieldVectors since they can live on multiple spaces, so the axes cannot be represented so simply.

This is only relevant to the prognostic stochastic closures, it seems. So, @haakon-e, please review when you have a chance.


charleskawczynski:
Closes #903.

trontrytel:
Would be great to get that fix merged in today. So that we can merge #902 and #672 too

haakon-e (approved):
LGTM

Issue #868, opened 2022-03-05
PR #862 difficults error interpretation during calibration, leads to ridiculous stack traces


ilopezgp:
Changes in PR #862 lead to unfathomable stack traces in CalibrateEDMF, see the log in the TurbulenceConvectionUtils test suite in https://github.com/CliMA/CalibrateEDMF.jl/pull/255. Imagine this message x 20 for a single ensemble member during calibration.

charleskawczynski:
I think there's been some effort in making truncated stack traces, if not we can look into doing this ourselves-- I don't think we need anything too fancy, just having function names and line numbers would probably be helpful enough.

ilopezgp:
For now, would eliminating the back_stacktrace be enough? I would like to understand what led @haakon-e to add this and perhaps we can find a simpler way to get the same information. I suppose it comes from a need to get errors from StochasticDiffEq?

charleskawczynski:
I'm fine with temporarily reverting #862 if that makes calibration more useable.

ilopezgp:
> I'm fine with temporarily reverting #862 if that makes calibration more useable.

Let's hear from @haakon-e first, since he seemed to have the need for this change. If he is ok with it we can revert and open a new issue.

haakon-e:
we can revert for now, but It'd be nice if we could make this a flag or similar later. I sometimes find full stack traces very useful even in the calibration mode.

ilopezgp:
> These should probably be written out as (parseable / structured) log files rather than dumped to stdout, at the very least it should be printing to `stderr` rather than `stdout` so it can be re-directed appropriately which might be enough for the cluster use case.

If we can write CalibrateEDMF errors to `stderr` and log TC.jl errors as warnings written to `stdout`, that would work for us.

PR #906, opened 2022-03-29
Revert full stack trace introduced in PR #862

ilopezgp:
Comments out the stack trace print introduced in PR #862, since it makes debugging very difficult, as discussed in #868 .


charleskawczynski (approved):
Fine by me.

PR #909, opened 2022-04-04
Monitor dependencies are updated

ilopezgp:
Right now we depend on users randomly enforcing that dependencies are updated when opening PRs, which can lead to conflicts between merged code and locally (correctly) ran code.

This PR implements a CI workflow that monitors whether all manifests are up to date when staging a PR.

### DIFF
--- a/.github/workflows/UpDeps.yml
+++ b/.github/workflows/UpDeps.yml
@@ -1,0 +1,34 @@
+name: UpDeps
+
+on:
+  push:
+    branches:
+      - main
+      - trying
+      - staging
+    tags: '*'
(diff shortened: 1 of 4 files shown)


charleskawczynski (commented):
I would just change the title and description a bit-- this is only monitoring / providing an indicator that deps can be updated. The only real enforcement we have is through bors checks, and we can't depend on this check because dep updates could happen during a merge, resulting in passing tests on the staging branch and failed tests on merge, which doesn't seem like a good idea.

ilopezgp:
Title and description changed!

PR #910, opened 2022-04-05
Draft of non-eqilibrium moisture

trontrytel:
@jbphyswx and me were brainstorming today about how to add the non-equlibrium moisture. We are still ironing out the details of how to do it in the environment. The updrafts and the grid mean equations however are easier to start from. The things to do would be to:
- [ ] Add `q_liq` and `q_ice` related fluxes, tendencies, etc wherever necessary. With that finished we should be able to run `TC.jl` with `q_liq` and `q_ice` computed offline for the updrafts.
- [ ] Change the thermodynamic state constructors to the non-equilibrium ones.

This draft tries to tackle the first issue. Of course we want those changes to be optional based on some `namelist` argument. I wasn't sure how to do it, especially on the host model dycore side.

@charleskawczynski, it would be great if you could take a look at the extent of the damage :) Maybe in the upcoming days we could pair program with @jbphyswx to make this draft a namelist optional list of changes. 

Additionally there is a lot of repetition between `q_liq` and `q_ice`. Maybe there are some other ways where we could avoid it?

### DIFF
--- a/driver/dycore.jl
+++ b/driver/dycore.jl
@@ -437,9 +437,9 @@ function compute_gm_tendencies!(
     ∇v_sgs = CCO.DivergenceF2C()
 
     @. tends_θ_liq_ice += -α0_c * ∇θ_liq_ice_sgs(wvec(sgs_flux_θ_liq_ice))
-    @. tends_q_tot += -α0_c * ∇q_tot_sgs(wvec(sgs_flux_q_tot))
-    @. tends_q_liq += -α0_c * ∇q_liq_sgs(wvec(sgs_flux_q_liq))
-    @. tends_q_ice += -α0_c * ∇q_ice_sgs(wvec(sgs_flux_q_ice))
+    @. tends_q_tot += -α0_c * ∇q_sgs(wvec(sgs_flux_q_tot))
+    @. tends_q_liq += -α0_c * ∇q_sgs(wvec(sgs_flux_q_liq))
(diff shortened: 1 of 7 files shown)


driver/dycore.jl
```diff
@@ -304,15 +312,22 @@ function compute_gm_tendencies!(
...
     RBθ = CCO.RightBiasedC2F(; top = CCO.SetValue(θ_liq_ice_gm_toa))
     RBq = CCO.RightBiasedC2F(; top = CCO.SetValue(q_tot_gm_toa))
+    RBq_liq = CCO.RightBiasedC2F(; top = CCO.SetValue(q_liq_gm_toa))
+    RBq_ice = CCO.RightBiasedC2F(; top = CCO.SetValue(q_ice_gm_toa))
     wvec = CC.Geometry.WVector
     ∇c = CCO.DivergenceF2C()
     @. ∇θ_liq_ice_gm = ∇c(wvec(RBθ(prog_gm.θ_liq_ice)))
     @. ∇q_tot_gm = ∇c(wvec(RBq(prog_gm.q_tot)))
+    @. ∇q_liq_gm = ∇c(wvec(RBq(prog_gm.q_liq)))
+    @. ∇q_ice_gm = ∇c(wvec(RBq(prog_gm.q_ice)))
```
charleskawczynski, line 326: Is this used anywhere?

trontrytel:
I'm closing this one. Everything is now part of #928

PR #917, opened 2022-04-08
add more parameters for calibration

trontrytel:
This PR adds more parameters for calibration. It requires CLIMAParameters v0.4.3 (the new patch release).

For future reference, the correct workflow to merge this is to: 
1) open a separate PR with up_deps
2) clear the depot for the CI
3) bump version (it's a breaking change since now we are only compatible with the new version of CP.jl)
4) merge
5) tag a new release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 authors = ["Climate Modeling Alliance"]
 name = "TurbulenceConvection"
 uuid = "8e072fc4-01f8-44fb-b9dc-f9336c367e6b"
-version = "0.17.4"
+version = "0.18.0"
 
 [compat]
 CLIMAParameters = "0.4"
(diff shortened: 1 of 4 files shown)


charleskawczynski (approved):

charleskawczynski:
Open an `up_deps.jl` PR first, then rebase/merge this

ilopezgp:
After this is merged, let's do a breaking release please, not a patch

trontrytel:
> After this is merged, let's do a breaking release please, not a patch

Yes I agree. It was a patch release for CLIMAParameters, because I was only extending the list of parameters. (So no breaking changes for its users downstream).

The changes proposed in this PR are only compatible with the new version of CLIMAParameters. So it is a breaking change for TC.jl

charleskawczynski:
Actually, I don't think a rebase is needed. Let's try

trontrytel:
bors try

PR #922, opened 2022-04-11
first attempt for adding RRTMGP in TC

yairchn:
I will followup here what we learn when working on adding RRTMGP in ClimaAtmos and hope to bring this to a complete PR


integration_tests/Manifest.toml
```diff
@@ -1279,6 +1279,12 @@ version = "0.5.6"
 deps = ["InteractiveUtils", "Markdown", "Sockets", "Unicode"]
 uuid = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 
+[[deps.RRTMGP]]
+deps = ["Adapt", "CLIMAParameters", "CUDA", "Compat", "Coverage", "DocStringExtensions", "GaussQuadrature", "NCDatasets", "Pkg", "Profile", "StaticArrays", "Test"]
+path = "../../RRTMGP.jl"
+uuid = "a01a1ee8-cea4-48fc-987c-fc7878d79da1"
+version = "0.3.0"
```
charleskawczynski, line 1286: @yairchn, I've released a new version of RRTMGP.jl, so we can add it proper, rather than adding a specific version.

driver/rrtmgp_model.jl
```diff
@@ -0,0 +1,947 @@
+push!(LOAD_PATH, joinpath(@__DIR__, "..", ".."))
```
charleskawczynski, line 1: I'm fine with most of these changes, but I think these file contents will be merged into RRTMGP.jl soon, and it may be better to wait until that happens before merging this PR (I think there are still details being worked out in this file).

charleskawczynski (commented):
Looks great, I think:
 - We should wait until `driver/rrtmgp_model.jl` is merged into RRTMGP.jl (which will be soon AFAICT)
 - add a buildkite job to test this! Feel free to add a cli option to specify radiation

driver/rrtmgp_model.jl
yairchn: define soon? I am happy to followup with a PR to undo these changes once its done

charleskawczynski: Fair point. I'm happy to temporarily merge this after we add a buildkite job to show the output.

charleskawczynski:
I think we just need to rebase now for CI to work.

PR #923, opened 2022-04-11
[WIP] Cc/refactor parameters

costachris:
Add ClimaParameters for perturbation pressure closure and entrainment/detrainment.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -714,20 +714,35 @@ function CasesBase(case::T; inversion_type, surf_params, Fo, Rad, LESDat = nothi
 end
 
 struct EDMFParameters{FT,AECPS,AEDSPS}
+    a_surf::FT
+    a_min::FT
+    a_max::FT
+    Pr_n::FT
+    α_d::FT
+    α_a::FT
+    α_b::FT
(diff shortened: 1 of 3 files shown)


src/types.jl
```diff
@@ -690,27 +713,30 @@ function CasesBase(case::T; inversion_type, surf_params, Fo, Rad, LESDat = nothi
     )
 end
 
-struct EDMFParameters{FT,AECPS}
+struct EDMFParameters{FT,AECPS,AEDSPS}
```
costachris, line 716: @odunbar Is this the structure we want to pass around through the code base (which will house all the parameters currently defined in `EDMFModel`)? So instead of `EDMFModel` having the attributes [`entr_closure_parameters`, `entr_dim_scale_parameters`, etc.], it will just have an `EDMFParameters` attribute?

odunbar: Right!  The EDMFParameters is a box that contains
1. parameters called directly by EDMF and not any sub-module (e.g. not called in any closure etc.)
2. The boxes for all the submodules. (e.g. the box for a closure)

PR #925, opened 2022-04-13
Add and update some docs and doc stubs

charleskawczynski:
This PR:
 - Moves the README table to the developer docs (I wrote this when trying to understand the equations)
 - Moves the index.md contents to a newly added API.md, where we can have/add doc string links
 - Adds Formulation.md, where we can add (perhaps some from the EDMFEquations.md) a story of the EDMF formulation including figures and equations
 - Silences the printed statements in ReferenceStates.md


ilopezgp (approved):
I actually like the prognostic/diagnostic table, I think it is very useful. I would keep it in the README; everything else looks good to me,

PR #928, opened 2022-04-18
Optional non-equilibrium moisture model

trontrytel:
This is a draft PR with the non-equilibrium moisture model. The general description of what we are trying to do is here: https://www.overleaf.com/read/hgzwbvbfkqzw

This PR implements the first step from the above document. It (only) adds the non-equilibrium moisture to the updraft equations (equation (7) from the above document). It doesn't do anything about the second order terms related to the cloud liquid and cloud ice variables. This means we are limiting ourselves to not using the quadratures for the buoyancy gradients and the sgs source terms in the environment.

I made sure to code it as an option for `TC.jl`. I don't expect to use it in "operational" `TC.jl` simulations soon. But it is a very interesting research project led by @jbphyswx . 

It would be great if @yairchn and @ilopezgp could take a look and check if I'm updating all the necessary equations and if @charleskawczynski could take a look and check if there is anything that can be optimised at this stage. And that @jbphyswx could double check that the proposed changes are in line with what we have drafted before. 

A couple of additional notes:
- I did not change how the buoyancy gradients are calculated. In theory the moist gas constants should be computed based on the non-equilibrium thermodynamic state. But for the sake of simplicity I did not want to add additional options there. I don't think it should matter.
- The `src/update_aux.jl` file needs some additional cleaning. For this PR my main goal was to introduce minimal changes to the code structure. As a result, for the non-equilibrium option we are updating the `q_liq` and `q_ice` values twice. - Once during the "prognostic update" of the auxiliary variables, and another time in the "diagnostic update" of the auxiliary variables. This should not matter because for the case of non-equilibrium moisture, the thermodynamic state is constructed by passing a full phase partition tuple which is already populated based on the prognostic variables. So in short, the values are being overwritten with the same values. In the future it would be good to improve the `src/update_aux.jl` file. 
- I have assumed that the boundary condition fluxes for `q_liq` and `q_ice` are zero.
- I have added the `ql_tendency` and `qi_tendency` to the `PrecipFormation` struct. We don't need them for the equilibrium moisture model. But I thought that the code readability is better this way than duplicating the code only to once populate the `qt_tendency` and other time to also populate the `ql_tendency` and `qi_tendency`.
- In the `driver/dycore.jl` I am extending the list of prognostic variables based on the `edmf.moisture_model` choice. I know that in principle the driver fields should not depend on the edmf model fields. But to me this seemed like a smaller change than introducing some hypothetical option for the hypothetical driver model. 

Those are the namelist fields that have to be changed from defaults to make the Bomex simulation run:

```
    # TODO - move to one of the "Vanilla + Deviations" tests                    
    namelist["thermodynamics"]["moisture_model"] = "nonequilibrium"             
    namelist["thermodynamics"]["sgs"] = "mean"                                  
    namelist["turbulence"]["EDMF_PrognosticTKE"]["env_buoy_grad"] = "mean"      
    namelist["time_stepping"]["adapt_dt"] = false  
```

### DIFF
--- a/driver/Surface.jl
+++ b/driver/Surface.jl
@@ -73,6 +73,8 @@ function get_surface(
         ρθ_liq_ice_flux = shf / TD.cp_m(param_set, ts_in),
         ρq_tot_flux = lhf / TD.latent_heat_vapor(param_set, ts_in),
         wstar = convective_vel,
+        ρq_liq_flux = FT(0),
+        ρq_ice_flux = FT(0),
     )
 end
 
@@ -131,6 +133,8 @@ function get_surface(
         ρq_tot_flux = lhf / TD.latent_heat_vapor(param_set, ts_in),
         bflux = result.buoy_flux,
         wstar = convective_vel,
+        ρq_liq_flux = FT(0),
+        ρq_ice_flux = FT(0),
     )
 end
 
@@ -189,6 +193,8 @@ function get_surface(
         ρq_tot_flux = lhf / TD.latent_heat_vapor(param_set, ts_in),
         bflux = result.buoy_flux,
         wstar = convective_vel,
+        ρq_liq_flux = FT(0),
+        ρq_ice_flux = FT(0),
     )
 end
 
@@ -249,5 +255,7 @@ function get_surface(
         ρq_tot_flux = lhf / TD.latent_heat_vapor(param_set, ts_in),
         bflux = result.buoy_flux,
         wstar = convective_vel,
+        ρq_liq_flux = FT(0),
+        ρq_ice_flux = FT(0),
     )
 end
--- a/driver/compute_diagnostics.jl
+++ b/driver/compute_diagnostics.jl
@@ -130,17 +130,41 @@ function compute_diagnostics!(
 
     surf = get_surface(case.surf_params, grid, state, t, param_set)
 
-    @inbounds for k in TC.real_center_indices(grid)
-        ts = TD.PhaseEquil_pθq(param_set, p0_c[k], prog_gm.θ_liq_ice[k], prog_gm.q_tot[k])
-        aux_gm.s[k] = TD.specific_entropy(param_set, ts)
-        ts_en = TD.PhaseEquil_pθq(param_set, p0_c[k], aux_en.θ_liq_ice[k], aux_en.q_tot[k])
-        aux_en.s[k] = TD.specific_entropy(param_set, ts_en)
-        @inbounds for i in 1:N_up
-            if aux_up[i].area[k] > 0.0
-                ts_up = TD.PhaseEquil_pθq(param_set, p0_c[k], aux_up[i].θ_liq_ice[k], aux_up[i].q_tot[k])
-                aux_up[i].s[k] = TD.specific_entropy(param_set, ts_up)
+    if edmf.moisture_model isa TC.EquilibriumMoisture
+        @inbounds for k in TC.real_center_indices(grid)
+            ts = TD.PhaseEquil_pθq(param_set, p0_c[k], prog_gm.θ_liq_ice[k], prog_gm.q_tot[k])
+            aux_gm.s[k] = TD.specific_entropy(param_set, ts)
+            ts_en = TD.PhaseEquil_pθq(param_set, p0_c[k], aux_en.θ_liq_ice[k], aux_en.q_tot[k])
+            aux_en.s[k] = TD.specific_entropy(param_set, ts_en)
+            @inbounds for i in 1:N_up
+                if aux_up[i].area[k] > 0.0
+                    ts_up = TD.PhaseEquil_pθq(param_set, p0_c[k], aux_up[i].θ_liq_ice[k], aux_up[i].q_tot[k])
+                    aux_up[i].s[k] = TD.specific_entropy(param_set, ts_up)
+                else
+                    aux_up[i].s[k] = FT(0)
+                end
             end
         end
+    elseif edmf.moisture_model isa TC.NonEquilibriumMoisture
+        @inbounds for k in TC.real_center_indices(grid)
+            q = TD.PhasePartition(prog_gm.q_tot[k], prog_gm.q_liq[k], prog_gm.q_ice[k])
+            ts = TD.PhaseNonEquil_pθq(param_set, p0_c[k], prog_gm.θ_liq_ice[k], q)
+            aux_gm.s[k] = TD.specific_entropy(param_set, ts)
+            q_en = TD.PhasePartition(aux_en.q_tot[k], aux_en.q_liq[k], aux_en.q_ice[k])
+            ts_en = TD.PhaseNonEquil_pθq(param_set, p0_c[k], aux_en.θ_liq_ice[k], q_en)
+            aux_en.s[k] = TD.specific_entropy(param_set, ts_en)
+            @inbounds for i in 1:N_up
+                if aux_up[i].area[k] > 0.0
+                    q_up = TD.PhasePartition(aux_up[i].q_tot[k], aux_up[i].q_liq[k], aux_up[i].q_ice[k])
+                    ts_up = TD.PhaseNonEquil_pθq(param_set, p0_c[k], aux_up[i].θ_liq_ice[k], q_up)
+                    aux_up[i].s[k] = TD.specific_entropy(param_set, ts_up)
+                else
+                    aux_up[i].s[k] = FT(0)
+                end
+            end
+        end
+    else
+        error("The expected moisture models are EquilibriumMoisture or NonEquilibriumMoisture, not $(edmf.moisture_model)")
     end
 
     # TODO(ilopezgp): Fix bottom gradient
--- a/driver/dycore.jl
+++ b/driver/dycore.jl
@@ -26,7 +26,22 @@ const APS = CLIMAParameters.AbstractEarthParameterSet
 aux_vars_ref_state(FT) = (; ref_state = (ρ0 = FT(0), α0 = FT(0), p0 = FT(0)))
 
 # Center only
-cent_aux_vars_gm(FT) = (;
+cent_aux_vars_gm_moisture(FT, ::TC.NonEquilibriumMoisture) = (;
+    ∇q_liq_gm = FT(0),
+    ∇q_ice_gm = FT(0),
+    dqldt_rad = FT(0),
+    dqidt_rad = FT(0),
+    dqldt = FT(0),
+    dqidt = FT(0),
+    dqldt_hadv = FT(0),
+    dqidt_hadv = FT(0),
+    ql_nudge = FT(0),
+    qi_nudge = FT(0),
+    dqldt_fluc = FT(0),
+    dqidt_fluc = FT(0),
+)
+cent_aux_vars_gm_moisture(FT, ::TC.EquilibriumMoisture) = ()
+cent_aux_vars_gm(FT, edmf) = (;
     tke = FT(0),
     Hvar = FT(0),
     QTvar = FT(0),
@@ -49,10 +64,10 @@ cent_aux_vars_gm(FT) = (;
     dTdt = FT(0), #Large-scale temperature tendency
     dqtdt = FT(0), #Large-scale moisture tendency
     dTdt_hadv = FT(0), #Horizontal advection of temperature
-    H_nudge = FT(0), #Reference H profile for relaxation tendency
-    dTdt_fluc = FT(0), #Vertical turbulent advection of temperature
     dqtdt_hadv = FT(0), #Horizontal advection of moisture
+    H_nudge = FT(0), #Reference H profile for relaxation tendency
     qt_nudge = FT(0), #Reference qt profile for relaxation tendency
+    dTdt_fluc = FT(0), #Vertical turbulent advection of temperature
     dqtdt_fluc = FT(0), #Vertical turbulent advection of moisture
     u_nudge = FT(0), #Reference u profile for relaxation tendency
     v_nudge = FT(0), #Reference v profile for relaxation tendency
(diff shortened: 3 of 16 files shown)


driver/compute_diagnostics.jl
```diff
@@ -130,17 +130,37 @@ function compute_diagnostics!(
...
+            @inbounds for i in 1:N_up
+                if aux_up[i].area[k] > 0.0
+                    q_up = TD.PhasePartition(aux_up[i].q_tot[k], aux_up[i].q_liq[k], aux_up[i].q_ice[k])
+                    ts_up = TD.PhaseNonEquil_pθq(param_set, p0_c[k], aux_up[i].θ_liq_ice[k], q_up)
+                    aux_up[i].s[k] = TD.specific_entropy(ts_up)
+                end
             end
         end
+    else
+        error("Something is wrong in diagnostics. The expected moisture models are Equilibrium or NonEquilibrium")
```
ilopezgp, line 163:
```suggestion
        error("The expected moisture models are EquilibriumMoisture or NonEquilibriumMoisture, not $(edmf.moisture_model)")
```

driver/dycore.jl
```diff
@@ -338,6 +387,14 @@ function compute_gm_tendencies!(
             end
             tendencies_gm.θ_liq_ice[k] += aux_gm.dTdt[k] / Π
             tendencies_gm.q_tot[k] += aux_gm.dqtdt[k]
+            if edmf.moisture_model isa TC.NonEquilibriumMoisture
+                if force.apply_subsidence
```
ilopezgp, line 391: why is this flag not used in L379 too?

src/EDMF_Precipitation.jl
```diff
@@ -75,8 +83,15 @@ function compute_precipitation_sink_tendencies(::Clima1M, grid::Grid, state::Sta
...
-        q = TD.PhasePartition(ts)
+        if edmf.moisture_model isa EquilibriumMoisture
+            ts = TD.PhaseEquil_pTq(param_set, p0, T_gm, q_tot_gm)
+            q = TD.PhasePartition(ts)
+        elseif edmf.moisture_model isa NonEquilibriumMoisture
+            q_liq_gm = prog_gm.q_liq[k]
+            q_ice_gm = prog_gm.q_ice[k]
+            q = TD.PhasePartition(q_tot_gm, q_liq_gm, q_ice_gm)
+            ts = TD.PhaseNonEquil_pTq(param_set, p0, T_gm, q)
+        end
```
ilopezgp, line 94: add else with error?

src/EDMF_Updrafts.jl
```diff
@@ -1,3 +1,43 @@
...
+    @inbounds for i in 1:N_up
+        @inbounds for k in real_center_indices(grid)
+            T_up = aux_up[i].T[k]
+            q_up = TD.PhasePartition(aux_up[i].q_tot[k], aux_up[i].q_liq[k], aux_up[i].q_ice[k])
+            ts_up = TD.PhaseNonEquil_pTq(param_set, p0_c[k], T_up, q_up)
+
+            # condensation/evaporation, deposition/sublimation
+            mph = noneq_moisture_sources(param_set, aux_up[i].area[k], ρ0_c[k], Δt, ts_up)
+            aux_up[i].ql_tendency_noneq_moisture[k] = mph.ql_tendency * aux_up[i].area[k]
+            aux_up[i].qi_tendency_noneq_moisture[k] = mph.qi_tendency * aux_up[i].area[k]
```
ilopezgp, line 27: maybe remove `moisture`? i.e. `aux_up[i].ql_tendency_noneq[k] = ...`

src/EDMF_functions.jl
```diff
@@ -533,6 +627,22 @@ function filter_updraft_vars(edmf::EDMFModel, grid::Grid, state::State, surf::Su
             end
         end
     end
+    if edmf.moisture_model isa NonEquilibriumMoisture
+        @inbounds for k in real_center_indices(grid)
+            @inbounds for i in 1:N_up
+                is_surface_center(grid, k) && continue
+                prog_up[i].ρaq_tot[k] = max(prog_up[i].ρaq_tot[k], 0)
```
ilopezgp, line 634: is this the centralized location where we clip things?

driver/compute_diagnostics.jl
```diff
@@ -130,17 +130,37 @@ function compute_diagnostics!(
...
+    if edmf.moisture_model isa TC.EquilibriumMoisture
+        @inbounds for k in TC.real_center_indices(grid)
+            ts = TD.PhaseEquil_pθq(param_set, p0_c[k], prog_gm.θ_liq_ice[k], prog_gm.q_tot[k])
+            aux_gm.s[k] = TD.specific_entropy(ts)
+            ts_en = TD.PhaseEquil_pθq(param_set, p0_c[k], aux_en.θ_liq_ice[k], aux_en.q_tot[k])
+            aux_en.s[k] = TD.specific_entropy(ts_en)
+            @inbounds for i in 1:N_up
+                if aux_up[i].area[k] > 0.0
+                    ts_up = TD.PhaseEquil_pθq(param_set, p0_c[k], aux_up[i].θ_liq_ice[k], aux_up[i].q_tot[k])
+                    aux_up[i].s[k] = TD.specific_entropy(ts_up)
```
charleskawczynski, line 142: Can we define `aux_up[i].s[k]` both inside and outside `if aux_up[i].area[k] > 0.0`? Doing this is risky because it's effectively a race condition in time. Concretely, the value, if `!aux_up[i].area[k] > 0.0`, can take on values computed at some previous time.

Same for below. I'd prefer these functions compute variables in the entire space, rather than conditionally compute variables-- even if that means setting the variables to zero in the else-case.

driver/initial_conditions.jl
```diff
@@ -15,9 +15,17 @@ function initialize_edmf(edmf::TC.EDMFModel, grid::TC.Grid, state::TC.State, cas
     prog_gm = TC.center_prog_grid_mean(state)
     p0_c = TC.center_ref_state(state).p0
     parent(aux_tc.prandtl_nvec) .= edmf.prandtl_number
-    @inbounds for k in TC.real_center_indices(grid)
-        ts = TD.PhaseEquil_pθq(param_set, p0_c[k], prog_gm.θ_liq_ice[k], prog_gm.q_tot[k])
-        aux_gm.θ_virt[k] = TD.virtual_pottemp(ts)
+    if edmf.moisture_model isa TC.EquilibriumMoisture
```
charleskawczynski, line 18: I'm wondering if it would be useful to change this to:

```julia
    if TC.is_equilibrium_moisture(edmf)
```
And then define

```julia
is_equilibrium_moisture(edmf) = edmf.moisture_model isa EquilibriumMoisture
```
in `TC.jl/src`.

src/EDMF_functions.jl
```diff
@@ -294,6 +346,12 @@ function q_surface_bc(surf::SurfaceBase{FT}, grid::Grid, state::State, edmf::EDM
     surface_scalar_coeff = percentile_bounds_mean_norm(1 - a_total + i * a_, 1 - a_total + (i + 1) * a_, 1000)
     return prog_gm.q_tot[kc_surf] + surface_scalar_coeff * sqrt(qt_var)
 end
+function ql_surface_bc(surf::SurfaceBase{FT})::FT where {FT}
+    return FT(0)
+end
+function qi_surface_bc(surf::SurfaceBase{FT})::FT where {FT}
+    return FT(0)
+end
```
charleskawczynski, line 354: There's a lot of `ql = 0`, `qi = 0`, is this just to support things for when they become non-zero?

charleskawczynski (commented):
The PR looks good to me. It looks like we would be able to benefit quite a bit if we could capture thermo states but, unfortunately, that's not been merged yet. Doing that would allow us to store thermo states in the aux field, and then avoid the many if-else statements. I'll bump the priority on this since it's starting to incur issues.

src/EDMF_functions.jl
trontrytel: I think that surface bc for liquid and ice should be zero. Or at least I can't imagine a good case for when they shouldn't be.

driver/initial_conditions.jl
trontrytel: We have two options `EquilibriumMoisture` and `NonEquilibriumMoisture` for `edmf.moisture_model`. In most places I was trying to code it with 
 
```
if EquilibriumMoisture
     ...
elseif NonEquilibriumMoisture
    ...
else
    error
end 
```
I did not follow with error here (i.e. in the `driver/initial_conditions.jl`) because I was thinking that for some dry cases we would want to initialise without any moisture. 

Anyway, I can define `is_equilibrium_moisture(edmf)` and `is_non_equilibrium_moisture(edmf)`, but I'm not sure if it's a big readability improvement? I thought that `if edmf.moisture_model isa TC.EquilibriumMoisture` was pretty self explanatory?

src/EDMF_functions.jl
trontrytel: For updrafts, yes AFAIK. This function (`filter_updraft_vars`) is called from `affect_filter` function which also adjusts the (co)variances: [
](https://github.com/CliMA/TurbulenceConvection.jl/blob/dc20da185aeed7f4b4b0f8ea36a6451dff4635c4/src/EDMF_functions.jl#L171-L199)

driver/dycore.jl
trontrytel: I was following the logic for the `qt` tendencies here. 

I guess that if we chose the `DYCOMS_RF01` forcing we have to apply the subsidence. Whereas if we chose the `Standard` forcing we can still pick and choose whether we apply subsidence or not?

driver/compute_diagnostics.jl
trontrytel: Sounds good. I'll put zeros there and see what happens 👀

driver/initial_conditions.jl
charleskawczynski: Fair point, either way sounds good to me

trontrytel:
Here are the preliminary results from TRMM.

[profiles_TRMM_neq.pdf](https://github.com/CliMA/TurbulenceConvection.jl/files/8535768/profiles_TRMM_neq.pdf)
[paths_TRMM_neq.pdf](https://github.com/CliMA/TurbulenceConvection.jl/files/8535771/paths_TRMM_neq.pdf)

I think it's a good start. I will add TRMM to Vanilla Experiments CI and then try to merge it in.

trontrytel:
Not sure why the windows job is breaking. I don't think it's related to this PR?

Otherwise it's ready. You can see the plots in Vanilla+deviations experiments which now has TRMM running with non-equilibrium moisture

charleskawczynski (approved):
Nicely done. The code-related changes all look good to me. I think the windows issue is unrelated--I don't think we need to worry about it, especially since it seems to happen in the IO section.

My only comments are:
 - Maybe we can open an issue to add thermo state fields and isolate the if-else statements. This should help trim things down a bit, but I think that's fine to put in a follow-up PR.
 - It might be good to also open an issue about moving some of the overleaf docs into the TC docs.
 - Thank you for adding a new job, the new physics is exciting! 🎉

ilopezgp (approved):
LGTM

src/EDMF_Precipitation.jl
```diff
@@ -75,8 +83,17 @@ function compute_precipitation_sink_tendencies(::Clima1M, grid::Grid, state::Sta
...
         # When we fuse loops, this should hopefully disappear
-        ts = TD.PhaseEquil_pTq(param_set, p0, T_gm, q_tot_gm)
-        q = TD.PhasePartition(param_set, ts)
+        if edmf.moisture_model isa EquilibriumMoisture
+            ts = TD.PhaseEquil_pTq(param_set, p0, T_gm, q_tot_gm)
+            q = TD.PhasePartition(param_set, ts)
+        elseif edmf.moisture_model isa NonEquilibriumMoisture
+            q_liq_gm = prog_gm.q_liq[k]
+            q_ice_gm = prog_gm.q_ice[k]
+            q = TD.PhasePartition(q_tot_gm, q_liq_gm, q_ice_gm)
```
yairchn, line 92: is `q` a typical name for phase partition ? is that a symbol you would be using in literature ? otherwise it might confusing with the various forms of specific humidity

src/microphysics_coupling.jl
```diff
@@ -1,3 +1,44 @@
...
+
+        # TODO - is that the state we want to be relaxing to?
+        ts_eq = TD.PhaseEquil_ρTq(param_set, ρ0, T, q.tot)
+        q_eq = TD.PhasePartition(param_set, ts_eq)
+
+        S_ql = CM1.conv_q_vap_to_q_liq_ice(param_set, liq_type, q_eq, q)
+        S_qi = CM1.conv_q_vap_to_q_liq_ice(param_set, ice_type, q_eq, q)
+
+        # TODO - handle limiters elswhere
+        if S_ql >= FT(0)
```
yairchn, line 25: I am assuming that once the limiter are moved "elsewhere" the dependency on `Δt` will be removed, right?

driver/Surface.jl
```diff
@@ -73,6 +73,8 @@ function get_surface(
         ρθ_liq_ice_flux = shf / TD.cp_m(param_set, ts_in),
         ρq_tot_flux = lhf / TD.latent_heat_vapor(param_set, ts_in),
         wstar = convective_vel,
+        ρq_liq_flux = FT(0),
```
yairchn, line 76: is there anything we need to add to SF.jl for that?

src/types.jl
```diff
@@ -476,6 +507,9 @@ struct EDMFModel{N_up, FT, PM, ENT, EBGC, EC, EDS, EPG}
         else
             error("Something went wrong. Invalid environmental buoyancy gradient closure type '$en_sgs_name'")
         end
+        if moisture_model_name == "nonequilibrium" && en_thermo_name == "quadrature"
```
yairchn, line 510: If you have the error message here do you need it in line 248 in `src/EDMF_Environment.jl` as well?

yairchn (approved):
I left a few minor comments / questions but otherwise looks great to me

src/EDMF_Precipitation.jl
trontrytel: I don't think there is a typical symbol for "phase partition". I think it's a name specific to our code and Thermodynamics.jl data structures. 

Some arguments in favour of it:
1) `q` is sometimes used to denote specific humidity. Thats why we are using `q_tot`, `q_liq`, etc for our water tracers. (One has to be careful though, because sometimes in the literature `q` is used for  mixing ratio. And sometimes some other symbol is used). 

2) Usually I'm against using one letter variable names. Here however `q` is created to encapsulate `(q_tot, q_liq, q_ice)` only to pass them into thermodynamic state. It's not used further. So I think it's ok. 

3) We have been using `q` like that in other places, because once you want to access it's fields you write `q.tot`, `q.liq`, etc. Which I think it's nice.

src/microphysics_coupling.jl
trontrytel: Yes :) As soon as we figure out the timestepping "elsewhere" all of this can be removed.

driver/Surface.jl
trontrytel: I'm not sure. I could not imagine a case for prescribing a flux of liquid water or ice into the atmosphere (a geyser?) So I  hard coded them as zeros in our surface conditions. But maybe we need more flexibility than that?

Do you think it's better to move it into `SF.jl`?

src/types.jl
trontrytel: Right. I'm probably checking the same thing twice here. I'll leave it for now, but maybe it would be good to take a look at all our error messages and do some cleanups.

trontrytel:
> * Maybe we can open an issue to add thermo state fields and isolate the if-else statements. This should help trim things down a bit, but I think that's fine to put in a follow-up PR.

👍 

> * It might be good to also open an issue about moving some of the overleaf docs into the TC docs.

💯 Great point! I forgot we now have the docs ;) I'll  try to merge this PR because it's already long enough. But I will follow with a docs update very soon.

PR #929, opened 2022-04-19
Add multiple updraft buildkite job

charleskawczynski:
I'm impressed that this is working out of the box! Are there other things we could add to visualization? I'm curious (and suspect) what the distribution is of different variables between updrafts.

This PR adds a `n_up` cli option, and fixes a pretty big bug introduced in #932 (parsed args were set to cli options, and not the other way around) 😬.


trontrytel:
What else do we need to do here? I see we should update the mse table for the 2 updraft case, but I could not find the new values in the logs?

charleskawczynski:
Working on it, I'm trying to fix the cli options because the driver is getting a bit out of control

charleskawczynski:
Alright, driver config is much simpler now, I'm pretty confident this should work

PR #932, opened 2022-04-20
Refactor cli options

charleskawczynski:
This PR refactors the CLI options by:

 - Adding explicit flags for skipping post-processing and tests
 - Using ArgParse.jl's parsing

@trontrytel, I've removed the `run_from_command_line` flag, but I think we can use `@isdefined parsed_args` to make REPL development easier. Once we create a `parsed_args`, we can reset any of the args to whatever we'd like:

```julia
parsed_args["entr"] = "FNO"
```

This should make #929 much easier to get working without piling on more hacks.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -92,43 +92,43 @@ steps:
     steps:
 
       - label: ":partly_sunny: Bomex with gaussian"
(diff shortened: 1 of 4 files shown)


trontrytel (approved):
If it makes the CI more manageable, then it sounds good to me.

ilopezgp (approved):
LGTM

PR #947, opened 2022-04-26
Use thermo state field cache more

charleskawczynski:
This PR adds more use of the thermo state field cache. I think this is refactoring for the most part, but there is one behavioral change in `src/EDMF_Environment.jl`. @trontrytel, can you take a look and confirm that this looks okay? (the cached thermo state was constructed using the if-else block, and the one replaced does not)


post_processing/mse_tables.jl
```diff
@@ -112,20 +112,20 @@ all_best_mse["life_cycle_Tan2018"]["updraft_thetal"] = 0.0
 all_best_mse["life_cycle_Tan2018"]["v_mean"] = 0.0
 all_best_mse["life_cycle_Tan2018"]["u_mean"] = 0.0
 all_best_mse["life_cycle_Tan2018"]["tke_mean"] = 0.0
-all_best_mse["life_cycle_Tan2018"]["temperature_mean"] = 6.285660087520289e-18
+all_best_mse["life_cycle_Tan2018"]["temperature_mean"] = 0.0
```
trontrytel, line 115: Why do we now have so many zeros in the mses?

src/update_aux.jl
```diff
@@ -146,30 +148,15 @@ function update_aux!(edmf::EDMFModel, grid::Grid, state::State, surf::SurfaceBas
...
-                aux_en.q_tot[k],
-                aux_en.q_liq[k],
-                aux_en.q_ice[k],
-            )
+            (aux_en.q_liq[k], aux_en.q_ice[k])
         else
             error("Something went wrong. The moisture_model options are equilibrium or nonequilibrium")
         end
-        aux_gm.θ_virt[k] = TD.virtual_pottemp(param_set, ts_gm)
+        ts_env[k] = thermo_state_pθq(param_set, p0_c[k], aux_en.θ_liq_ice[k], aux_en.q_tot[k], thermo_args...)
```
trontrytel, line 158: very cool

trontrytel:
>the cached thermo state was constructed using the if-else block, and the one replaced does not

Not sure I see the thing you are asking about @charleskawczynski ? Seems like there was an if/else due to moisture model in the environment?

src/EDMF_Environment.jl
```diff
@@ -323,7 +317,7 @@ function microphysics(
 
         else
             # if variance and covariance are zero do the same as in SA_mean
-            ts = thermo_state_pθq(param_set, p0_c[k], aux_en.θ_liq_ice[k], aux_en.q_tot[k])
+            ts = ts_env[k]
```
charleskawczynski, line 320: @trontrytel this is the line that is technically behavior changing. This doesn’t have equil-nonequil branches

trontrytel: Right. It doesn't matter. We don't support the non-equilibrium option with the sgs quadratures, so the user will get an error somewhere before that line. 

I'll add the docs soon.

trontrytel (approved):
LGTM. The only sus part are the zeros in the mse tables. And why there are so many changes there in the first place

post_processing/mse_tables.jl
charleskawczynski: This was a head scratcher. Then I remembered, we have a precedence order for comparing datasets, and TC main comes [second](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/post_processing/compute_mse.jl#L445) after LES (and we don’t have LES for life cycle). We should simplify our regression tests, or make them clearer somehow.. one thing at a time.

charleskawczynski:
I can explain the many changes tomorrow, it’s a bit confusing..

charleskawczynski:
But unrelated to this PR

charleskawczynski:
Bors r+

PR #955, opened 2022-04-28
Fix massflux bug for N_up>1

yairchn:
The current computation of the mass flux is true to what is written in the second row in equation (10) Cohen et al 2020 assuming `i=0` in the first summation and summing over `j` only. To code the second line in this equation for multiple updraft one would have to use two for-loops (for the two summations) and add factor `1/2`. 
Instead I am coding here the first row in equation (10) Cohen et al 2020, by taking the difference from the grid mean for each subdomain and having the correct flux coded with only one for-loop. 
The bug originates in SCAMPy https://github.com/CliMA/SCAMPy/blob/master/Turbulence_PrognosticTKE.pyx


charleskawczynski (approved):
Code changes look good to me.

src/EDMF_functions.jl
```diff
@@ -70,51 +72,58 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
-        aux_up_i = aux_up[i]
         θ_liq_ice_up = aux_up_i.θ_liq_ice
         q_tot_up = aux_up_i.q_tot
-        massflux = aux_up_f[i].massflux
+        @. aux_up_f[i].massflux = ρ0_f * Ifau(a_up) * (w_up - w_gm)
         # We know that, since W = 0 at z = 0, m = 0 also, and
         # therefore θ_liq_ice / q_tot values do not matter
-        @. massflux_h += massflux * (If(θ_liq_ice_up) - If(θ_liq_ice_en))
-        @. massflux_qt += massflux * (If(q_tot_up) - If(q_tot_en))
+        @. massflux_h += ρ0_f * (Ifau(a_up) * (w_up - w_gm) * (If(θ_liq_ice_up) - If(θ_liq_ice_en)))
```
ilopezgp, line 100: if you are adding the contribution over all subdomains, you should subtract the gmv values to the up values, not the en values.

src/EDMF_functions.jl
```diff
@@ -70,51 +72,58 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
+        @. massflux_qt += ρ0_f * (Ifau(a_up) * (w_up - w_gm) * (If(q_tot_up) - If(q_tot_en)))
```
ilopezgp, line 101: Same here

src/EDMF_functions.jl
```diff
@@ -70,51 +72,58 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
             aux_up_f_i = aux_up_f[i]
             aux_up_i = aux_up[i]
             q_liq_up = aux_up_i.q_liq
             q_ice_up = aux_up_i.q_ice
-            massflux = aux_up_f[i].massflux
-            @. massflux_ql += massflux * (If(q_liq_up) - If(q_liq_en))
-            @. massflux_qi += massflux * (If(q_ice_up) - If(q_ice_en))
+            massflux_up_i = aux_up_f[i].massflux
+            @. massflux_ql += massflux_up_i * (w_up - w_gm) * (If(q_liq_up) - If(q_liq_gm))
+            @. massflux_qi += massflux_up_i * (w_up - w_gm) * (If(q_ice_up) - If(q_ice_gm))
```
ilopezgp, line 123: This doesn't seem right. You are multiplying a mass flux by vertical velocity again.

ilopezgp:
Apart from the technicalities, I am all in favor to change to the first line of eqn (10), instead of the second.

ilopezgp (approved):
LGTM

yairchn:
Bors r+

PR #956, opened 2022-04-28
change of grid mean prognostic variables

yairchn:
In this PR I take the first step in modifying the grid mean prognostic variables. 
The current PR changes are from 
`θ_liq_ice` --> `ρθ_liq_ice`
`q_tot` --> `ρq_tot`
I do so will adding `θ_liq_ice` and `q_tot` to the `aux_gm` fields and using it in places. 
My hope is that once I change to `ρθ_liq_ice` --> `ρe_tot` we can add `aux_up.e_tot` and `aux_en.e_tot` for computing the SGS energy flux while effectively not requiring any mathematical translation of the flux from `θ_liq_ice` to energy.


charleskawczynski (approved):
The PR looks great, I think you just need to
* Update the mse tables
* Update your local version of the julia formatter (`julia --project=.dev ] up`), then apply the formatter
* Fix the `--entr FNO` case
* Fix the `--entr NN_nonlocal` case

yairchn:
Bors r+

PR #957, opened 2022-04-28
Add moisture model docs

trontrytel:
As promised, following up with a quick docs entry related to `moisture_model`.


docs/src/EDMFEquations.md
```diff
@@ -664,6 +664,23 @@ r_{con} & = \frac{\qt+\ql}{1 - \qt} \\
...
 \end{align}
 ```
+## Moisture model
+
+There are two options for representing moisture (water vapor, cloud liquid water and cloud ice):
+  (i) equilibrium, and (ii) non-equilibrium.
+In the equilibrium approach we are solving prognostic equations for total water specific humidity $q_t$ only.
+The cloud condensate is diagnosed using saturation adjustment algorithm and then
+  partitioned into cloud liquid water $q_l$ and cloud ice $q_i$ using an empirical phase partition function $\lambda(T)$.
+In the non-equilibroum approach we are solving prognostic equations for $q_t$, $q_l$ and $q_i$.
```
ilopezgp, line 674:
```suggestion
In the non-equilibrium approach we are solving prognostic equations for $q_t$, $q_l$ and $q_i$.
```

docs/src/EDMFEquations.md
```diff
@@ -664,6 +664,23 @@ r_{con} & = \frac{\qt+\ql}{1 - \qt} \\
...
+The source terms for $q_l$ and $q_i$ are parameterized as relaxation terms.
+We don't solve for the additional (co)variances related to $q_l$ and $q_i$ in the environment.
+As a result, the non-equilibroum option is only available when running simulations
```
ilopezgp, line 677:
```suggestion
As a result, the non-equilibrium option is only available when running simulations
```

docs/src/EDMFEquations.md
trontrytel: Thanks!

charleskawczynski (approved):
LGTM

We could decompose the EDMF equations docs into sections, I didn’t do that in climate machine to avoid too many files, but it may actually make sense in the TC repo. We can certainly leave that for a separate PR, though.

PR #959, opened 2022-04-29
Enforce definition of nz or dz in LESdrivenSCM, not both.

ilopezgp:
Enforce definition of nz or dz in LESdrivenSCM, not both. The behavior before this PR accepted a definition of `nz`, but discarded it silently, which is misleading to users.

The current implementation allows defining either nz or dz.

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.2"
+julia_version = "1.7.0"
 manifest_format = "2.0"
 
 [[deps.ANSIColoredPrinters]]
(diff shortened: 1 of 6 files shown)


docs/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
```
yairchn, line 2: you might need to merge an updated main branch

ilopezgp: What do you mean?

yairchn: I am wondering why you have Manifest changes

ilopezgp: Because I used up_deps.jl to keep the dependencies up to date. You can see below that the UpDeps CI check is now passing.

ilopezgp:
I will deal with the mse flag as soon as I get approval.

charleskawczynski (approved):
Code changes look fine to me. I'd prefer we update deps in separate PRs, just because we can get mse table changes from that alone and doing this helps differentiate changes. That said, I don't want to slow anything down and this looks fine to me.

ilopezgp:
> Code changes look fine to me. I'd prefer we update deps in separate PRs, just because we can get mse table changes from that alone and doing this helps differentiate changes. That said, I don't want to slow anything down and this looks fine to me.

I will take it into account for next time, thanks!

PR #960, opened 2022-04-30
use total energy as grid mean variable

yairchn:


driver/Surface.jl
```diff
@@ -67,7 +67,7 @@ function get_surface(
         ch = result.Ch,
         ρu_flux = surf_params.zero_uv_fluxes ? FT(0) : result.ρτxz,
         ρv_flux = surf_params.zero_uv_fluxes ? FT(0) : result.ρτyz,
-        ρθ_liq_ice_flux = shf / TD.cp_m(param_set, ts_in),
+        ρe_tot_flux = shf,
```
tapios, line 70: Is the lhf contribution to `ρe_tot_flux` taken into account somewhere?

driver/Surface.jl
```diff
@@ -123,7 +123,7 @@ function get_surface(
         ustar = result.ustar,
         ρu_flux = result.ρτxz,
         ρv_flux = result.ρτyz,
-        ρθ_liq_ice_flux = shf / TD.cp_m(param_set, ts_in),
+        ρe_tot_flux = shf,
```
tapios, line 126: Same as above: lhf contribution to `ρe_tot_flux`?

driver/Surface.jl
yairchn: you are right, I kept mixing up the evaporation that included in SHF with the LHF contribution to the energy. 
I believe the right flux is:
`ρe_tot_flux = shf + lhf`

Sadly it did not fix the problems completely but things look better and the code is more correct.

driver/dycore.jl
```diff
@@ -250,14 +257,45 @@ function set_thermo_state!(state, grid, moisture_model, param_set)
...
+    prog_gm_f = TC.face_prog_grid_mean(state)
+    aux_gm = TC.center_aux_grid_mean(state)
+    p0_c = TC.center_ref_state(state).p0
+    ρ0_c = TC.center_ref_state(state).ρ0
+    g = CPP.grav(param_set)
+    @inbounds for k in TC.real_center_indices(grid)
+        thermo_args = if moisture_model isa TC.EquilibriumMoisture
+            ()
+        elseif moisture_model isa TC.NonEquilibriumMoisture
+            (prog_gm.q_liq[k], prog_gm.q_ice[k])
```
charleskawczynski, line 282: We'll need to (eventually) update this to `(aux_gm.q_liq[k], aux_gm.q_ice[k])` or `(prog_gm.ρq_liq[k], prog_gm.ρq_ice[k])`

driver/dycore.jl
```diff
@@ -485,41 +531,40 @@ function compute_gm_tendencies!(
...
-            aux_tc.qt_tendency_precip_sinks[k]
-        tendencies_gm.θ_liq_ice[k] +=
-            aux_bulk.θ_liq_ice_tendency_precip_formation[k] +
-            aux_en.θ_liq_ice_tendency_precip_formation[k] +
-            aux_tc.θ_liq_ice_tendency_precip_sinks[k]
-        if edmf.moisture_model isa TC.NonEquilibriumMoisture
-            tendencies_gm.q_liq[k] += aux_bulk.ql_tendency_precip_formation[k] + aux_en.ql_tendency_precip_formation[k]
-            tendencies_gm.q_ice[k] += aux_bulk.qi_tendency_precip_formation[k] + aux_en.qi_tendency_precip_formation[k]
-        end
+        # tendencies_gm.ρq_tot[k] +=
```
yairchn, line 534: @trontrytel can you help converting these tendencies to total energy (I believe you already have it done elsewhere)

driver/Surface.jl
```diff
@@ -67,7 +67,7 @@ function get_surface(
         ch = result.Ch,
         ρu_flux = surf_params.zero_uv_fluxes ? FT(0) : result.ρτxz,
         ρv_flux = surf_params.zero_uv_fluxes ? FT(0) : result.ρτyz,
-        ρθ_liq_ice_flux = shf / TD.cp_m(param_set, ts_in),
+        ρe_tot_flux = shf + lhf,
```
yairchn, line 70: @tapios @ilopezgp 
please review this change

driver/dycore.jl
```diff
@@ -376,14 +404,14 @@ function compute_gm_tendencies!(
...
-    RBθ = CCO.RightBiasedC2F(; top = CCO.SetValue(θ_liq_ice_gm_toa))
+    h_tot_gm_toa = (prog_gm.ρe_tot[kc_toa] + p0_c[kc_toa]) / ρ0_c[kc_toa]
+    q_tot_gm_toa = prog_gm.ρq_tot[kc_toa] / ρ0_c[kc_toa]
+    RBe = CCO.RightBiasedC2F(; top = CCO.SetValue(h_tot_gm_toa))
     RBq = CCO.RightBiasedC2F(; top = CCO.SetValue(q_tot_gm_toa))
     wvec = CC.Geometry.WVector
     ∇c = CCO.DivergenceF2C()
-    @. ∇θ_liq_ice_gm = ∇c(wvec(RBθ(aux_gm.θ_liq_ice)))
-    @. ∇q_tot_gm = ∇c(wvec(RBq(aux_gm.q_tot)))
+    @. ∇h_tot_gm = ∇c(wvec(RBe((prog_gm.ρe_tot + p0_c) / ρ0_c)))
```
yairchn, line 413: @tapios  @@ilopezgp 
please review this change and the `toa` BC above. 
I am not happy with this change since we are using the advective form for subsidence `w(∂ϕ/∂z)` it is not clear to me if `ϕ` should be total enthalpy (`h`) or total energy (`e`). The formulation that introduces total enthalpy is a flux form one. Formally, starting from the flux form total energy equation and continuity equation I get: 

```
∂ρe/∂t + ∇[u(ρe + p)] = ρS
e(∂ρ/∂t + ∇[ρu]) = 0 --> ignoring S_ρ in the continuity here

∂ρe/∂t - e∂ρ/∂t = ρ(∂e/∂t)
∇[u(ρe + p)] - e∇[ρu] = ρu∇e + ∇(up)

∂e/∂t + u∇e + (1/ρ)∇(up) = S
```
that the subsidence should be

`w∂e/∂z + (1/ρ)(∂wp/∂z)`

if that makes sense?

driver/dycore.jl
```diff
@@ -483,9 +516,13 @@ function compute_gm_tendencies!(
                 end
             end
 
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * (H_horz_adv + gm_H_nudge_k + H_fluc + gm_H_subsidence_k)
             tendencies_gm.ρq_tot[k] +=
                 ρ0_c[k] * (aux_gm.dqtdt_hadv[k] + gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k] + gm_QT_subsidence_k)
+            tendencies_gm.ρe_tot[k] +=
```
yairchn, line 521: ignore microphysics changes @trontrytel has a PR for that

src/EDMF_functions.jl
```diff
@@ -74,17 +77,24 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
     w_gm = prog_gm_f.w
-    θ_liq_ice_en = aux_en.θ_liq_ice
-    θ_liq_ice_gm = aux_gm.θ_liq_ice
+    h_tot_en = copy(a_en)
     q_tot_gm = aux_gm.q_tot
     q_tot_en = aux_en.q_tot
     a_en_bcs = a_en_boundary_conditions(surf, edmf)
     Ifae = CCO.InterpolateC2F(; a_en_bcs...)
     If = CCO.InterpolateC2F(; bottom = CCO.SetValue(FT(0)), top = CCO.SetValue(FT(0)))
+    Ic = CCO.InterpolateF2C()
```
yairchn, line 86: @tapios @ilopezgp 
below are changes to the mass flux, writing it for total enthalpy. It could be that the `p/ ρ` would cancel out as they are identical to all subdomains and grid mean, but I kept them in for consistency.

src/EDMF_functions.jl
```diff
@@ -183,29 +207,37 @@ function compute_diffusive_fluxes(edmf::EDMFModel, grid::Grid, state::State, sur
...
+    @. h_tot_en =
+        TD.total_energy(param_set, ts_en, 0.5 * (prog_gm.u^2 + prog_gm.v^2 + Ic(aux_en_f.w)^2), grid.zc.z * g) +
+        p0_c / ρ0_c
 
     @. aux_tc_f.ρ_ae_KH = IfKH(aeKH) * ρ0_f
     @. aux_tc_f.ρ_ae_KM = IfKM(aeKM) * ρ0_f
 
     aeKHq_tot_bc = -surf.ρq_tot_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KH[kf_surf]
-    aeKHθ_liq_ice_bc = -surf.ρθ_liq_ice_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KH[kf_surf]
+    aeKHh_tot_bc = -surf.ρe_tot_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KH[kf_surf]
```
yairchn, line 229: @tapios @ilopezgp 
Here the lower BC for turbulent flux of `h` is using what SurfaceFluxes.jl gives, i.e. the energy flux (shf+lhf).
Is that correct?

src/variables.jl
```diff
@@ -38,6 +38,7 @@ cent_aux_vars_up(FT, edmf) = (;
     q_tot = FT(0),
     θ_liq_ice = FT(0),
     θ_liq_ice_tendency_precip_formation = FT(0),
```
yairchn, line 40: @trontrytel removing `θ_liq_ice_tendency_precip_formation` here breaks the code, I am wondering which one did we mean to keep

trontrytel: It shouldn't. I was able to run TRMM simulation with the microphysics vars changed. Something must have gone wrong with the merge. 

Do you have a version without it? I would rather apply the changes again and push than debug this

driver/Surface.jl
yairchn: These are computed in SF.jl using:

`shf =  -ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E`
`lhf = (hv_sfc + Φ_sfc) * E`
`E = -ρ_sfc * Ch * windspeed(sc) * Δqt(param_set, sc)`

src/EDMF_functions.jl
tapios: Does `surf.ρe_tot_flux` have the enthalpy flux in it? This would be important.

driver/dycore.jl
```diff
@@ -240,15 +239,41 @@ function compute_ref_state!(state, grid::TC.Grid, param_set::PS; ts_g) where {PS
     return nothing
 end
 
-function set_thermo_state!(state, grid, moisture_model, param_set)
+function set_thermo_state_peq!(state, grid, moisture_model, param_set)
     ts_gm = TC.center_aux_grid_mean(state).ts
     prog_gm = TC.center_prog_grid_mean(state)
+    prog_gm_f = TC.face_prog_grid_mean(state)
     aux_gm = TC.center_aux_grid_mean(state)
     p0_c = TC.center_ref_state(state).p0
```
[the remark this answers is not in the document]
yairchn: yes - we would do that in a separate PR if that is ok?

src/EDMF_functions.jl
yairchn: here 
`surf.ρe_tot_flux = shf + lhf` 
and 
`shf = -ρ_sfc * Ch * windspeed(sc) * (cp_m * ΔT + ΔΦ) - (hd_sfc) * E`
I believe the `cp_m` is there to make it an enthalpy flux, right?

driver/dycore.jl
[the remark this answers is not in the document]
charleskawczynski: Opened #971. We can follow that up.

src/EDMF_functions.jl
```diff
@@ -97,12 +107,17 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
         @. aux_up_f[i].massflux = ρ0_f * Ifau(a_up) * (w_up - w_gm)
         # We know that, since W = 0 at z = 0, m = 0 also, and
-        # therefore θ_liq_ice / q_tot values do not matter
-        @. massflux_h += ρ0_f * (Ifau(a_up) * (w_up - w_gm) * (If(θ_liq_ice_up) - If(θ_liq_ice_gm)))
+        # therefore h_tot / q_tot values do not matter
+        ts_up_i = copy(ts_en)
+        h_tot_up_i = copy(q_tot_up)
+        @. ts_up_i = thermo_state_pθq(param_set, p0_c, θ_liq_ice_up, q_tot_up)
+        @. h_tot_up_i =
+            TD.total_energy(param_set, ts_up_i, 0.5 * (prog_gm.u^2 + prog_gm.v^2 + Ic(w_up)^2), grid.zc.z * g) +
```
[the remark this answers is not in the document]
charleskawczynski: @yairchn, feel free to cache `geopotential` if it makes this easier.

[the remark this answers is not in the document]
trontrytel: That would be helpful. I need geopotential for microphysics coupling as well

src/EDMF_functions.jl
tapios: That looks correct. The important part is that there is also the gravitational term, so it's a DSE difference that enters, as it should.

src/EDMF_functions.jl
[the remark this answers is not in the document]
tapios: @charleskawczynski It would be good to do this throughout ClimaAtmos. Will make it much easier to generalize the code to other situations.

driver/dycore.jl
```diff
@@ -240,15 +239,41 @@ function compute_ref_state!(state, grid::TC.Grid, param_set::PS; ts_g) where {PS
...
     ρ0_c = TC.center_ref_state(state).ρ0
+    g = CPP.grav(param_set)
     @inbounds for k in TC.real_center_indices(grid)
-        aux_gm.θ_liq_ice[k] = prog_gm.ρθ_liq_ice[k] / ρ0_c[k]
+        thermo_args = if moisture_model isa TC.EquilibriumMoisture
+            ()
```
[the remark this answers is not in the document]
charleskawczynski: Yes, the compiler does inference on the type of `moisture_model` and does "union-splitting"-- effectively eliminating the unused branch.

src/EDMF_functions.jl
```diff
@@ -74,17 +77,24 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
+    @. h_tot_en =
+        TD.total_energy(param_set, ts_en, 0.5 * (prog_gm.u^2 + prog_gm.v^2 + Ic(w_en)^2), grid.zc.z * g) + p0_c / ρ0_c
 
     # Compute the mass flux and associated scalar fluxes
+    parent(massflux) .= 0
+    parent(massflux_h) .= 0
+    parent(massflux_qt) .= 0
     @. massflux = ρ0_f * Ifae(a_en) * (w_en - w_gm)
-    @. massflux_h = ρ0_f * Ifae(a_en) * (w_en - w_gm) * (If(θ_liq_ice_en) - If(θ_liq_ice_gm))
+    @. massflux_h = ρ0_f * Ifae(a_en) * (w_en - w_gm) * (If(h_tot_en) - If(prog_gm.ρe_tot / ρ0_c + p0_c / ρ0_c))
```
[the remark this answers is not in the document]
charleskawczynski: Yes, we can add that to the cache.

driver/dycore.jl
yairchn: I could use 
`DSE = h_tot - Lv*q_tot` 
right ?

Also, currently on main branch and in scampy (and I believe also in pycles) we simply multiply the subsidence velocity by the gradient of the prognostic variable ([here](https://github.com/CliMA/TurbulenceConvection.jl/blob/92ed24bd713250f2ebe9f2ab40f64e5a89b1d5c5/driver/dycore.jl#L425)), which is not consistent with the DSE. If we chose DSE here it could have behavioral changes even if it is more correct.

src/EDMF_functions.jl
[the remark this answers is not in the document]
charleskawczynski: Opened [CA#401](https://github.com/CliMA/ClimaAtmos.jl/issues/401).

driver/dycore.jl
```diff
@@ -257,6 +282,10 @@ function set_thermo_state!(state, grid, moisture_model, param_set)
             error("Something went wrong. The moisture_model options are equilibrium or nonequilibrium")
         end
         ts_gm[k] = TC.thermo_state_pθq(param_set, p0_c[k], aux_gm.θ_liq_ice[k], aux_gm.q_tot[k], thermo_args...)
+        e_kin = 0.5 * (prog_gm.u[k]^2 + prog_gm.v[k]^2) # add w here?
```
tapios, line 285: We need w^2 in here too. And I think there's a ClimaCore function that computes the vector norm directly, with the correct metrics in curvilinear coordinates. We should use it.

driver/dycore.jl
```diff
@@ -257,6 +282,10 @@ function set_thermo_state!(state, grid, moisture_model, param_set)
...
+        e_pot = grid.zc.z[k] * g
```
tapios, line 286: Replace by geopotential throughout.

driver/dycore.jl
tapios: > I could use `DSE = h_tot - Lv*q_tot` right ?
> 

This is only approximately right (without condensate, and it has a kinetic energy term, which shouldn't be there). Simpler to use `DSE = c_pd (T-T_0) + geopotential`. (The T_0 term might be omitted because it's constant, but it may be nice to include it here for greater consistency with the moist case.)

Re: behavioral changes. If we are correcting errors (and having subsidence act on theta_li in cloudy areas is an error), we should have behavioral changes. However, I believe the changes will be small because they mainly effect of subsidence relative to other terms is in cloud-free areas.

driver/dycore.jl
```diff
@@ -398,20 +428,23 @@ function compute_gm_tendencies!(
...
             tendencies_gm.u[k] -= force.coriolis_param * (aux_gm.vg[k] - prog_gm.v[k])
             tendencies_gm.v[k] += force.coriolis_param * (aux_gm.ug[k] - prog_gm.u[k])
         end
         if TC.rad_type(radiation) <: Union{TC.RadiationDYCOMS_RF01, TC.RadiationLES}
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * aux_gm.dTdt_rad[k] / Π
+            tendencies_gm.ρe_tot[k] += ρ0_c[k] * cp_m * aux_gm.dTdt_rad[k]
         end
         if TC.force_type(force) <: TC.ForcingDYCOMS_RF01
             tendencies_gm.ρq_tot[k] += ρ0_c[k] * aux_gm.dqtdt[k]
+            tendencies_gm.ρe_tot[k] += ρ0_c[k] * (L_v0 - R_d * T_0) * aux_gm.dqtdt[k]
```
tapios, line 443: This doesn't quite look right. It should be `h_v` (enthalpy of vapor) in place of the term in parenthesis, which also includes a sensible heat component (but is an enthalpy, rather than an internal energy, as here). And even that is only true in the unsaturated case, where dq_t = dq_v.

driver/dycore.jl
```diff
@@ -398,20 +428,23 @@ function compute_gm_tendencies!(
...
+
             # Apply large-scale subsidence tendencies
-            tendencies_gm.ρθ_liq_ice[k] -= ρ0_c[k] * ∇θ_liq_ice_gm[k] * aux_gm.subsidence[k]
-            tendencies_gm.ρq_tot[k] -= ρ0_c[k] * ∇q_tot_gm[k] * aux_gm.subsidence[k]
+            tendencies_gm.ρe_tot[k] -= ρ0_c[k] * aux_gm.subsidence[k] * ∇h_tot_gm[k]
```
tapios, line 446: This should act on DSE.

driver/dycore.jl
```diff
@@ -422,11 +455,11 @@ function compute_gm_tendencies!(
 
         if TC.force_type(force) <: TC.ForcingStandard
             if force.apply_subsidence
-                tendencies_gm.ρθ_liq_ice[k] -= ρ0_c[k] * ∇θ_liq_ice_gm[k] * aux_gm.subsidence[k]
-                tendencies_gm.ρq_tot[k] -= ρ0_c[k] * ∇q_tot_gm[k] * aux_gm.subsidence[k]
+                tendencies_gm.ρe_tot[k] -= ρ0_c[k] * aux_gm.subsidence[k] * ∇h_tot_gm[k]
```
tapios, line 458: DSE

driver/dycore.jl
```diff
@@ -466,8 +499,8 @@ function compute_gm_tendencies!(
 
             if force.apply_subsidence
                 # Apply large-scale subsidence tendencies
-                gm_H_subsidence_k = -∇θ_liq_ice_gm[k] * aux_gm.subsidence[k]
-                gm_QT_subsidence_k = -∇q_tot_gm[k] * aux_gm.subsidence[k]
+                gm_H_subsidence_k = - ρ0_c[k] * aux_gm.subsidence[k] * ∇h_tot_gm[k]
```
tapios, line 502: DSE

driver/dycore.jl
```diff
@@ -483,9 +516,15 @@ function compute_gm_tendencies!(
                 end
             end
 
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * (H_horz_adv + gm_H_nudge_k + H_fluc + gm_H_subsidence_k)
             tendencies_gm.ρq_tot[k] +=
                 ρ0_c[k] * (aux_gm.dqtdt_hadv[k] + gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k] + gm_QT_subsidence_k)
+            tendencies_gm.ρe_tot[k] +=
+                ρ0_c[k] * (
+                    gm_H_subsidence_k +
+                    cp_m * Π * (H_horz_adv + gm_H_nudge_k + H_fluc) +
```
tapios, line 524: This looks odd. Why is there a `Π` (Exner function?) here? Are the other tendencies potential temperature tendencies, e.g., owing to advection? If it's a temperature tendency, there should be no Exner function, and we should have cv_m in place of cp_m.

driver/dycore.jl
```diff
@@ -483,9 +516,15 @@ function compute_gm_tendencies!(
...
+                    (L_v0 - R_d * T_0) *
```
tapios, line 525: Need enthalpy of vapor `h_v` here as above (and even that only holds when it's unsaturated).

src/EDMF_functions.jl
```diff
@@ -74,17 +77,27 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
     Ifae = CCO.InterpolateC2F(; a_en_bcs...)
     If = CCO.InterpolateC2F(; bottom = CCO.SetValue(FT(0)), top = CCO.SetValue(FT(0)))
+    Ic = CCO.InterpolateF2C()
+
+    # compute total enthalpies
+    ts_en = center_aux_environment(state).ts
+    ts_gm = center_aux_grid_mean(state).ts
+    @. h_tot_gm = prog_gm.ρe_tot / ρ0_c + TD.gas_constant_air(param_set, ts_gm) * aux_gm.T
+    @. h_tot_en = (
+        TD.total_energy(param_set, ts_en, 0.5 * (prog_gm.u^2 + prog_gm.v^2 + Ic(w_en)^2), grid.zc.z * g) +
```
tapios, line 94: I believe we have a ClimaCore function to correctly compute kinetic energies from 2-norms. This here only holds in Cartesian geometry.

tapios (commented):
A couple issues in forcing terms that do not look consistent.

src/EDMF_functions.jl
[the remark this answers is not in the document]
yairchn: @tapios I am assuming you referring to the geopoetntial expression:
`Φ(z) = Gm(1/(a) - 1/(a+z))` with Gm = 6.673^-11 as gravitational constant?

at the moment I cannot seem to find a geopotential function either than 
`Φ(z) = grav * z` see [here](https://github.com/CliMA/ClimaCore.jl/blob/b9cc1abbdeb3b8b8db0b11aca6216d655e0684e4/examples/hybrid/sphere/solid_body_rotation_3d.jl#L72) in ClimaCore, or ClimaAtmos
nor could I find the gravitational constant in ClimaParameters.

driver/dycore.jl
yairchn: I think will have to wait for us to change the horizontal velocities to `UVVector`, right @charleskawczynski ?

src/EDMF_functions.jl
yairchn: I think will have to wait for us to change the horizontal velocities to `UVVector`, right @charleskawczynski ?

driver/dycore.jl
yairchn: This `(L_v0 - R_d * T_0)` term came when converting thetali flux to enthalpy flux to avoid behavioral changes. 
@trontrytel can you share the derivation of the conversion? 
I will test the result of using `h_v ` and send a comparison

driver/dycore.jl
yairchn: these 
`H_horz_adv + gm_H_nudge_k + H_fluc` 
are tendencies of potential temperature, e.g. [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/9adff06b016ea08c9849c100c47675895873a557/driver/dycore.jl#L483)

driver/dycore.jl
yairchn: I will test the result of using `h_v` and send a comparison

driver/dycore.jl
yairchn: there is no subsidence acting on momentum?

tapios: No, we don't usually do that.

driver/dycore.jl
charleskawczynski: No, we don’t have to wait, just add what’s needed and we’ll follow up on the types after.

src/EDMF_functions.jl
charleskawczynski: Again, no need to wait. Add what’s needed and we’ll follow up on the types in the next PR. That’s when we can add the use of the ClimaCore norm function

driver/dycore.jl
tapios: We should add a DSE function to thermodynamics and compute it there. Let's just define it so that MSE is DSE plus the latent heat terms at the reference temperature, i.e., using the decomposition in Eq. (2.13) of the design doc:

`DSE = c_pm(q)(T - T_0) + (1-q_t)R_d T_0 + Phi'

so that `MSE = DSE + q+l L_{v,0} - q_i L_{f,0}`.

tapios: (The (1-q_t) term in the DSE is approximately constant, so it will not matter much. It has to be somewhere for consistency; we could stick it either into the DSE or into the MSE definition.)

driver/dycore.jl
tapios: It probably won't make a practical difference. The above form in the code assumes constant `q_l` and `q_i` while `q_t` changes--which is reasonable. You can leave it as it is.

driver/dycore.jl
tapios: Probably still needs to be c_vm in place of c_pm.

driver/dycore.jl
tapios: As above, c_vm in place of c_pm?

src/EDMF_functions.jl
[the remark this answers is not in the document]
yairchn: @tapios @charleskawczynski I added these to clima parameters in this [PR](https://github.com/CliMA/CLIMAParameters.jl/pull/72) 
but we would need to wait of the next CP release before we can use it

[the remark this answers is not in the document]
tapios: I only meant to define `Φ(z) = grav * z` somewhere for all functions to use. No need for anything more complicated right now. I just want to make it easy to convert to deep atmospheres one day. 

(And then, in the deep case, you also need to add the centrifugal potential. But let's not trouble with it now.)

driver/Cases.jl
```diff
@@ -1223,7 +1202,7 @@ function surface_params(case::LES_driven_SCM, grid::TC.Grid, surf_ref_state, par
         mean_qt_prof = Statistics.mean(data.group["profiles"]["qt_mean"][:][:, imin:imax], dims = 2)[:]
         field = TC.FieldFromNamedTuple(TC.face_space(grid), (; ρq_tot = FT(0)))
         Ic = CCO.InterpolateF2C()
-        q_tot_c = Ic.(field.ρq_tot) #Yair devide by rho here
+        q_tot_c = Ic.(field.ρq_tot)
```
ilopezgp, line 1205: why do we have q = ρq here?

driver/Cases.jl
```diff
@@ -1064,9 +1045,8 @@ function initialize_profiles(self::CasesBase{SP}, grid::Grid, param_set, state)
         prog_gm.u[k] = APL.SP_u(FT)(z)
         prog_gm.v[k] = APL.SP_v(FT)(z)
         aux_gm.θ_liq_ice[k] = APL.SP_θ_liq_ice(FT)(z)
-        prog_gm.ρθ_liq_ice[k] = ρ0_c[k] * aux_gm.θ_liq_ice[k]
+
         aux_gm.q_tot[k] = APL.SP_q_tot(FT)(z)
-        prog_gm.ρq_tot[k] = ρ0_c[k] * aux_gm.q_tot[k]
```
ilopezgp, line 1069: why are we no longer initializing prognostic thermodynamic profiles?

driver/dycore.jl
```diff
@@ -240,15 +239,45 @@ function compute_ref_state!(state, grid::TC.Grid, param_set::PS; ts_g) where {PS
...
-function set_thermo_state!(state, grid, moisture_model, param_set)
+function set_thermo_state_peq!(state, grid, moisture_model, param_set)
+    Ic = CCO.InterpolateF2C()
     ts_gm = TC.center_aux_grid_mean(state).ts
     prog_gm = TC.center_prog_grid_mean(state)
+    prog_gm_f = TC.face_prog_grid_mean(state)
     aux_gm = TC.center_aux_grid_mean(state)
     p0_c = TC.center_ref_state(state).p0
     ρ0_c = TC.center_ref_state(state).ρ0
+    w_c = copy(prog_gm.ρe_tot)
```
ilopezgp, line 250: is this setting the vertical velocity to energy?

driver/dycore.jl
```diff
@@ -422,11 +462,11 @@ function compute_gm_tendencies!(
...
         if TC.force_type(force) <: TC.ForcingStandard
             if force.apply_subsidence
-                tendencies_gm.ρθ_liq_ice[k] -= ρ0_c[k] * ∇θ_liq_ice_gm[k] * aux_gm.subsidence[k]
-                tendencies_gm.ρq_tot[k] -= ρ0_c[k] * ∇q_tot_gm[k] * aux_gm.subsidence[k]
+                tendencies_gm.ρe_tot[k] -= ρ0_c[k] * aux_gm.subsidence[k] * ∇h_tot_gm[k]
+                tendencies_gm.ρq_tot[k] -= ρ0_c[k] * aux_gm.subsidence[k] * ∇q_tot_gm[k]
             end
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * aux_gm.dTdt[k] / Π
             tendencies_gm.ρq_tot[k] += ρ0_c[k] * aux_gm.dqtdt[k]
+            tendencies_gm.ρe_tot[k] += ρ0_c[k] * (cp_m * aux_gm.dTdt[k] + h_v * aux_gm.dqtdt[k])
```
ilopezgp, line 469: shouldn't this be `cv_m`, since we are adding it to an energy (not enthalpy)?

driver/dycore.jl
```diff
@@ -438,8 +478,8 @@ function compute_gm_tendencies!(
         end
 
         if TC.force_type(force) <: TC.ForcingLES
-            H_horz_adv = aux_gm.dTdt_hadv[k] / Π
-            H_fluc = aux_gm.dTdt_fluc[k] / Π
+            H_horz_adv = aux_gm.dTdt_hadv[k]
+            H_fluc = aux_gm.dTdt_fluc[k]
```
ilopezgp, line 482: This should be made more clear. H before was theta, now we are saying that H is sensible temperature? Shouldn't it be total energy?

driver/dycore.jl
```diff
@@ -466,8 +506,8 @@ function compute_gm_tendencies!(
 
             if force.apply_subsidence
                 # Apply large-scale subsidence tendencies
-                gm_H_subsidence_k = -∇θ_liq_ice_gm[k] * aux_gm.subsidence[k]
-                gm_QT_subsidence_k = -∇q_tot_gm[k] * aux_gm.subsidence[k]
+                gm_H_subsidence_k = - ρ0_c[k] * aux_gm.subsidence[k] * ∇h_tot_gm[k]
+                gm_QT_subsidence_k = - ρ0_c[k] * aux_gm.subsidence[k] * ∇q_tot_gm[k]
```
ilopezgp, line 510: This should not have a density in front. You are multiplying this by another density below in the tendencies. Same for `gm_H_subsidence_k`.

driver/dycore.jl
```diff
@@ -483,9 +523,15 @@ function compute_gm_tendencies!(
                 end
             end
 
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * (H_horz_adv + gm_H_nudge_k + H_fluc + gm_H_subsidence_k)
             tendencies_gm.ρq_tot[k] +=
                 ρ0_c[k] * (aux_gm.dqtdt_hadv[k] + gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k] + gm_QT_subsidence_k)
+            tendencies_gm.ρe_tot[k] +=
+                ρ0_c[k] * (
```
ilopezgp, line 529: this is the density you are multiplying by terms already multiplied by density earlier.

driver/dycore.jl
```diff
@@ -483,9 +523,15 @@ function compute_gm_tendencies!(
...
+                    gm_H_subsidence_k +
+                    cv_m * Π * (H_horz_adv + gm_H_nudge_k + H_fluc) +
```
ilopezgp, line 531: why is this term multiplied by the Exner function? You already define it as a temperature tendency above. Please, change naming convention from `H_horz_adv` to `T_horz_adv` so that this is clear and these errors are avoided in the future.

driver/Cases.jl
yairchn: I had a discussion with @costachris about this a few days ago, we are just using the field and setting it to zero anyways in 
`field = TC.FieldFromNamedTuple(TC.face_space(grid), (; ρq_tot = FT(0)))`
This qt is not used here as surface fluxes are prescribed. it would be a problem in the future if we compute the fluxes 
this is what @charleskawczynski asked to change to avoid grid this morning.

driver/Cases.jl
yairchn: they are initialized in [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/3d3260f427a2c4fdf4d36e376948641aad602dbb/driver/dycore.jl#L270) to avoid code duplicity.

driver/dycore.jl
yairchn: I am just copying the field to have something `center` it is overwritten in the [next line](https://github.com/CliMA/TurbulenceConvection.jl/blob/3d3260f427a2c4fdf4d36e376948641aad602dbb/driver/dycore.jl#L251)

driver/dycore.jl
yairchn: I think that might be a rebase error, it should be `cv_m` like in L446. good catch

driver/dycore.jl
yairchn: I am collecting temperature tendencies from LES and converting to total energy [below](https://github.com/CliMA/TurbulenceConvection.jl/blob/3d3260f427a2c4fdf4d36e376948641aad602dbb/driver/dycore.jl#L531)

yairchn:
@tapio @ilopezgp 
regarding the comments above about the various `H` forcing in the LES_driven_SCM I changes the code to read the LES temperature [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/b6789928dfa2df1581d1f1d1e82b0250273f37d6/driver/Forcing.jl#L12)
and use temperature tendencies in the nudging consistently with fluctuates and others so that all are converted to energy in a clear manner.

driver/dycore.jl
yairchn: See below, I changes the code to have only temperature tendencies which is now clear how they are converted to energy tendencies

driver/Cases.jl
charleskawczynski: This is getting removed in #981.

src/microphysics_coupling.jl
```diff
@@ -71,6 +73,12 @@ function precipitation_formation(
         c_pm = TD.cp_m(param_set, ts)
         L_v0 = CPP.LH_v0(param_set)
         L_s0 = CPP.LH_s0(param_set)
+        I_l = TD.internal_energy_liquid(param_set, ts)
+        I_i = TD.internal_energy_ice(param_set, ts)
+        I = TD.internal_energy(param_set, ts)
+        #Φ = gravitational_potential(atmos.orientation, aux) #TODO how to use it here?
+        g = CPP.grav(param_set)
+        Φ = g * z
```
ilopezgp, line 81: Use the `Φ` function?

src/EDMF_functions.jl
```diff
@@ -97,12 +110,17 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
         q_tot_up = aux_up_i.q_tot
         @. aux_up_f[i].massflux = ρ0_f * Ifau(a_up) * (w_up - w_gm)
         # We know that, since W = 0 at z = 0, m = 0 also, and
-        # therefore θ_liq_ice / q_tot values do not matter
-        @. massflux_h += ρ0_f * (Ifau(a_up) * (w_up - w_gm) * (If(θ_liq_ice_up) - If(θ_liq_ice_gm)))
+        # therefore h_tot / q_tot values do not matter
+        ts_up_i = copy(ts_en)
+        h_tot_up_i = copy(q_tot_up)
```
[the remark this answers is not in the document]
charleskawczynski: This is fine for now, we can add variables to the cache once we work out the equations

src/EDMF_functions.jl
```diff
@@ -74,17 +77,27 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
+    Ic = CCO.InterpolateF2C()
+
+    # compute total enthalpies
+    ts_en = center_aux_environment(state).ts
+    ts_gm = center_aux_grid_mean(state).ts
+    @. h_tot_gm = prog_gm.ρe_tot / ρ0_c + TD.gas_constant_air(param_set, ts_gm) * aux_gm.T
+    @. h_tot_en = (
+        TD.total_energy(param_set, ts_en, 0.5 * (prog_gm.u^2 + prog_gm.v^2 + Ic(w_en)^2), grid.zc.z * g) +
+        TD.gas_constant_air(param_set, ts_en) * aux_en.T
+    )
```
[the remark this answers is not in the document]
tapios: +1

driver/dycore.jl
charleskawczynski: This is fine for now, we can clean up the copies in a follow-up PR.

src/EDMF_functions.jl
```diff
@@ -97,12 +110,17 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
...
         # We know that, since W = 0 at z = 0, m = 0 also, and
-        # therefore θ_liq_ice / q_tot values do not matter
-        @. massflux_h += ρ0_f * (Ifau(a_up) * (w_up - w_gm) * (If(θ_liq_ice_up) - If(θ_liq_ice_gm)))
+        # therefore h_tot / q_tot values do not matter
+        ts_up_i = copy(ts_en)
+        h_tot_up_i = copy(q_tot_up)
+        @. ts_up_i = thermo_state_pθq(param_set, p0_c, θ_liq_ice_up, q_tot_up)
+        @. h_tot_up_i = (
+            TD.total_energy(param_set, ts_up_i, 0.5 * (prog_gm.u^2 + prog_gm.v^2 + Ic(w_up)^2), grid.zc.z * g) +
+            TD.gas_constant_air(param_set, ts_up_i) * aux_up_i.T
```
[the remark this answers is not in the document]
yairchn: now its aux_gm.T

src/EDMF_functions.jl
```diff
@@ -333,29 +372,39 @@ end
...
+    cp = TD.cp_m(param_set, ts_gm[kc_surf])
+    UnPack.@unpack ustar, zLL, oblength, α0LL = surface_helper(surf, grid, state)
+
     surf.bflux > 0 || return FT(0)
     a_total = edmf.surface_area
     a_ = area_surface_bc(surf, edmf, i)
-    UnPack.@unpack ustar, zLL, oblength, α0LL = surface_helper(surf, grid, state)
-    ρθ_liq_ice_flux = surf.ρθ_liq_ice_flux
-    h_var = get_surface_variance(ρθ_liq_ice_flux * α0LL, ρθ_liq_ice_flux * α0LL, ustar, zLL, oblength)
+    θ_liq_ice_flux = surf.shf / cp # assuming no ql,qi flux
```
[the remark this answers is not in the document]
yairchn: I corrected the bug by dividing by `ρ` here

src/EDMF_functions.jl
```diff
@@ -333,29 +372,39 @@ end
...
     aux_gm = center_aux_grid_mean(state)
+    ρ0_c = center_ref_state(state).ρ0
     kc_surf = kc_surface(grid)
     surf.bflux > 0 || return aux_gm.q_tot[kc_surf]
     a_total = edmf.surface_area
     a_ = area_surface_bc(surf, edmf, i)
     UnPack.@unpack ustar, zLL, oblength, α0LL = surface_helper(surf, grid, state)
-    ρq_tot_flux = surf.ρq_tot_flux
-    qt_var = get_surface_variance(ρq_tot_flux * α0LL, ρq_tot_flux * α0LL, ustar, zLL, oblength)
+    q_tot_flux = surf.ρq_tot_flux
```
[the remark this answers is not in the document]
yairchn: I corrected the bug by dividing by `ρ` here

src/EDMF_functions.jl
```diff
@@ -74,17 +77,27 @@ function compute_sgs_flux!(edmf::EDMFModel, grid::Grid, state::State, surf::Surf
     a_en = aux_en.area
     w_en = aux_en_f.w
     w_gm = prog_gm_f.w
-    θ_liq_ice_en = aux_en.θ_liq_ice
-    θ_liq_ice_gm = aux_gm.θ_liq_ice
+    h_tot_en = copy(a_en)
+    h_tot_gm = copy(a_en)
```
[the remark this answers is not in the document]
yairchn: talk to the folks that designed CC

driver/dycore.jl
```diff
@@ -501,38 +552,47 @@ function compute_gm_tendencies!(
...
-
-    sgs_flux_θ_liq_ice = aux_gm_f.sgs_flux_θ_liq_ice
+    TC.compute_sgs_flux!(edmf, grid, state, surf, param_set)
+    sgs_flux_h_tot = aux_gm_f.sgs_flux_h_tot
     sgs_flux_q_tot = aux_gm_f.sgs_flux_q_tot
     sgs_flux_u = aux_gm_f.sgs_flux_u
     sgs_flux_v = aux_gm_f.sgs_flux_v
     # apply surface BC as SGS flux at lowest level
-    sgs_flux_θ_liq_ice[kf_surf] = surf.ρθ_liq_ice_flux
+    sgs_flux_h_tot[kf_surf] = surf.ρe_tot_flux
```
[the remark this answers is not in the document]
yairchn: the flux from the surface is the SGS flux at the lowest level (first face), at any other level that flux is given by the EDMF. I named it `sgs_flux_h_tot` to make sure it is clear that this is the value of our `sgs_flux_h_tot` at the lowest level and nothing else is missing here

driver/dycore.jl
```diff
@@ -257,6 +286,10 @@ function set_thermo_state!(state, grid, moisture_model, param_set)
             error("Something went wrong. The moisture_model options are equilibrium or nonequilibrium")
         end
         ts_gm[k] = TC.thermo_state_pθq(param_set, p0_c[k], aux_gm.θ_liq_ice[k], aux_gm.q_tot[k], thermo_args...)
+        e_kin = 0.5 * (prog_gm.u[k]^2 + prog_gm.v[k]^2 + w_c[k]^2)
+        e_pot = TC.Φ(grid.zc.z[k], param_set)
+        prog_gm.ρe_tot[k] = ρ0_c[k] * TD.total_energy(param_set, ts_gm[k], e_kin, e_pot)
+        prog_gm.ρq_tot[k] = ρ0_c[k] * aux_gm.q_tot[k]
```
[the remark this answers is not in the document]
yairchn: @charleskawczynski what's your take on that?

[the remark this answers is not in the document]
charleskawczynski: I agree with Ignacio, prognostic variables respond to tendencies, the only place where we are "allowed" to overwrite them is in the filters, and I wouldn't recommend overwriting them like this, because it then means that we're not actually solving the `ρe_tot` equations-- it becomes diagnostic to whatever the RHS depends on.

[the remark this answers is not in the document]
charleskawczynski: And the same goes for moisture.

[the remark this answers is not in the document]
yairchn: this function  [`set_thermo_state_pθq!`](https://github.com/CliMA/TurbulenceConvection.jl/blob/8a69b3b8a07c28070ead50770dd053bb0469dda7/driver/dycore.jl#L270) is called only in initialization [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/8a69b3b8a07c28070ead50770dd053bb0469dda7/driver/main.jl#L186)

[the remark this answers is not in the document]
charleskawczynski: Okay, let's define
```julia
function set_grid_mean_from_thermo_state!(...)
...
        e_kin = 0.5 * (prog_gm.u[k]^2 + prog_gm.v[k]^2 + w_c[k]^2)
        e_pot = TC.geopotential(param_set, grid.zc.z[k])
        prog_gm.ρe_tot[k] = ρ0_c[k] * TD.total_energy(param_set, ts_gm[k], e_kin, e_pot)
        prog_gm.ρq_tot[k] = ρ0_c[k] * aux_gm.q_tot[k]
...
end
```
And call this after calling `set_thermo_state_pθq!` in `main.jl`.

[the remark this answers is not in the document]
charleskawczynski: This way, `set_thermo_state_pθq!` keeps the accuracy of what it does (even if the way it does this is somehow different from the other `set_thermo_state!`s)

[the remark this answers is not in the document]
yairchn: done

src/thermodynamics.jl
```diff
@@ -9,3 +9,36 @@ function thermo_state_pθq(param_set::APS, p::FT, θ_liq_ice::FT, q_tot::FT, q_l
...
+    q = TD.PhasePartition(q_tot, q_liq, q_ice)
+    return TD.PhaseNonEquil_peq(param_set, p, e_int, q, config...)
+end
+
+function geopotential(param_set, z::FT) where {FT}
+    grav = CPP.grav(param_set)
+    return grav * z
+end
+
+function anelastic_total_enthalpy(param_set::APS, e_tot, ts) where {FT}
```
charleskawczynski, line 32:
```suggestion
function anelastic_total_enthalpy(param_set::APS, e_tot::FT, ts) where {FT}
```

src/thermodynamics.jl
```diff
@@ -9,3 +9,36 @@ function thermo_state_pθq(param_set::APS, p::FT, θ_liq_ice::FT, q_tot::FT, q_l
...
+end
+
+function thermo_state_peq(param_set::APS, p::FT, e_int::FT, q_tot::FT, q_liq::FT, q_ice::FT) where {FT}
+    config = ()
+    q = TD.PhasePartition(q_tot, q_liq, q_ice)
+    return TD.PhaseNonEquil_peq(param_set, p, e_int, q, config...)
+end
+
+function geopotential(param_set, z::FT) where {FT}
+    grav = CPP.grav(param_set)
```
charleskawczynski, line 28:
```suggestion
    grav = FT(CPP.grav(param_set))
```

src/thermodynamics.jl
```diff
@@ -9,3 +9,36 @@ function thermo_state_pθq(param_set::APS, p::FT, θ_liq_ice::FT, q_tot::FT, q_l
...
+    return grav * z
+end
+
+function anelastic_total_enthalpy(param_set::APS, e_tot, ts) where {FT}
+    Rm = TD.gas_constant_air(param_set, ts)
+    T = TD.air_temperature(param_set, ts)
+    return e_tot + Rm * T
+end
+
+function anelastic_total_enthalpy(param_set::APS, e_tot, p, ρ) where {FT}
```
charleskawczynski, line 38:
```suggestion
function anelastic_total_enthalpy(param_set::APS, e_tot::FT, p::FT, ρ::FT) where {FT}
```

src/thermodynamics.jl
```diff
@@ -9,3 +9,36 @@ function thermo_state_pθq(param_set::APS, p::FT, θ_liq_ice::FT, q_tot::FT, q_l
...
+    return e_tot + p / ρ
+end
+
+function kinetic_energy(u, v, w) where {FT}
```
charleskawczynski, line 42:
```suggestion
function kinetic_energy(u::FT, v::FT, w::FT) where {FT}
```

src/thermodynamics.jl
```diff
@@ -9,3 +9,36 @@ function thermo_state_pθq(param_set::APS, p::FT, θ_liq_ice::FT, q_tot::FT, q_l
...
+    return 0.5 * (u^2 + v^2 + w^2)
```
charleskawczynski, line 43:
```suggestion
    return FT(0.5) * (u^2 + v^2 + w^2)
```

driver/dycore.jl
```diff
@@ -501,38 +568,47 @@ function compute_gm_tendencies!(
...
             ρ0_c[k] * (
-                aux_bulk.θ_liq_ice_tendency_precip_formation[k] +
-                aux_en.θ_liq_ice_tendency_precip_formation[k] +
-                aux_tc.θ_liq_ice_tendency_precip_sinks[k]
+                aux_bulk.e_tot_tendency_precip_formation[k] +
+                aux_en.e_tot_tendency_precip_formation[k] +
+                aux_tc.e_tot_tendency_precip_sinks[k]
             )
+        #tendencies_gm.ρe_tot[k] +=
+        #    ρ0_c[k] * cp_m * Π * (
```
trontrytel, line 578: This can be deleted now?

src/EDMF_Environment.jl
```diff
@@ -48,10 +49,10 @@ function microphysics(
         end
 
         # update_env_precip_tendencies
-        # TODO: move qt_tendency_precip_formation and θ_liq_ice_tendency_precip_formation
-        # to diagnostics
+        # TODO: move ..._tendency_precip_formation to diagnostics
         aux_en.qt_tendency_precip_formation[k] = mph.qt_tendency * aux_en.area[k]
-        aux_en.θ_liq_ice_tendency_precip_formation[k] = mph.θ_liq_ice_tendency * aux_en.area[k]
+        #aux_en.θ_liq_ice_tendency_precip_formation[k] = mph.θ_liq_ice_tendency * aux_en.area[k]
```
trontrytel, line 54:
```suggestion
```

src/EDMF_Environment.jl
```diff
@@ -272,12 +276,13 @@ function microphysics(
...
             θ_liq_ice_tendency = outer_src.SH
+            e_tot_tendency = outer_src.Se_tot
             qr_tendency = outer_src.Sqr
             qs_tendency = outer_src.Sqs
-            # TODO: move qt_tendency_precip_formation and θ_liq_ice_tendency_precip_formation
-            # to diagnostics
+            # TODO: move ..._tendency_precip_formation to diagnostics
             aux_en.qt_tendency_precip_formation[k] = qt_tendency * aux_en.area[k]
-            aux_en.θ_liq_ice_tendency_precip_formation[k] = θ_liq_ice_tendency * aux_en.area[k]
+            #aux_en.θ_liq_ice_tendency_precip_formation[k] = θ_liq_ice_tendency * aux_en.area[k]
```
trontrytel, line 284:
```suggestion
```

src/EDMF_Environment.jl
```diff
@@ -325,15 +330,16 @@ function microphysics(
...
                 ts,
             )
 
             # update_env_precip_tendencies
-            # TODO: move qt_tendency_precip_formation and θ_liq_ice_tendency_precip_formation
-            # to diagnostics
+            # TODO: move ..._tendency_precip_formation to diagnostics
             aux_en.qt_tendency_precip_formation[k] = mph.qt_tendency * aux_en.area[k]
-            aux_en.θ_liq_ice_tendency_precip_formation[k] = mph.θ_liq_ice_tendency * aux_en.area[k]
+            #aux_en.θ_liq_ice_tendency_precip_formation[k] = mph.θ_liq_ice_tendency * aux_en.area[k]
```
trontrytel, line 341:
```suggestion
```

src/EDMF_Precipitation.jl
```diff
@@ -97,6 +97,13 @@ function compute_precipitation_sink_tendencies(
         L_s = TD.latent_heat_sublim(param_set, ts)
         L_f = TD.latent_heat_fusion(param_set, ts)
 
+        I_l = TD.internal_energy_liquid(param_set, ts)
+        I_i = TD.internal_energy_ice(param_set, ts)
+        I = TD.internal_energy(param_set, ts)
+        #Φ = gravitational_potential(atmos.orientation, aux) #TODO how to use it here?
```
trontrytel, line 103:
```suggestion
```

src/EDMF_Precipitation.jl
```diff
@@ -121,12 +128,15 @@ function compute_precipitation_sink_tendencies(
...
 
         aux_tc.qt_tendency_precip_sinks[k] = -S_qr_evap - S_qs_sub_dep
-        aux_tc.θ_liq_ice_tendency_precip_sinks[k] =
-            1 / Π_m / c_pm * (
-                S_qr_evap * (L_v - R_v * T_gm) * (1 + R_m / c_vm) +
-                S_qs_sub_dep * (L_s - R_v * T_gm) * (1 + R_m / c_vm) +
-                S_qs_melt * L_f * (1 + R_m / c_vm)
-            )
+
+        #aux_tc.θ_liq_ice_tendency_precip_sinks[k] =
```
trontrytel, line 132:
```suggestion
        #aux_tc.θ_liq_ice_tendency_precip_sinks[k] =
```

trontrytel, line 132: We can delete this

src/EDMF_Updrafts.jl
```diff
@@ -97,10 +99,12 @@ function compute_precipitation_formation_tendencies(
     end
     # TODO - to be deleted once we sum all tendencies elsewhere
     @inbounds for k in real_center_indices(grid)
-        aux_bulk.θ_liq_ice_tendency_precip_formation[k] = 0
+        #aux_bulk.θ_liq_ice_tendency_precip_formation[k] = 0
```
trontrytel, line 102:
```suggestion
```

src/EDMF_Updrafts.jl
```diff
@@ -97,10 +99,12 @@ function compute_precipitation_formation_tendencies(
...
+        aux_bulk.e_tot_tendency_precip_formation[k] = 0
         aux_bulk.qt_tendency_precip_formation[k] = 0
         @inbounds for i in 1:N_up
-            aux_bulk.θ_liq_ice_tendency_precip_formation[k] += aux_up[i].θ_liq_ice_tendency_precip_formation[k]
+            #aux_bulk.θ_liq_ice_tendency_precip_formation[k] += aux_up[i].θ_liq_ice_tendency_precip_formation[k]
```
trontrytel, line 106:
```suggestion
```

src/microphysics_coupling.jl
```diff
@@ -71,6 +73,12 @@ function precipitation_formation(
         c_pm = TD.cp_m(param_set, ts)
         L_v0 = CPP.LH_v0(param_set)
         L_s0 = CPP.LH_s0(param_set)
+        I_l = TD.internal_energy_liquid(param_set, ts)
+        I_i = TD.internal_energy_ice(param_set, ts)
+        I = TD.internal_energy(param_set, ts)
+        #Φ = gravitational_potential(atmos.orientation, aux) #TODO how to use it here?
```
trontrytel, line 79:
```suggestion
```

driver/dycore.jl
```diff
@@ -396,28 +452,31 @@ function compute_gm_tendencies!(
...
         if edmf.moisture_model isa TC.NonEquilibriumMoisture
             tendencies_gm.q_liq[k] -= ∇q_liq_gm[k] * aux_gm.subsidence[k]
             tendencies_gm.q_ice[k] -= ∇q_ice_gm[k] * aux_gm.subsidence[k]
         end
         # Radiation
         if TC.rad_type(radiation) <: Union{TC.RadiationDYCOMS_RF01, TC.RadiationLES, TC.RadiationTRMM_LBA}
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * aux_gm.dTdt_rad[k] / Π
+            tendencies_gm.ρe_tot[k] += ρ0_c[k] * cv_m * aux_gm.dTdt_rad[k]
         end
         # LS advection
```
yairchn, line 476: @tapios @ilopezgp 
here are the revised and re-orgenized LS advective tendencies after rebasing with #999 
I am wondering if `tendencies_gm.ρe_tot`in 479 is missing a Geopotential component to have a complete MSE as in the subsidence?

driver/dycore.jl
```diff
@@ -426,21 +485,22 @@ function compute_gm_tendencies!(
...
             gm_q_tot_nudge_k = Γᵣ * (aux_gm.qt_nudge[k] - aux_gm.q_tot[k])
             if edmf.moisture_model isa TC.NonEquilibriumMoisture
                 gm_q_liq_nudge_k = Γᵣ * (aux_gm.ql_nudge[k] - prog_gm.q_liq[k])
                 gm_q_ice_nudge_k = Γᵣ * (aux_gm.qi_nudge[k] - prog_gm.q_ice[k])
             end
 
-            tendencies_gm.ρθ_liq_ice[k] += ρ0_c[k] * (gm_H_nudge_k + H_fluc)
             tendencies_gm.ρq_tot[k] += ρ0_c[k] * (gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k])
+            tendencies_gm.ρe_tot[k] +=
+                ρ0_c[k] * (cv_m * (gm_T_nudge_k + T_fluc) + h_v * (gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k]))
```
yairchn, line 503: I assume that `gm_T_nudge_k` and  `T_fluc` are tendencies that should apply to energy not enthalpy and thus multiply `c_vm` here

driver/dycore.jl
```diff
@@ -426,21 +485,22 @@ function compute_gm_tendencies!(
...
         if TC.force_type(force) <: TC.ForcingLES
-            H_fluc = aux_gm.dTdt_fluc[k] / Π
+            T_fluc = aux_gm.dTdt_fluc[k]
 
             gm_U_nudge_k = (aux_gm.u_nudge[k] - prog_gm.u[k]) / force.nudge_tau
             gm_V_nudge_k = (aux_gm.v_nudge[k] - prog_gm.v[k]) / force.nudge_tau
 
             Γᵣ = TC.compute_les_Γᵣ(grid.zc[k])
-            gm_H_nudge_k = Γᵣ * (aux_gm.H_nudge[k] - aux_gm.θ_liq_ice[k])
+            gm_T_nudge_k = Γᵣ * (aux_gm.T_nudge[k] - aux_gm.θ_liq_ice[k])
```
ilopezgp, line 494: Shouldn't we be nudging the sensible temperature difference? Here you are nudging `aux_gm.θ_liq_ice[k]`.

driver/dycore.jl
tapios: Maybe this train has left the station (I don't see l. 479), and I'm not quite sure what this is referring to. But if it's explicit time derivatives you are adding, that would be zero for geopotential, wouldn't it?

driver/dycore.jl
tapios: Then it should also be internal energy for the vapor term, right?

driver/dycore.jl
yairchn: @tapios this question is still standing for me:
I meant the tendency in [479](https://github.com/CliMA/TurbulenceConvection.jl/blob/cb1c4c3fa5909a04484c916040f08dbf84294ef0/driver/dycore.jl#L479) that reads:
`tendencies_gm.ρe_tot[k] += ρ0_c[k] * (cp_m * aux_gm.dTdt_hadv[k] + h_v * aux_gm.dqtdt_hadv[k])`

Should there be some horizontal advection of Geopotential here? what can guide us is the expectation that the combined effect of horizontal advection and subsidence will balance out in the sensible heat budget in Rico. 
where subsidence tendency is given by:
 
` tendencies_gm.ρe_tot[k] -= ρ0_c[k] * aux_gm.subsidence[k] * ∇MSE_gm[k]`

driver/dycore.jl
yairchn: does the `h_v = cp_v * (aux_gm.T[k] - T_0) + Lv_0` multiplying the `q_t` tendencies ensures it is  internal energy for the vapor term? if not I need to understand what would provide this?

ilopezgp: From our discussion, this should be something like:

`h_v = cv_v * (aux_gm.T[k] - T_0) + Lv_0 - Rv*aux_gm.T[k] `

I think.

driver/dycore.jl
yairchn: in that case I think there should not be a Geopotential contribution here

src/EDMF_functions.jl
```diff
@@ -346,22 +400,34 @@ end
...
+    ts_gm = aux_gm.ts
+    c_p = TD.cp_m(param_set, ts_gm[kc_surf])
+    UnPack.@unpack ustar, zLL, oblength, ρLL = surface_helper(surf, grid, state)
+
     surf.bflux > 0 || return FT(0)
     a_total = edmf.surface_area
     a_ = area_surface_bc(surf, edmf, i)
-    UnPack.@unpack ustar, zLL, oblength, ρLL = surface_helper(surf, grid, state)
-    ρθ_liq_ice_flux = surf.ρθ_liq_ice_flux
+    ρθ_liq_ice_flux = surf.shf / c_p # assuming no ql,qi flux
```
charleskawczynski, line 422: Why are we computing this differently here?

yairchn: that is how `ρθ_liq_ice_flux ` was defined in surface fluxes in [main](https://github.com/CliMA/TurbulenceConvection.jl/blob/a4e6e8096ee444e93e5da64b45ad7b1bc73fbc34/driver/Surface.jl#L71)

charleskawczynski: So, can't we leave this as it was?

yairchn:
bots r-

PR #331, opened 2021-10-01
Remove unused and unnecessary code

charleskawczynski:

PR #961, opened 2022-05-01
Compartmentalize grid construction into standalone function.

ilopezgp:
This PR creates a standalone function for the construction of the SCM grid from the namelist. This allows users to construct the grid without having to run the model.

### DIFF
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -72,51 +72,7 @@ function Simulation1d(namelist)
     cfl_limit = namelist["time_stepping"]["cfl_limit"]
     dt_min = namelist["time_stepping"]["dt_min"]
 
-    truncated_gcm_mesh = TC.parse_namelist(namelist, "grid", "stretch", "flag"; default = false)
-
-    if Cases.get_case(namelist) == Cases.LES_driven_SCM()
-        Δz = get(namelist["grid"], "dz", nothing)
-        nz = get(namelist["grid"], "nz", nothing)
-        @assert isnothing(Δz) ⊻ isnothing(nz) string(
-            "LES_driven_SCM supports nz or Δz, not both.",
-            "The domain height is enforced to be the same as in LES.",
-        )
-
-        les_filename = namelist["meta"]["lesfile"]
-        zmax = NC.Dataset(les_filename, "r") do data
-            Array(TC.get_nc_data(data, "zc"))[end]
-        end
-        nz = isnothing(nz) ? Int(zmax ÷ Δz) : Int(nz)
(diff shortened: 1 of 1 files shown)


yairchn:
@ilopezgp would make sense for you if the function would be `construct_mesh` (returning `mesh = CC.Meshes.IntervalMesh(domain, nelems = nz)`) instead of `construct_grid` and the call to TC.grid would be called after it in `Simulation1d`? The reason would be to keep the separation of `src` better.

ilopezgp:
> @ilopezgp would make sense for you if the function would be `construct_mesh` (returning `mesh = CC.Meshes.IntervalMesh(domain, nelems = nz)`) instead of `construct_grid` and the call to TC.grid would be called after it in `Simulation1d`? The reason would be to keep the separation of `src` better.

We still would need to load TC src for `TC.parse_namelist`, `TC.get_nc_data` and `TC.TCMeshFromGCMMesh`, so we wouldn't really be saving any compilation or separating from source. In that sense, I think returning grid is a better solution. We can reconsider later if the code coupling is refactored.

yairchn:
@ilopezgp you see that it would create a dependency in CEDMF that will have to be handled once such changes are made in TC in the next months. And with such dependency on release versions it would create unneeded friction IMO.
@charleskawczynski what do you think?

ilopezgp:
> @ilopezgp you see that it would create a dependency in CEDMF that will have to be handled once such changes are made in TC in the next months. And with such dependency on release versions it would create unneeded friction IMO.
> @charleskawczynski what do you think?

I don't understand what unnecessary dependency you are referring to. We already need to import the entire main driver and the source code in CEDMF.

costachris (changes requested):
I didn't mean to request changes.

costachris (approved):
Changes LGTM. ~~This change will effectively prevent us from having to run the SCM for one iteration to generate the TC grid in CEDMF? or is this more related to calibrating with stretch grid?~~ Found my answer here. [CEDMF #331](https://github.com/CliMA/CalibrateEDMF.jl/pull/331)

charleskawczynski (approved):
I don’t think there’s any dependency issues. LGTM

yairchn (approved):

ilopezgp:
> Changes LGTM. ~This change will effectively prevent us from having to run the SCM for one iteration to generate the TC grid in CEDMF? or is this more related to calibrating with stretch grid?~ Found my answer here. [CEDMF #331](https://github.com/CliMA/CalibrateEDMF.jl/pull/331)

Both! I want to stop using the reference_SCM files in CEDMF, since they take up unnecessary compute/memory and users have to worry about deleting them every now and then.

PR #966, opened 2022-05-03
Ycaj/gm energy change var

trontrytel:
Dont merge, still have some questions.


src/EDMF_functions.jl
```diff
@@ -526,7 +526,7 @@ function compute_up_tendencies!(edmf::EDMFModel, grid::Grid, state::State, param
         θ_liq_ice_up = aux_up_i.θ_liq_ice
         entr_turb_dyn = aux_up_i.entr_turb_dyn
         detr_turb_dyn = aux_up_i.detr_turb_dyn
-        θ_liq_ice_tendency_precip_formation = aux_up_i.θ_liq_ice_tendency_precip_formation
+        e_tot_tendency_precip_formation = aux_up_i.e_tot_tendency_precip_formation
```
yairchn, line 529: just to make sure, in the `yc/gm_energy_change_var` branch we change only the grid mean to total energy, updraft and environment are still using `θ_liq_ice`. Is this tendency only entering the grid mean?

trontrytel: damn. ok forgot about that. This will make it more complicated for me. Should I just compute the tendencies twice for theta and e_tot?

PR #969, opened 2022-05-04
add e_tot sources from microphysics

trontrytel:
This is a draft of the changes with microphysics source terms. As discussed with @yairchn the theta tendencies from microphysics are used twice:
- in updraft equations
- in environment covariance equation

I don't do conversions between theta source terms and energy source terms. Instead I compute and save both of them. Once we are done with the transition we can clean this up, but for testing I thought I should keep it. 

TRMM runs with it. The results are not amazing, but they are also not horrible. (What a precise analysis from me)
contours
profiles


src/EDMF_Precipitation.jl
```diff
@@ -97,6 +97,13 @@ function compute_precipitation_sink_tendencies(
         L_s = TD.latent_heat_sublim(param_set, ts)
         L_f = TD.latent_heat_fusion(param_set, ts)
 
+        I_l = TD.internal_energy_liquid(param_set, ts)
+        I_i = TD.internal_energy_ice(param_set, ts)
+        I = TD.internal_energy(param_set, ts)
+        #Φ = gravitational_potential(atmos.orientation, aux) #TODO how to use it here?
+        g = CPP.grav(param_set)
+        Φ = g * grid.zc[k]
```
yairchn, line 105: is potential energy involved in microphysics energy src ? 🤯 
isn't that only true when there is a mass source from microphysics?

yairchn (approved):
left one question regarding the geopoetntial, otherwise looks good. I see there are some TODOs but we can merge it and keep working on the energy PR together - ok ?

src/EDMF_Precipitation.jl
trontrytel: Right. Good point. I'm not sure now. It's there because of the mass transfer between working fluid and precipitation. But as we have already discussed we are not actually changing the density in the anelastic approximation. So maybe we should not count that one either

trontrytel:
This was pulled into @yairchn branch with all change variables work

PR #973, opened 2022-05-05
Remove Pkg.develop in build

charleskawczynski:
This PR removes the develop step. We shouldn't need this since we're checking in the Manifest.

@yairchn, this should fix the issues you're having.


charleskawczynski:
Had to double check something: this will still test with the latest version--not the _release_ of TurbulenceConvection specified in the Manifest (v0.21.0). @simonbyrne pointed out that the `path` takes precedence (and there is no SHA entry).

PR #975, opened 2022-05-06
Truncate printing of ClimaCore field types

charleskawczynski:
Who wants shorter stack traces? 📒 This PR truncates printing for ClimaCore Field types (which people may know can be pretty long).

### DIFF
--- a/integration_tests/cli_options.jl
+++ b/integration_tests/cli_options.jl
@@ -41,6 +41,10 @@ function parse_commandline()
         arg_type = Int
         "--moisture_model" # Moisture model (equilibrium or non-equilibrium)
         arg_type = String
+        "--trunc_stack_traces"
+        help = "Set to `true` to truncate printing of ClimaCore `Field`s"
+        arg_type = Bool
+        default = true
     end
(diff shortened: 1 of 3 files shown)


haakon-e:
Cool! Looking forwards to using this. Do we have to re-implement https://github.com/CliMA/TurbulenceConvection.jl/pull/862 for that to show? (ref https://github.com/CliMA/TurbulenceConvection.jl/issues/868)

charleskawczynski:
> Cool! Looking forwards to using this.

Me too! 🚀

> Do we have to re-implement #862 for that to show? (ref #868)

No, I don't think so. This changes how to types themselves are printed-- Also, it's only applies to the top-level. So we'll want to implement this in CalibrateEDMF.jl, too if ya'll want it there.

I think we'd be type pirates if we changed how ClimaCore (CC) field type are printed inside `TC.jl/src`, because then it would change printing CC field types for anyone using the package, and that would be difficult to hunt down where / why that's happening.

ilopezgp (approved):
LGTM!

PR #978, opened 2022-05-08
Add error checking to CL overwrites

charleskawczynski:
This PR adds error checking to the CL arguments to prevent silently running unexpected configurations. This requires that we keep two lists up to date: `overwrite_namelist_map` (which we already had) and `no_overwrites` (now added). Before running, we now check that we've overwritten the correctly prescribed number of CL arguments.

This PR also attempts to fix the case that uses the nonequilibrium model, which we forgot to add to the `overwrite_namelist_map`.

~It seems like it's not yet working, but I made some changes that I'm not sure about.~ Looks like it's working!


trontrytel (approved):
Thank you for fixing everything @charleskawczynski !

Not sure where the `massflux_en` is coming from. Is it something we need for the variable change @yairchn ?

PR #992, opened 2022-05-13
add relaxation for deep convective LES_driven_SCM cases

costachris:
The GCM relaxation is different for deep convective cases. This PR specifies different relaxations in `Cases.jl` in a manner consistent with the LES runs.

### DIFF
--- a/driver/Cases.jl
+++ b/driver/Cases.jl
@@ -1107,7 +1107,25 @@ end
 ##### LES_driven_SCM
 #####
 
-forcing_kwargs(::LES_driven_SCM, namelist) = (; nudge_tau = namelist["forcing"]["nudging_timescale"])
+function forcing_kwargs(::LES_driven_SCM, namelist)
+    les_filename = namelist["meta"]["lesfile"]
+    cfsite_number, forcing_model, month, experiment = TC.parse_les_path(les_filename)
+    LES_library = TC.get_LES_library()
+    les_type = LES_library[forcing_model][string(month, pad = 2)]["cfsite_numbers"][string(cfsite_number, pad = 2)]
+    if les_type == "shallow"
+        wind_nudge_τᵣ = 6.0 * 3600.0
+        scalar_nudge_zᵢ = 3000.0
+        scalar_nudge_zᵣ = 3500.0
+        scalar_nudge_τᵣ = 24.0 * 3600.0
+    elseif les_type == "deep"
+        wind_nudge_τᵣ = 3600.0
+        scalar_nudge_zᵢ = 16000.0
+        scalar_nudge_zᵣ = 20000.0
+        scalar_nudge_τᵣ = 2.0 * 3600.0
+    end
+
+    return (; wind_nudge_τᵣ, scalar_nudge_zᵢ, scalar_nudge_zᵣ, scalar_nudge_τᵣ)
+end
 
 function surface_param_kwargs(::LES_driven_SCM, namelist)
     les_filename = namelist["meta"]["lesfile"]
@@ -1126,8 +1144,8 @@ function surface_param_kwargs(::LES_driven_SCM, namelist)
     return (; LESDat)
 end
 
-function ForcingBase(case::LES_driven_SCM, param_set::APS; nudge_tau)
-    ForcingBase(get_forcing_type(case); apply_coriolis = false, coriolis_param = 0.376e-4, nudge_tau = nudge_tau)
+function ForcingBase(case::LES_driven_SCM, param_set::APS; kwargs...)
+    ForcingBase(get_forcing_type(case); apply_coriolis = false, coriolis_param = 0.376e-4, kwargs...)
 end
 
 function surface_ref_state(::LES_driven_SCM, param_set::APS, namelist)
--- a/src/utility_functions.jl
+++ b/src/utility_functions.jl
@@ -70,3 +70,77 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
         end
     end
 end
+### Utility functions for LES_driven_SCM cases
+"""
+    get_LES_library
+Hierarchical dictionary of available LES simulations described in [Shen2022](@cite).
+The following cfsites are available across listed models, months,
+and experiments.
+"""
+function get_LES_library()
+    LES_library = Dict("HadGEM2-A" => Dict(), "CNRM-CM5" => Dict(), "CNRM-CM6-1" => Dict())
+    Shen_et_al_sites = collect(2:15)
+    append!(Shen_et_al_sites, collect(17:23))
+    Shen_et_al_deep_convection_sites = (collect(30:33)..., collect(66:70)..., 82, 92, 94, 96, 99, 100)
+    append!(Shen_et_al_sites, Shen_et_al_deep_convection_sites)
+
+    LES_library["HadGEM2-A"]["10"] = Dict()
+    LES_library["HadGEM2-A"]["10"]["cfsite_numbers"] = setdiff(Shen_et_al_sites, [94, 100])
+    LES_library["HadGEM2-A"]["07"] = Dict()
+    LES_library["HadGEM2-A"]["07"]["cfsite_numbers"] = Shen_et_al_sites
+    LES_library["HadGEM2-A"]["04"] = Dict()
+    LES_library["HadGEM2-A"]["04"]["cfsite_numbers"] = setdiff(Shen_et_al_sites, [15, 17, 18, 32, 92, 94])
+    LES_library["HadGEM2-A"]["01"] = Dict()
+    LES_library["HadGEM2-A"]["01"]["cfsite_numbers"] = setdiff(Shen_et_al_sites, [15, 17, 18, 19, 20])
+
+    for month in ["01", "04", "07", "10"]
+        LES_library["HadGEM2-A"][month]["experiments"] = ["amip", "amip4K"]
+    end
+
+    LES_library_full = deepcopy(LES_library)
+    for month in ["01", "04", "07", "10"]
+        LES_library_full["HadGEM2-A"][month]["cfsite_numbers"] = Dict()
+        for cfsite_number in LES_library["HadGEM2-A"][month]["cfsite_numbers"]
+            cfsite_number_str = string(cfsite_number, pad = 2)
+            LES_library_full["HadGEM2-A"][month]["cfsite_numbers"][cfsite_number_str] = if cfsite_number >= 30
+                "deep"
+            else
+                "shallow"
+            end
+        end
+    end
+    return LES_library_full
+end
+
+"""
+    parse_les_path(les_path)
+Given path to LES stats file, return cfsite_number, forcing_model, month, and experiment from filename.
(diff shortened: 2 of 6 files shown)


driver/Cases.jl
```diff
@@ -1130,7 +1130,28 @@ end
 ##### LES_driven_SCM
 #####
 
-forcing_kwargs(::LES_driven_SCM, namelist) = (; nudge_tau = namelist["forcing"]["nudging_timescale"])
+function forcing_kwargs(::LES_driven_SCM, namelist)
+    if namelist["les_type"] == "shallow"
```
yairchn, line 1134: can't we just move these values to the namelist  and read them from there? this way you could change them from `get_scm_config` in CEDMF if you ever need to.

costachris: I thought about doing that at first, but it leaves it up to users to remember which cases correspond to which relaxation parameters in the config, since we'll be calibrating a variety of cases. That seems a bit strange for users on the calibration side to deal with, and it's prone to bugs. Since `Cases.jl` is already a collection of case-specific settings/parameters, it made more sense to me to put it there.

ilopezgp: Along the same lines of worry as @costachris suggests. I think we should leave to room for error here, since we know a priori which LES cases were forced with which relaxation profiles.

I think what we need is a look-up function that assigns whether it was deep or shallow convection. We can do this with nested dictionaries, as we do in CalibrateEDMF to check whether a requested LES is "valid": see https://github.com/CliMA/CalibrateEDMF.jl/blob/main/src/LESUtils.jl#L19

yairchn (approved):

src/utility_functions.jl
```diff
@@ -70,3 +70,85 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+    get_LES_library
+Hierarchical dictionary of available LES simulations described in [Shen2022](@cite).
+The following cfsites are available across listed models, months,
+and experiments.
+"""
+function get_LES_library()
+    LES_library = Dict("HadGEM2-A" => Dict(), "CNRM-CM5" => Dict(), "CNRM-CM6-1" => Dict())
+    Shen_et_al_sites = collect(2:15)
+    append!(Shen_et_al_sites, collect(17:23))
+    Shen_et_al_deep_convection_sites = (collect(30:33)..., collect(66:70)..., 82, 92, 94, 96, 99, 100)
```
ilopezgp, line 84: From these new cases, I think only 30:33 (Indo-Pacific warm pool) and 66:70 (off Central America) represent deep convection. Cases 82, 92, 94, 96, 99, 100 are Southern Ocean cases. 

It would be great to clarify what the setup was when running cases 82, 92, 94, 96, 99, 100; and whether it was identical to the deep convective cases. We should get @szy21's OK on this before proceeding.

ilopezgp:
Screen Shot 2022-05-17 at 5 21 00 PM

Posting this here for future reference.

driver/Cases.jl
```diff
@@ -1129,8 +1152,23 @@ function surface_param_kwargs(::LES_driven_SCM, namelist)
...
+    ForcingBase(
+        get_forcing_type(case);
+        apply_coriolis = false,
+        coriolis_param = 0.376e-4,
+        wind_nudge_τᵣ = wind_nudge_τᵣ,
+        scalar_nudge_zᵢ = scalar_nudge_zᵢ,
+        scalar_nudge_zᵣ = scalar_nudge_zᵣ,
+        scalar_nudge_τᵣ = scalar_nudge_τᵣ,
+    )
 end
```
charleskawczynski, line 1172:
```suggestion
ForcingBase(
    case::LES_driven_SCM,
    param_set::APS;
    kwargs...,
)
    ForcingBase(
        get_forcing_type(case);
        apply_coriolis = false,
        coriolis_param = 0.376e-4,
        kwargs...,
    )
```
Can we get away with this instead?

src/utility_functions.jl
```diff
@@ -70,3 +70,85 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+Given path to LES stats file on central cluster, return cfsite_number, forcing_model, month, and experiment.
+Inputs:
+ - les_path - path to les simulation containing stats folder
+Outputs:
+ - cfsite_number  :: cfsite number
+ - forcing_model :: {"HadGEM2-A", "CNRM-CM5", "CNRM-CM6-1", "IPSL-CM6A-LR"} - name of climate model used for forcing. Currently, only "HadGEM2-A" simulations are available reliably.
+ - month :: {1, 4, 7, 10} - month of simulation.
+ - experiment :: {"amip", "amip4K"} - experiment from which LES was forced.
+"""
+
```
charleskawczynski, line 127:
```suggestion
```

driver/Cases.jl
```diff
@@ -1110,7 +1110,30 @@ end
...
+        scalar_nudge_zᵣ = 20000.0
+        scalar_nudge_τᵣ = 2.0 * 3600.0
+    end
+
+    return (;
+        wind_nudge_τᵣ = wind_nudge_τᵣ,
+        scalar_nudge_zᵢ = scalar_nudge_zᵢ,
+        scalar_nudge_zᵣ = scalar_nudge_zᵣ,
+        scalar_nudge_τᵣ = scalar_nudge_τᵣ,
+    )
```
charleskawczynski, line 1135:
```suggestion
    return (;
        wind_nudge_τᵣ,
        scalar_nudge_zᵢ,
        scalar_nudge_zᵣ,
        scalar_nudge_τᵣ,
    )
```

src/utility_functions.jl
```diff
@@ -70,3 +70,85 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+            @assert forcing_model in keys(LES_library)
+            @assert month in keys(LES_library[forcing_model])
+            @assert cfsite_number in keys(LES_library[forcing_model][month]["cfsite_numbers"])
+            @assert experiment in LES_library[forcing_model][month]["experiments"]
+        catch e
+            @error "The requested cfsite LES does not exist."
+            throw(e)
+        end
+    end
+end
```
charleskawczynski, line 154: I'd like to propose changes to these functions. But I need some time to think about what we can do to improve it.

src/utility_functions.jl
szy21: Yes, the setup for 82, 92, 94, 96, 99, 100 is identical to the one for 30-33 and 66-70.

src/utility_functions.jl
```diff
@@ -70,3 +70,84 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+Inputs:
+ - les_path - path to les simulation containing stats folder
+Outputs:
+ - cfsite_number  :: cfsite number
+ - forcing_model :: {"HadGEM2-A", "CNRM-CM5", "CNRM-CM6-1", "IPSL-CM6A-LR"} - name of climate model used for forcing. Currently, only "HadGEM2-A" simulations are available reliably.
+ - month :: {1, 4, 7, 10} - month of simulation.
+ - experiment :: {"amip", "amip4K"} - experiment from which LES was forced.
+"""
+function parse_les_path(les_path)
+    fname = split(les_path, "/")[end]
```
charleskawczynski, line 128: Can we use `splitext` instead

src/utility_functions.jl
```diff
@@ -70,3 +70,84 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+    fname_split = split(fname, ('.', '_'))
+    forcing_model = fname_split[3]
+    experiment = fname_split[4]
+    month = parse(Int64, fname_split[6])
+    cfsite_number = parse(Int64, replace(fname_split[2], "cfsite" => ""))
+    return (cfsite_number, forcing_model, month, experiment)
+end
+
+function valid_lespath(les_path)
+    if ~occursin(".julia/artifacts", les_path)
```
charleskawczynski, line 138: Why is this condition used?

src/utility_functions.jl
```diff
@@ -70,3 +70,84 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+        try
```
charleskawczynski, line 139:
```suggestion
        try
```
Can we please remove the try-catch?

src/utility_functions.jl
costachris: Originally to ensure any LES stats files used for tests still work, but as long as we follow the naming convention used in the les library it's fine (and we do for the LES_driven_SCM case we have now). I removed this and added a comment to `generate_namelist.jl` so it's clear.

src/utility_functions.jl
costachris: `splitext` didn't quite do what I want (it returned the file extension), but `basename` does and is cleaner, so I switched to that.

src/utility_functions.jl
```diff
@@ -70,3 +70,77 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+        LES_library_full["HadGEM2-A"][month]["cfsite_numbers"] = Dict()
+        for cfsite_number in LES_library["HadGEM2-A"][month]["cfsite_numbers"]
+            if cfsite_number >= 30
+                cfsite_number = string(cfsite_number, pad = 2)
+                LES_library_full["HadGEM2-A"][month]["cfsite_numbers"][cfsite_number] = "deep"
+            else
+                cfsite_number = string(cfsite_number, pad = 2)
+                LES_library_full["HadGEM2-A"][month]["cfsite_numbers"][cfsite_number] = "shallow"
+            end
+        end
```
charleskawczynski, line 111:
```suggestion
            cfsite_number = string(cfsite_number, pad = 2)
            LES_library_full["HadGEM2-A"][month]["cfsite_numbers"][cfsite_number] = if cfsite_number >= 30
                "deep"
            else
                "shallow"
            end
```

src/utility_functions.jl
```diff
@@ -70,3 +70,77 @@ function compare(a::F, b::F, pn0 = "") where {F <: CC.Fields.Field}
...
+    parse_les_path(les_path)
+Given path to LES stats file, return cfsite_number, forcing_model, month, and experiment from filename.
+Inputs:
+ - les_path - path to les simulation containing stats folder
+Outputs:
+ - cfsite_number  :: cfsite number
+ - forcing_model :: {"HadGEM2-A", "CNRM-CM5", "CNRM-CM6-1", "IPSL-CM6A-LR"} - name of climate model used for forcing. Currently, only "HadGEM2-A" simulations are available reliably.
+ - month :: {1, 4, 7, 10} - month of simulation.
+ - experiment :: {"amip", "amip4K"} - experiment from which LES was forced.
+"""
```
charleskawczynski, line 126:
```suggestion
 - experiment :: {"amip", "amip4K"} - experiment from which LES was forced.

# LES filename should follow pattern:
```
Stats.cfsite<SITE-NUMBER>_<FORCING-MODEL>_<EXPERIMENT>_2004-2008.<MONTH>.nc
```
"""
```

charleskawczynski (approved):

PR #996, opened 2022-05-15
organize the microphysics and terminal velocity computation with respect to update_aux and tendencies 

yairchn:


driver/dycore.jl
```diff
@@ -324,26 +324,57 @@ function ∑tendencies!(tendencies::FV, prog::FV, params::NT, t::Real) where {NT
...
+)
 
+    ρ0_c = TC.center_ref_state(state).ρ0
+    aux_tc = TC.center_aux_turbconv(state)
+    if precip_model isa TC.Clima1M
+        # helper to calculate the rain velocity
+        # TODO: assuming w_gm = 0
+        # TODO: verify translation
+        term_vel_rain = aux_tc.term_vel_rain
+        term_vel_snow = aux_tc.term_vel_snow
```
charleskawczynski, line 370: If these are TC-specific variables, then we should probably put these computations inside `update_aux!`. If they're grid-mean computations, then this separate function is fine but we should probably move the variables from the TC aux to the grid-mean aux.

yairchn: I think `term_vel_rain` and `term_vel_snow` are grid variables independent of TC.jl (you could run microphysics without EDMF). @trontrytel

trontrytel:
Should we close this for now?

PR #999, opened 2022-05-17
Separate large scale advective from radiative tendencies

yairchn:
The motivation for this PR is organizing and making a clear distinction between the various large scale tendencies in the various cases. Currently, several cases mix the application of advective and radiative tendencies. This has no effect when `θ_liq_ice` is the prognostic variable and both tendencies are applied in the same way. When `e_tot` is the prognostic variable however these two tendencies need to be applied differently, i.e. acting on total enthalpy (with `c_pm`) or on total energy (with `c_vm`).
* I removed `dTdt` and `dqtdt` in `aux_gm` as it is ambiguous and resorted to use `dTdt_hadv` and `dqtdt_hadv` instead. 
* The flag `apply subsidence` is not needed anymore as in cases where subsidence should not be applied the subsidence velocity is simply zero anyways.

### DIFF
--- a/driver/Radiation.jl
+++ b/driver/Radiation.jl
@@ -1,10 +1,10 @@
-update_radiation(self::TC.RadiationBase, grid, state, param_set) = nothing
+update_radiation(self::TC.RadiationBase, grid, state, t::Real, param_set) = nothing
 initialize(self::TC.RadiationBase{TC.RadiationNone}, grid, state) = nothing
 
 """
 see eq. 3 in Stevens et. al. 2005 DYCOMS paper
 """
-function update_radiation(self::TC.RadiationBase{TC.RadiationDYCOMS_RF01}, grid, state, param_set)
+function update_radiation(self::TC.RadiationBase{TC.RadiationDYCOMS_RF01}, grid, state, t::Real, param_set)
     cp_d = CPP.cp_d(param_set)
     ρ0_f = TC.face_ref_state(state).ρ0
     ρ0_c = TC.center_ref_state(state).ρ0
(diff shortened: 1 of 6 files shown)


driver/dycore.jl
```diff
@@ -396,105 +394,62 @@ function compute_gm_tendencies!(
         @. ∇q_ice_gm = ∇c(wvec(RBq(prog_gm.q_ice)))
     end
 
+    # Apply forcing and radiation
```
yairchn, line 397: @costachris can you double check that all changes here make sense ?

driver/Radiation.jl
```diff
@@ -83,3 +83,22 @@ function initialize(self::TC.RadiationBase{TC.RadiationLES}, grid, state, LESDat
     end
     return
 end
+
+
+function initialize(self::TC.RadiationBase{TC.RadiationTRMM_LBA}, grid, state)
+    aux_gm = TC.center_aux_grid_mean(state)
+    rad = APL.TRMM_LBA_radiation(Float64)
```
charleskawczynski, line 90:
```suggestion
    rad = APL.TRMM_LBA_radiation(eltype(grid))
```

driver/Radiation.jl
```diff
@@ -83,3 +83,22 @@ function initialize(self::TC.RadiationBase{TC.RadiationLES}, grid, state, LESDat
...
+    @inbounds for k in real_center_indices(grid)
+        aux_gm.dTdt_rad[k] = rad(0, grid.zc[k].z)
+    end
+    return nothing
+end
+
+function update_radiation(self::TC.RadiationBase{TC.RadiationTRMM_LBA}, grid, state, t::Real, param_set)
+    aux_gm = TC.center_aux_grid_mean(state)
+    rad = APL.TRMM_LBA_radiation(Float64)
```
charleskawczynski, line 99:
```suggestion
    rad = APL.TRMM_LBA_radiation(eltype(grid))
```

charleskawczynski (approved):
Code changes look good to me. Bravo for deleting more than adding! 🎉

driver/dycore.jl
```diff
@@ -396,105 +394,62 @@ function compute_gm_tendencies!(
...
-                tendencies_gm.q_ice[k] += aux_gm.dqidt[k]
-            end
+        if edmf.moisture_model isa TC.NonEquilibriumMoisture
+            tendencies_gm.q_liq[k] += aux_gm.dqldt[k]
+            tendencies_gm.q_ice[k] += aux_gm.dqidt[k]
         end
 
+        # LES specific forcings
         if TC.force_type(force) <: TC.ForcingLES
-            H_horz_adv = aux_gm.dTdt_hadv[k] / Π
```
costachris, line 441: I am wondering why removing `H_horz_adv` didn't change the results. Does this term only matter if you have subsidence ?

costachris (approved):

driver/dycore.jl
yairchn: if you follow [above](https://github.com/CliMA/TurbulenceConvection.jl/blob/e835545c38a67968cc520554114c3d3194ea9070/driver/dycore.jl#L417) I removed an `if` condition so that advection is applied to all cases there (including LES_driven_SCM)

costachris: got it

yairchn:
Bors r+

Issue #997, opened 2022-05-16
Internalize Clima parameters


charleskawczynski:
We should probably internalize TC's CLIMAParameters, like what we did in [Thermodynamics 91](https://github.com/CliMA/Thermodynamics.jl/commit/8dfd1a3cb730ff3bff831841564156e82c6ad13a) before doing any parameter boxing. This will help smooth the change to the new way we handle climaparameters.

This affects #974.

trontrytel:
I can take that on. Already did it in CloudMicrophysics.jl

trontrytel:
closed by #1000

PR #1000, opened 2022-05-17
Add internal clima parameters

trontrytel:
As suggested by Issue https://github.com/CliMA/TurbulenceConvection.jl/issues/997 this adds `InternalClimaParams.jl` as a way to interface with the default parameters of CliMA. The already existing `ClimaParameters.jl` file (with experimental params that are not yet part of CliMAParameters) is now renamed to `ExperimentalClimaParams.jl`.

HTH with moving parameters around!

### DIFF
--- a/docs/src/API.md
+++ b/docs/src/API.md
@@ -5,5 +5,6 @@ CurrentModule = TurbulenceConvection
 ```
 
 ```@docs
-TurbulenceConvection.ClimaParams
+TurbulenceConvection.ExperimentalClimaParams
+TurbulenceConvection.InternalClimaParams
 ```
--- a/src/EDMF_Precipitation.jl
+++ b/src/EDMF_Precipitation.jl
@@ -90,16 +90,16 @@ function compute_precipitation_sink_tendencies(
         c_pm = TD.cp_m(param_set, ts)
         c_vm = TD.cv_m(param_set, ts)
         R_m = TD.gas_constant_air(param_set, ts)
-        R_v = CPP.R_v(param_set)
-        L_v0 = CPP.LH_v0(param_set)
-        L_s0 = CPP.LH_s0(param_set)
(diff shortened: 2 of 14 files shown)


odunbar:
Just to be clear - this is something which will improve the **current** CLIMAParameters interface only, it is (I believe) incompatable with the new interface. 

So It would be good to have this on the parts of the code which are not currently being worked on (e.g. those parts-outside of EDMF) only - as there we have already effectively completed the transition.

trontrytel:
> Just to be clear - this is something which will improve the **current** CLIMAParameters interface only, it is (I believe) incompatable with the new interface.
> 
> So It would be good to have this on the parts of the code which are not currently being worked on (e.g. those parts-outside of EDMF) only - as there we have already effectively completed the transition.
> 
> My concern is that the parameter boxes are built hierarchically currently.

Did not mean to step on someone elses PR. I thought, as suggested by @charleskawczynski, that this makes the move easier because it limits the needed changes to one file only. 

I thought that `InternalClimaParams.jl` file can still be populated based on the parameter boxes and the structure you are introducing? But feel free to close the PR if it's going in the wrong direction.

charleskawczynski:
@odunbar and I chatted. This PR is compatible, but we'll also need to add a few methods for hiding the structure of the outer-most parameter boxes. Something like:

```julia
thermo_params(ps::EarthParameterSet) = ps # which will become ps.thermo_params
microphys_params(ps::EarthParameterSet) = ps # which will become ps.microphys_params
surf_fluxes_params(ps::EarthParameterSet) = ps # which will become ps.surf_fluxes_params
```

We can do the methods in a separate PR.

charleskawczynski (approved):

trontrytel:
Sounds good. It would be sad not to merge a PR with such a momentous number. 

I'll rebase and merge this then ;)

PR #1001, opened 2022-05-17
update to CloudMicrophysics v0.5.0

trontrytel:
I think compat is lying here https://github.com/CliMA/TurbulenceConvection.jl/pull/998 and we need the below changes to update to `CloudMicrophysics.jl v0.5.0`

Is compat lying because we would have to clean up the depot for the `CM.jl` changes to show up?


charleskawczynski (commented):
Yeah, I think CompatHelper tends to be optimistic with suggestions. This looks fine

charleskawczynski (approved):

PR #1004, opened 2022-05-17
Remove reference state and move ref variables to grid mean

yairchn:
This RP removes the reference state from the model and moves the references variables to the grid mean.
* `p` is added to `aux_gm` and `aux_gm_f`
* `ρ` is added to `prog_gm` and `aux_gm_f`
* `α` is removed completely from the model (use `1/ρ` when instead). 
The grid mean `p`, `ρ` are initialized as the references state by assuming an hydrostatic balance and their tendencies are zero. 
This prepares the ground to having both Anelastic (constant `ρ`) and compressible (time dependent `ρ`) options when coupling to a global model.


docs/src/PlotReferenceStates.jl
```diff
@@ -24,7 +24,7 @@ function export_ref_profile(case_name::String)
     case = Cases.get_case(namelist)
     surf_ref_state = Cases.surface_ref_state(case, param_set, namelist)
 
-    aux_vars(FT) = (; ref_state = (ρ0 = FT(0), α0 = FT(0), p0 = FT(0)))
+    aux_vars(FT) = (; ref_state = (ρ = FT(0), p = FT(0)))
```
yairchn, line 27: @charleskawczynski can you suggest code changes here that follow the changes below ?

charleskawczynski: I need to take a closer look, this script basically mocks the reference state for plotting, so it all needs to match what we do in the real case

charleskawczynski: I think the simplest thing here would be to provide two ways to call `compute_ref_state!`. I'll open a PR to make the changes. Then this call can be the one that works with the raw data and this will get simplified.

charleskawczynski: Opened #1008.

charleskawczynski (approved):

charleskawczynski:
Let's squash before merging

yairchn:
Bors r+

PR #1005, opened 2022-05-17
Verbose stack traces by default

### LINKED ISSUE #1003: Optionally truncate stack traces
It looks like we need a more fine-tuned solution than what was applied in #906. For example, [this build](https://buildkite.com/clima/turbulenceconvection-ci/builds/10245#d3d55529-f153-410b-910e-758f1aa3c408/133-168) is only saying:

```
Error: TurbulenceConvection simulation crashed. UndefVarError(:prog_gm)
```

I think that the right thing to do here is to add
```julia
        @error "TurbulenceConvection simulation crashed. Stacktrace for failed simulation" exception =
            (e, catch_backtrace())
```
back in by default but add a namelist arg to switch to
```julia
@error "TurbulenceConvection simulation crashed. $(e)"
```
for Calibration

charleskawczynski:
Closes #1003.

This means that the calibration pipeline will likely want to set `truncate_stack_trace = false` when calling `default_namelist`.


haakon-e (approved):
LGTM, thanks!

PR #1008, opened 2022-05-18
Provide two interfaces to compute ref state

charleskawczynski:
This PR will allow us to compute the reference state given several individual fields, or the TC state.

This will help make #1004 easier.

### DIFF
--- a/docs/src/PlotReferenceStates.jl
+++ b/docs/src/PlotReferenceStates.jl
@@ -23,24 +23,28 @@ function export_ref_profile(case_name::String)
     grid = TC.Grid(FT(namelist["grid"]["dz"]), namelist["grid"]["nz"])
     case = Cases.get_case(namelist)
     surf_ref_state = Cases.surface_ref_state(case, param_set, namelist)
+    ts_g = surf_ref_state
 
-    aux_vars(FT) = (; ref_state = (ρ0 = FT(0), α0 = FT(0), p0 = FT(0)))
-    aux_cent_fields = TC.FieldFromNamedTuple(TC.center_space(grid), aux_vars(FT))
-    aux_face_fields = TC.FieldFromNamedTuple(TC.face_space(grid), aux_vars(FT))
+    aux_vars(FT) = (; ρ0 = FT(0), α0 = FT(0), p0 = FT(0))
+    cent = TC.FieldFromNamedTuple(TC.center_space(grid), aux_vars(FT))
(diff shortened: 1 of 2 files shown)


yairchn:
but #1004 deletes `center_ref_state`, and `face_ref_state`

charleskawczynski:
> but #1004 deletes `center_ref_state`, and `face_ref_state`

That's fine, this makes it so that we don't need `state` in the ref state plotting. We're still computing the reference state in #1004.

driver/dycore.jl
```diff
@@ -183,14 +183,28 @@ The reference profiles, given
...
     p0_c = TC.center_ref_state(state).p0
     ρ0_c = TC.center_ref_state(state).ρ0
     α0_c = TC.center_ref_state(state).α0
     p0_f = TC.face_ref_state(state).p0
     ρ0_f = TC.face_ref_state(state).ρ0
     α0_f = TC.face_ref_state(state).α0
+    compute_ref_state!(p0_c, ρ0_c, α0_c, p0_f, ρ0_f, α0_f, grid, param_set; ts_g)
+end
+
+function compute_ref_state!(
```
yairchn, line 195: we should remove all `0` from the names here and remove `α` as #1004 replace it with `1/ρ` anyways

docs/src/PlotReferenceStates.jl
```diff
@@ -23,24 +23,28 @@ function export_ref_profile(case_name::String)
     grid = TC.Grid(FT(namelist["grid"]["dz"]), namelist["grid"]["nz"])
     case = Cases.get_case(namelist)
     surf_ref_state = Cases.surface_ref_state(case, param_set, namelist)
+    ts_g = surf_ref_state
 
-    aux_vars(FT) = (; ref_state = (ρ0 = FT(0), α0 = FT(0), p0 = FT(0)))
-    aux_cent_fields = TC.FieldFromNamedTuple(TC.center_space(grid), aux_vars(FT))
-    aux_face_fields = TC.FieldFromNamedTuple(TC.face_space(grid), aux_vars(FT))
+    aux_vars(FT) = (; ρ0 = FT(0), α0 = FT(0), p0 = FT(0))
```
yairchn, line 28: we should remove all 0 from the names here and remove α as https://github.com/CliMA/TurbulenceConvection.jl/pull/1004 replace it with 1/ρ anyways

yairchn (approved):

PR #1023, opened 2022-05-20
Add prescribed and diagnostic precip fraction

trontrytel:
This PR adds a precipitation fraction model. The supported options are:
 - prescribed (done via experimental parameter, defaults to precipitation_fraction = 1, which is what we have now)
 - diagnostic based on the max cloud fraction (it has an experimental parameter that limits the precip fraction to to be smaller than the limiter value)
 
The basic idea is to divide the precipitation variables by the precip fraction before computing any source terms with them. And then multiply by precip fraction at the end to recover back the grid mean tendency. The non-linearities in the microphysics source terms should make a difference compared to precipitation fraction = 1

Edit: After some discussions with @yairchn and @ilopezgp we decided to apply this only to source terms. We need to think more on how to handle advection term and if we want to include those non-linear effects there.

### DIFF
--- a/src/EDMF_Precipitation.jl
+++ b/src/EDMF_Precipitation.jl
@@ -1,4 +1,23 @@
 """
+    compute_precip_fraction
+
+Computes diagnostic precipitation fraction
+"""
+function compute_precip_fraction(edmf::EDMFModel, state::State, param_set::APS)
+
+    pf = if edmf.precip_fraction_model isa PrescribedPrecipFraction
+        ECP.prescribed_precip_frac_value(param_set)
+    elseif edmf.precip_fraction_model isa DiagnosticPrecipFraction
+        aux_gm = center_aux_grid_mean(state)
+        maxcf = maximum(aux_gm.cloud_fraction)
+        max(maxcf, ECP.precip_fraction_limiter(param_set))
+    else
+        error("Failed to compute precipitation fraction.")
+    end
+    return pf
+end
+
+"""
 Computes the rain and snow advection (down) tendency
 """
 compute_precipitation_advection_tendencies(
@@ -30,8 +49,10 @@ function compute_precipitation_advection_tendencies(
     term_vel_rain = aux_tc.term_vel_rain
     term_vel_snow = aux_tc.term_vel_snow
 
-    q_rai = prog_pr.q_rai
-    q_sno = prog_pr.q_sno
+    precip_fraction = compute_precip_fraction(edmf, state, param_set)
+
+    q_rai = prog_pr.q_rai #./ precip_fraction
+    q_sno = prog_pr.q_sno #./ precip_fraction
 
     If = CCO.DivergenceF2C()
     RB = CCO.RightBiasedC2F(; top = CCO.SetValue(FT(0)))
@@ -39,8 +60,8 @@ function compute_precipitation_advection_tendencies(
     wvec = CC.Geometry.WVector
 
     # TODO - some positivity limiters are needed
-    @. aux_tc.qr_tendency_advection = ∇(wvec(RB(ρ_c * q_rai * term_vel_rain))) / ρ_c
-    @. aux_tc.qs_tendency_advection = ∇(wvec(RB(ρ_c * q_sno * term_vel_snow))) / ρ_c
+    @. aux_tc.qr_tendency_advection = ∇(wvec(RB(ρ_c * q_rai * term_vel_rain))) / ρ_c# * precip_fraction
+    @. aux_tc.qs_tendency_advection = ∇(wvec(RB(ρ_c * q_sno * term_vel_snow))) / ρ_c# * precip_fraction
 
     @. tendencies_pr.q_rai += aux_tc.qr_tendency_advection
     @. tendencies_pr.q_sno += aux_tc.qs_tendency_advection
@@ -76,9 +97,11 @@ function compute_precipitation_sink_tendencies(
     tendencies_pr = center_tendencies_precipitation(state)
     ts_gm = aux_gm.ts
 
+    precip_fraction = compute_precip_fraction(edmf, state, param_set)
+
     @inbounds for k in real_center_indices(grid)
-        qr = prog_pr.q_rai[k]
-        qs = prog_pr.q_sno[k]
+        qr = prog_pr.q_rai[k] / precip_fraction
+        qs = prog_pr.q_sno[k] / precip_fraction
         ρ = ρ_c[k]
         q_tot_gm = aux_gm.q_tot[k]
         T_gm = aux_gm.T[k]
@@ -105,9 +128,10 @@ function compute_precipitation_sink_tendencies(
         # TODO - move limiters elsewhere
         # TODO - when using adaptive timestepping we are limiting the source terms
         #        with the previous timestep dt
-        S_qr_evap = -min(qr / Δt, -α_evp * CM1.evaporation_sublimation(param_set, rain_type, q, qr, ρ, T_gm))
-        S_qs_melt = -min(qs / Δt, α_melt * CM1.snow_melt(param_set, qs, ρ, T_gm))
-        tmp = α_dep_sub * CM1.evaporation_sublimation(param_set, snow_type, q, qs, ρ, T_gm)
+        S_qr_evap =
+            -min(qr / Δt, -α_evp * CM1.evaporation_sublimation(param_set, rain_type, q, qr, ρ, T_gm)) * precip_fraction
+        S_qs_melt = -min(qs / Δt, α_melt * CM1.snow_melt(param_set, qs, ρ, T_gm)) * precip_fraction
+        tmp = α_dep_sub * CM1.evaporation_sublimation(param_set, snow_type, q, qs, ρ, T_gm) * precip_fraction
         if tmp > 0
             S_qs_sub_dep = min(qv / Δt, tmp)
         else
--- a/src/microphysics_coupling.jl
+++ b/src/microphysics_coupling.jl
@@ -52,6 +52,7 @@ function precipitation_formation(
     ρ::FT,
     Δt::Real,
     ts,
+    precip_fraction,
 ) where {FT}
 
     # TODO - when using adaptive timestepping we are limiting the source terms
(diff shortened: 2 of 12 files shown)


src/EDMF_Precipitation.jl
```diff
@@ -1,3 +1,35 @@
...
+
+    if edmf.precip_fraction_prescribed
+        pf = ECP.prescribed_precip_fraction_value(param_set)
+    end
+    if edmf.precip_fraction_from_cloud_fraction
+        aux_bulk = center_aux_bulk(state)
+        aux_en = center_aux_environment(state)
+
+        pf = maximum(aux_bulk.cloud_fraction .* aux_bulk.area .+ aux_en.cloud_fraction .* aux_en.area)
+        pf = max(pf, 0.1)
```
ilopezgp, line 28: Can you make this `0.1` a kwarg or add it to the TC.jl parameter set?

src/EDMF_Precipitation.jl
```diff
@@ -76,6 +110,8 @@ function compute_precipitation_sink_tendencies(
     tendencies_pr = center_tendencies_precipitation(state)
     ts_gm = aux_gm.ts
 
+    precip_fraction = compute_precip_fraction(precip_model, edmf, state, param_set)
+
     @inbounds for k in real_center_indices(grid)
         qr = prog_pr.q_rai[k]
         qs = prog_pr.q_sno[k]
```
ilopezgp, line 117: Shouldn't these be divided by the precip fraction?

src/microphysics_coupling.jl
```diff
@@ -149,9 +153,15 @@ function precipitation_formation(
 
             # accretion rain - snow
             if T < T_fr
-                S_qs = min(qr / Δt, CM1.accretion_snow_rain(param_set, snow_type, rain_type, qs, qr, ρ))
+                S_qs =
+                    min(qr / Δt, CM1.accretion_snow_rain(param_set, snow_type, rain_type, qs, qr, ρ)) *
+                    precip_fraction *
+                    precip_fraction
```
ilopezgp, line 159: why squared?

src/types.jl
```diff
@@ -188,6 +187,11 @@ struct NoPrecipitation <: AbstractPrecipitationModel end
 struct Clima0M <: AbstractPrecipitationModel end
 struct Clima1M <: AbstractPrecipitationModel end
 
+abstract type AbstractPrecipFractionModel end
+struct NoPrecipFraction <: AbstractPrecipFractionModel end
+struct PrescribedPrecipFraction <: AbstractPrecipFractionModel end
```
ilopezgp, line 192: Can mer merge these into 1? Isn't NoPrecipFraction the same as PrescribedPrecipFraction with value of 1?

src/EDMF_Precipitation.jl
trontrytel: Thank you for catching it. Yes. Must have disappeared when I was rebasing

src/EDMF_Precipitation.jl
trontrytel: Will do

src/microphysics_coupling.jl
trontrytel: Because I have both  `q_snow` and `q_rain` entering the source term. In the previous sources it was always either or.

src/types.jl
trontrytel: Sounds good

src/EDMF_Precipitation.jl
```diff
@@ -1,3 +1,24 @@
...
+    pf::FT = FT(1.0)
+
+    if edmf.precip_fraction_model isa PrescribedPrecipFraction
+        pf = ECP.prescribed_precip_fraction_value(param_set)
+    end
+    if edmf.precip_fraction_model isa DiagnosticPrecip_fraction
+        aux_bulk = center_aux_bulk(state)
+        aux_en = center_aux_environment(state)
+
+        pf = maximum(aux_bulk.cloud_fraction .* aux_bulk.area .+ aux_en.cloud_fraction .* aux_en.area)
```
yairchn, line 16: the produce of bulk upd area and cloud fraction is almost upd_area^2 . is this what you meant ? 
I was suggesting the cloud fraction alone

src/EDMF_Precipitation.jl
```diff
@@ -30,17 +51,19 @@ function compute_precipitation_advection_tendencies(
...
 
     If = CCO.DivergenceF2C()
     RB = CCO.RightBiasedC2F(; top = CCO.SetValue(FT(0)))
     ∇ = CCO.DivergenceF2C(; bottom = CCO.Extrapolate())
     wvec = CC.Geometry.WVector
 
     # TODO - some positivity limiters are needed
-    @. aux_tc.qr_tendency_advection = ∇(wvec(RB(ρ_c * q_rai * term_vel_rain))) / ρ_c
-    @. aux_tc.qs_tendency_advection = ∇(wvec(RB(ρ_c * q_sno * term_vel_snow))) / ρ_c
+    @. aux_tc.qr_tendency_advection = ∇(wvec(RB(ρ_c * q_rai * term_vel_rain))) / ρ_c * precip_fraction
```
yairchn, line 65: you might wanna change the variable to be rhoaqr with a as the precip frac and solve conservative equation

src/EDMF_Precipitation.jl
trontrytel: Right. But I wanted to compute the grid mean cloud fraction. So I thought I should take a subdomain cloud fraction, multiply by the subdomain area and sum?

src/EDMF_Precipitation.jl
trontrytel: Should we just go for conservative form and switch to having `rho * precip_fraction * q_rai` as our precipitation state variables?

trontrytel:
This is now ready for review. The final verdict will come after the calibrations. But what I like is that the rain fraction PR does not change anything in DYCOMS_RF02 while increasing precipitation in Rico:

[pf_Dycoms_profiles.pdf](https://github.com/CliMA/TurbulenceConvection.jl/files/8756699/pf_Dycoms_profiles.pdf)
[pf_Rico_profiles.pdf](https://github.com/CliMA/TurbulenceConvection.jl/files/8756702/pf_Rico_profiles.pdf)
[pf_TRMM_profiles.pdf](https://github.com/CliMA/TurbulenceConvection.jl/files/8756707/pf_TRMM_profiles.pdf)

src/EDMF_Precipitation.jl
```diff
@@ -1,3 +1,24 @@
...
+        pf = ECP.prescribed_precip_frac_value(param_set)
+    end
+    if edmf.precip_fraction_model isa DiagnosticPrecipFraction
+
+        aux_gm = center_aux_grid_mean(state)
+
+        pf = maximum(aux_gm.cloud_fraction)
+        pf = max(pf, ECP.precip_fraction_limiter(param_set))
+    end
+    return pf
```
charleskawczynski, line 19:
```suggestion
    pf = if edmf.precip_fraction_model isa PrescribedPrecipFraction
        ECP.prescribed_precip_frac_value(param_set)
    elseif edmf.precip_fraction_model isa DiagnosticPrecipFraction
        aux_gm = center_aux_grid_mean(state)
        maxcf = maximum(aux_gm.cloud_fraction)
        max(maxcf, ECP.precip_fraction_limiter(param_set))
    else
        FT = eltype(edmf)
        FT(1)
    end
    return pf
```

Sorry, I should have recommended this the first time

src/EDMF_Precipitation.jl
```diff
@@ -1,3 +1,24 @@
+"""
+Computes diagnostic precipitation fraction
```
charleskawczynski, line 2:
```suggestion
    compute_precip_fraction

Computes diagnostic precipitation fraction
```

src/types.jl
```diff
@@ -504,6 +508,17 @@ struct EDMFModel{N_up, FT, MM, TCM, PM, ENT, EBGC, EC, EDS, EPG}
...
         precip_model = precip_model
+
+        precip_fraction_model_name =
+            parse_namelist(namelist, "microphysics", "precip_fraction_model"; default = "prescribed")
+        precip_fraction_model = if precip_fraction_model_name == "prescribed"
+            PrescribedPrecipFraction()
+        elseif precip_fraction_model_name == "cloud_cover"
+            DiagnosticPrecipFraction()
+        else
+            error("Something went wrong. Invalid precip_fraction model: '$precip_fraction_model_name'")
```
charleskawczynski, line 519:
```suggestion
            error("Something went wrong. Invalid `precip_fraction` model: `$precip_fraction_model_name`")
```

charleskawczynski (approved):
I left a few suggested changes. Code changes look fine to me overall.

src/EDMF_Precipitation.jl
trontrytel: We now only have `edmf.precip_fraction_model` be `PrescribedPrecipFraction` or `DiagnosticPrecipFraction
`. So I don't think there should be the last `else` here

charleskawczynski:
> But what I like is that the rain fraction PR does not change anything in DYCOMS_RF02 while increasing precipitation in Rico

Should we add these variables/plots to buildkite?

src/EDMF_Precipitation.jl
```diff
@@ -30,17 +49,19 @@ function compute_precipitation_advection_tendencies(
...
     If = CCO.DivergenceF2C()
     RB = CCO.RightBiasedC2F(; top = CCO.SetValue(FT(0)))
     ∇ = CCO.DivergenceF2C(; bottom = CCO.Extrapolate())
     wvec = CC.Geometry.WVector
 
     # TODO - some positivity limiters are needed
-    @. aux_tc.qr_tendency_advection = ∇(wvec(RB(ρ_c * q_rai * term_vel_rain))) / ρ_c
-    @. aux_tc.qs_tendency_advection = ∇(wvec(RB(ρ_c * q_sno * term_vel_snow))) / ρ_c
+    @. aux_tc.qr_tendency_advection = ∇(wvec(RB(ρ_c * q_rai * term_vel_rain))) / ρ_c * precip_fraction
+    @. aux_tc.qs_tendency_advection = ∇(wvec(RB(ρ_c * q_sno * term_vel_snow))) / ρ_c * precip_fraction
```
yairchn, line 64: did you mean to multiply or divide here ?

trontrytel: Multiply. They are divided in lines 54-55. So I'm multiplying back to get the grid mean advection tendency

ilopezgp (approved):
LGTM

yairchn (approved):

PR #1027, opened 2022-05-23
fix advection of rain and snow

yairchn:


yairchn:
this formulation is wrong, the correct one is what we are in main

PR #1032, opened 2022-05-24
Fix updraft theta_li plot

ilopezgp:


charleskawczynski:
We're not planning to merge this, correct? @ilopezgp

ilopezgp:
Not before the energy branch, I just want to fix the plots at some point

charleskawczynski:
> Not before the energy branch, I just want to fix the plots at some point

I think it'd be better to fix the data than to apply fixes in the plots. It's far less confusing to plot the data as it is, and not apply filters (like `data_les_cont_mapped[upd_area_les_cont_mapped .< eps] .= NaN`) in the plotting script

charleskawczynski:
Also, I'd like to split the regression tests and plotting infra, it'll make things much easier

PR #1043, opened 2022-05-25
Update dependencies

charleskawczynski:


charleskawczynski:
This seems to downgrade OperatorFlux.jl. @zongyi-li, @bischtob, can you please look into widening OperatorFlux's compat entries so that `julia .dev/up_deps.jl` in TC results in OperatorFlux 0.1.2? This doesn't mean that TC is incompatible with OperatorFlux 0.1.2, just that the resolve picks this in favor of upgrading other packages (some of which include fixes that remove precompilation warnings).

PR #1048, opened 2022-05-25
add SCT1 NN calibration to default namelist

costachris:
Add SCT1 NN calibration to default namelist and update buildkite to be consistent with the calibration.


ilopezgp (approved):
LGTM

charleskawczynski (approved):

costachris:
The builds for the NN/ NN_nonlocal look good, but some of the vanilla cases changed so I'm trying to figure out why

PR #1056, opened 2022-05-26
Fix non-equilibroum buoyancy calculations

trontrytel:
This reverts https://github.com/CliMA/TurbulenceConvection.jl/pull/1018

Right now we are inconsistent and taking `p[k]` for equilibrium `ts` but `p[k-1]` for non-equilibrium `ts`. 

Also, taking `p[k-1]` results in no updrafts for shallow convection cases and delays TRMM convection by around an hour.  (Compare the buildkite artifacts from before and after merging https://github.com/CliMA/TurbulenceConvection.jl/pull/1018 when looking at TRMM with non-equilibrium moisture model)

### DIFF
--- a/src/update_aux.jl
+++ b/src/update_aux.jl
@@ -192,7 +192,7 @@ function update_aux!(edmf::EDMFModel, grid::Grid, state::State, surf::SurfaceBas
                 elseif edmf.moisture_model isa NonEquilibriumMoisture
                     ql = aux_up[i].q_liq[k - 1]
                     qi = aux_up[i].q_ice[k - 1]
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
I'm fine with merging, but we really need to fix these buoyancy hacks. This PR may fix the consistency with the other moisture models, but it makes the call to `thermo_state_pθq` using inconsistent data.

trontrytel:
Yeah. We just talked about it with @yairchn . I think we really should revisit how the buoyancy is calculated for the edge cases. Maybe then we will be able to get rig of this "feature"

PR #1057, opened 2022-05-26
fix buoyancy calculations for non-equilibrium moisture model

trontrytel:
This reverts https://github.com/CliMA/TurbulenceConvection.jl/pull/1018

Right now we are inconsistent and taking `p[k]` for equilibrium `ts` but `p[k-1]` for non-equilibrium `ts`.

Also, taking `p[k-1]` results in no updrafts for shallow convection cases and delays TRMM convection by around an hour. (Compare the buildkite artifacts from before and after merging https://github.com/CliMA/TurbulenceConvection.jl/pull/1018 when looking at TRMM with non-equilibrium moisture model)


ilopezgp (approved):
LGTM

haakon-e (approved):
LGTM -- do we have described anywhere why we do (did) buoyancy calculations with `k-1` vs `k` in some document somewhere, by the way?

trontrytel:
> LGTM -- do we have described anywhere why we do (did) buoyancy calculations with `k-1` vs `k` in some document somewhere, by the way?

Not that I'm aware of. 

In general, we want to get rid of this "feature". But it needs rethinking how to handle buoyancy extrapolations at the edge cases.

PR #1058, opened 2022-05-26
add fno/rfm defaults to namelist

costachris:


yairchn:
can't we just add some set of calibrated values that seem to work well to the namelist instead? This will be reproducible behavior and we can make PR with better parameters as we find them

PR #1062, opened 2022-05-27
Decouple tests that depend on rand state

charleskawczynski:


costachris (approved):

charleskawczynski:
After several discussions, and thinking about what direction we want to move in, I think we've sorted out some things.

The initialization of some of these parameters were implemented in a way that somewhat resembles how we would want to initialize them in CEDMF, but that doesn't really matter for TC. The only thing that seems to matter from CEDMF's perspective is that we can run TC with a _given_ set of model parameters (e.g., `nn_ent_params`, `rf_opt_ent_params`, `rf_fix_ent_params` etc.).

The purpose of this PR is, at the moment, purely pragmatic: we need to further refactor namelist changes incrementally, which is significantly easier when changing model parameters do not affect other model parameters.

There are several ways to solve the issue:

 1) Use `reset_seed` (with or without an `rng`) to repeatedly reset the seed in order to decouple the model parameters. This is very quick and dirty, and does not impact CEDMF since CEDMF will set `reset_seed = false`

 2) Conditionally define the parameters, for example:
     - This is a nice option, and perhaps make more sense in general since it's a bit strange to have namelist arguments for models that are not even being used. The other thing to consider is that some of this logic will be changed when these parameters are moved into the toml dictionary (the new way we'll be handling clima parameters). I think the main question is: where and how will these parameters live--whether they are function calls to random numbers, or hard-coded numbers themselves, see 3).

```julia
    εδ_model = namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["entrainment"]
    if εδ_model == "NN"
        namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_arc"] = (6, 5, 4, 2)
        namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["nn_ent_params"] = ...
    elseif εδ_model == "RF"
        namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["rf_fix_ent_params"] = ...
        namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["rf_opt_ent_params"] = ...
    ...
    end
```

 3) Don't sample random numbers at all in TC! Instead, use already-optimized parameters, that we can store somewhere. We could store them in a function in a separate file, or functions in a separate repo, or in the clima parameters toml dict?

I think this needs a bit more thought, but I'm posting this now to track my most recent thoughts

PR #1064, opened 2022-05-27
fixes in microphysics source terms to e_tot

trontrytel:
Removing the leftover `-I` from microphysics source terms. 

The `-I` was there because we don't change the density whenever we change `q_tot` in EDMF (anelastic approximation). However we decided that we prefer to live with this small inconsistency, and keep the same formulation of source terms across the setups that do/don't change the density via microphysics. 

This PR also removes GATE from the testing pipeline


yairchn (approved):
LGTM

PR #1083, opened 2022-05-31
Limit benchmark samples

charleskawczynski:
We currently don't limit the number of samples in our benchmark, which is problematic for cases that are not intended to be run past a certain time. This PR limits the samples based on the final time and dt.

### DIFF
--- a/perf/benchmark.jl
+++ b/perf/benchmark.jl
@@ -1,23 +1,26 @@
 include(joinpath(@__DIR__, "common.jl"))
 import BenchmarkTools
 
-sim = init_sim("Bomex"; prefix = "bm_Bomex")
+sim = init_sim("Bomex"; prefix = "bm_Bomex", single_timestep = false)
 (; tendencies, prog, params, TS) = unpack_params(sim)
 ∑tendencies!(tendencies, prog, params, TS.t) # compile first
-trial = BenchmarkTools.@benchmark ∑tendencies!($tendencies, $prog, $params, $(TS.t))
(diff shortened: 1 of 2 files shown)


yairchn (approved):

perf/benchmark.jl
```diff
@@ -1,23 +1,25 @@
...
+trial = BenchmarkTools.run(b, samples = floor(Int, TS.t_max / TS.dt))
 show(stdout, MIME("text/plain"), trial)
 println()
 
-sim = init_sim("Rico"; prefix = "bm_Rico")
+sim = init_sim("Rico"; prefix = "bm_Rico", single_timestep = false)
 (; tendencies, prog, params, TS) = unpack_params(sim)
-∑tendencies!(tendencies, prog, params, TS.t) # compile first
-trial = BenchmarkTools.@benchmark ∑tendencies!($tendencies, $prog, $params, $(TS.t))
+b = BenchmarkTools.@benchmarkable ∑tendencies!(tendencies, prog, params, TS.t) # compile first
```
charleskawczynski, line 22: Whoops, forgot to interpolate here

charleskawczynski:
For some reason this is only showing 2 benchmarks, instead of 3

charleskawczynski:
Ah, only 1 sample was taken, looks like we need to compile first when using `@benchmarkable`.

charleskawczynski:
We can merge if results from this push look good, I'm optimistic

PR #1092, opened 2022-06-02
Fix SGS sampling

ilopezgp:
This PR fixes a bug in the code that enforced the use of the mean conditions in the environmental buoyancy flux, regardless of the input to thermodynamics sgs.


ilopezgp:
@trontrytel let me know whether you want "mean" or "quadratures" as the default. You can compare the buildkite of each from the two commits here.

trontrytel (approved):

Issue #897, opened 2022-03-23
Potential IO bug


charleskawczynski:
I think we may have an IO bug: I think that (based on looking at the code), for a given `nsteps`, some variables have `nsteps` outputs while others have `nsteps+1`.

charleskawczynski:
We should be able to confirm this easily by looping through each variable and checking the length of the time dimension.

trontrytel:
I wonder if it's the same bug as this: https://github.com/CliMA/TurbulenceConvection.jl/issues/821

charleskawczynski:
True, I think it's highly likely

PR #1093, opened 2022-06-02
Limit io to callbacks, sync profiles and timeseries

### LINKED ISSUE #821: Bad values in the last element of timeseries output
I'm getting garbage as a last element of the timeseries output data. 

For Dycoms_RF02 case that I'm running, for example

`cb_tc = np.array(ds_tc['timeseries']['cloud_base_mean'])`
`print(cb_tc[-1], cb_tc[-2])`

gives
`9.969209968386869e+36 375.0`

It's the same for a couple of other timeseries outputs that I checked. 

So far I have checked that the simulation itself is fine, i.e. it's not actually about to crash. It really seems to just be the output problem. I will be looking into it, but any help from more output savvy people is appreciated.

ilopezgp:
- Deletes the first io call within main, and limits all io calls to callbacks.
- Syncs timeseries and profiles writing time.
- Fixes #821 and #897

### DIFF
--- a/driver/TimeStepping.jl
+++ b/driver/TimeStepping.jl
@@ -7,6 +7,7 @@ mutable struct TimeStepping
     dt_max::Float64
     dt_max_edmf::Float64
     dt_io::Float64
+    initialized::Bool
 end
 
 function TimeStepping(namelist)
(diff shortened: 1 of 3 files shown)


charleskawczynski (commented):
I was kind of thinking to add the missing parts / unify with the callbacks so that we can maintain the initial condition

ilopezgp:
> I was kind of thinking to add the missing parts / unify with the callbacks so that we can maintain the initial condition

This should probably be still done through the callback in ODE, I will see if I can force a write at `t=0`.

charleskawczynski:
> This should probably be still done through the callback in ODE, I will see if I can force a write at t=0.

Great point, let's do that.

PR #1094, opened 2022-06-02
Use correct vector types

### LINKED ISSUE #783: Use correct field types for variables
Currently, our field variables are all scalar types (e.g., Float64), but we need to change this, to be compatible with ClimaCore's API, to use
 - `Contravariant12Vector` for horizontal velocity components (`u_h`)
 - `Contravariant3Vector` for the vertical velocity component.
 - `Covariant123Vector` for , e.g., `uvw` vectors

This will also require changing the use of `Divergence` / `Gradient` operators throughout the code, and perhaps removing some use of avoiding unnecessary `ClimaCore.Geometry.WVector`s.

charleskawczynski:
This PR changes the host dycore and TC types to be `Covariant3Vector`s where we are supposed to 
Closes #783.


charleskawczynski:
Will reapply later if necessary, this has gotten pretty stale

PR #1095, opened 2022-06-03
Limit io to callbacks, sync profiles and timeseries (v2)

### LINKED ISSUE #821: Bad values in the last element of timeseries output

ilopezgp:
- Synchronize first io call within main for all profiles and time series.
- Fixes #821 and #897.
- Tests for filled values in integration tests, which were not covered previously (hence the errors).
- Initializes integrator during initialization, and passes it as an argument to `run`.

### DIFF
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -248,30 +248,32 @@ function initialize(sim::Simulation1d)
         Cases.initialize_forcing(case, grid, state, param_set)
         Cases.initialize_radiation(case, grid, state, param_set)
         initialize_edmf(edmf, grid, state, case, param_set, t)
-        skip_io && return nothing
-        stats = Stats[inds...]
-        initialize_io(stats.nc_filename, io_nt.aux, io_nt.diagnostics)
-        initialize_io(stats.nc_filename, ts_list)
-        surf = get_surface(case.surf_params, grid, state, t, param_set)
-        open_files(stats)
-        try
-            write_simulation_time(stats, t)
-            io(io_nt.aux, stats, state)
-            io(io_nt.diagnostics, stats, diagnostics_col)
-            io(surf, case.surf_params, grid, state, stats, t)
-        catch e
-            if truncate_stack_trace
-                @warn "IO during initialization failed."
-            else
-                @warn "IO during initialization failed." exception = (e, catch_backtrace())
-            end
-            return :simulation_crashed
-        finally
-            close_files(stats)
+        if !skip_io
+            stats = Stats[inds...]
+            initialize_io(stats.nc_filename, io_nt.aux, io_nt.diagnostics)
+            initialize_io(stats.nc_filename, ts_list)
         end
     end
 
-    return
+    (prob, alg, kwargs) = solve_args(sim)
+    integrator = ODE.init(prob, alg; kwargs...)
+    skip_io && return (integrator, :success)
+
+    open_files(sim)
+    try
+        affect_io!(integrator)
+    catch e
+        if truncate_stack_trace
+            @warn "IO during initialization failed."
+        else
+            @warn "IO during initialization failed." exception = (e, catch_backtrace())
+        end
+        return (integrator, :simulation_crashed)
+    finally
+        close_files(sim)
+    end
+
+    return (integrator, :success)
 end
 
 include("callbacks.jl")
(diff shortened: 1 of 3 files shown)


ilopezgp:
Would this work @charleskawczynski ? Dealing with it through ODE is nasty

ilopezgp:
All cases seem to work, except for Bomex on the sphere.

driver/main.jl
```diff
@@ -252,13 +252,13 @@ function initialize(sim::Simulation1d)
...
+        (prob, alg, kwargs) = solve_args(sim)
+        integrator = ODE.init(prob, alg; kwargs...)
+
         open_files(stats)
         try
-            write_simulation_time(stats, t)
-            io(io_nt.aux, stats, state)
-            io(io_nt.diagnostics, stats, diagnostics_col)
-            io(surf, case.surf_params, grid, state, stats, t)
+            affect_io!(integrator)
```
charleskawczynski, line 261: Affect up loops over columns, so we don’t want to do it like this.

charleskawczynski:
What was the issue with doing this through callbacks?

ilopezgp:
> What was the issue with doing this through callbacks?

I did a deep dive into callbacks with ODE and bottomline, the package does not handle well callbacks at initialization. I tried save_start, save_at, different callbacks... None worked as I expected

ilopezgp:
@charleskawczynski I fixed the column loop and added tests for fillvalues, so we know we are not making the same mistake as before.

charleskawczynski (commented):
I think this is close.

I think we can avoid initializing the integrator twice by making an `affect_io` wrapper and calling it with the raw components in ‘initialize(::Simulation1D)`.

ilopezgp:
> I think this is close.
> 
> I think we can avoid initializing the integrator twice (which is expensive and could cause confusion with print debugging) by making an `affect_io` wrapper and calling it with the raw components in ‘initialize(::Simulation1D)`.
> 
> If what I said doesn’t make sense I can try to add what I was thinking
> 
> also, thanks for adding tests!!

I can just pass the integrator downstream

driver/main.jl
```diff
@@ -252,26 +252,26 @@ function initialize(sim::Simulation1d)
...
+        else
+            @warn "IO during initialization failed." exception = (e, catch_backtrace())
         end
+        return :simulation_crashed
+    finally
+        close_files(sim)
     end
 
-    return
+    return integrator
```
charleskawczynski, line 274:
```suggestion
    return (integrator, :success)
```
And catch / use downstream

charleskawczynski:
> I can just pass the integrator downstream

Yeah, I think that should work fine, too

ilopezgp:
I think this is ready @charleskawczynski

driver/main.jl
```diff
@@ -397,11 +397,12 @@ function main1d(namelist; time_run = true)
         end
     end
     sim = Simulation1d(namelist)
-    initialize(sim)
+    (integrator, return_init) = initialize(sim)
+    return_init == :success && @info "The initialization has completed."
```
charleskawczynski, line 401: What should we do if `return_init` is `:simulation_crashed`?

ilopezgp: I guess same as before, nothing? It will break when it starts running

charleskawczynski (approved):
Not a blocker, but we may want to more carefully handle the case when `return_init` is `:simulation_crashed` (this is probably more important for CEDMF).

driver/main.jl
charleskawczynski: It's not totally clear to me that that will be the case, but you're probably right in 99% of the cases. Again, not a blocker.

ilopezgp:
> Not a blocker, but we may want to more carefully handle the case when `return_init` is `:simulation_crashed` (this is probably more important for CEDMF).

I will implement this solution in theta-var and report back if this results in unexpected behavior during calibration.

PR #1096, opened 2022-06-03
Fix timeseries IO, theta-var branch

ilopezgp:
Fixes timeseries IO, syncing it with profiles IO.

Adds tests to check for fill values.

In order to merge this PR, I changed all performance benchmarks to using a single time step. Otherwise they broke for some reason. All case runs in the test suite ran smoothly.

This is the theta-var version of #1095 .


ilopezgp:
bors try

ilopezgp:
Do you know what's going on with the performance tests here @charleskawczynski ?

charleskawczynski:
> Do you know what's going on with the performance tests here @charleskawczynski ?

I don't know what's going on with the precompilation speed up tests, but the benchmark step one was somewhat fixed by #1083.

ilopezgp:
> > Do you know what's going on with the performance tests here @charleskawczynski ?
> 
> I don't know what's going on with the precompilation speed up tests, but the benchmark step one was somewhat fixed by #1083.

I tried implementing the changes from #1083 in a new commit, and I still get all the performance issues from before, including benchmark_step!

ilopezgp:
bors try

PR #1105, opened 2022-06-04
Remove inversion types

charleskawczynski:
I just realized that we no longer use `inversion_type`, can we remove it?


yairchn (approved):
good catch

Issue #1110, opened 2022-06-04
Move experimental parameters to appropriate structs


charleskawczynski:
We should move our experimental clima parameters into appropriate structs in preparation of the new clima parameter handling.

charleskawczynski:
After all the referenced PRs, I think this can be closed

PR #1120, opened 2022-06-05
Move more parameters into structs

charleskawczynski:
A step towards #1110.


driver/parameter_set.jl
```diff
@@ -94,6 +71,8 @@ function create_parameter_set(namelist)
         Ri_c = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "Ri_crit"),
         smin_ub = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "smin_ub"),
         smin_rm = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "smin_rm"),
+
+        # Can be removed before CP overhaul:
```
yairchn, line 75: this can be a parameter like any other that you move to struct

charleskawczynski: I'll continue with the other parameters in subsequent PRs, this PR got a bit large because we have a decent sized zoo of entrainment-detrainment models.

PR #1123, opened 2022-06-05
use the reference in initializing thetali in TRMM

yairchn:
I found that the use of the reference pressure in the computation of thetali from the initial data in TRMM creates a profile that deviates substantially from the profiles in scampy and pycles (where the reference state is defined using surface moist entropy and not using surface thetali). An alternative to use the input pressure form the paper data produces a better match. With this profile our behavior using total energy as prognostic variable is much more physical


yairchn:
This is the new theta profile in trmm (stable throughout)
Screen Shot 2022-06-05 at 1 29 21 PM

charleskawczynski (approved):
Code changes look fine to me.

yairchn:
Bors r+

PR #1126, opened 2022-06-06
change ref state to use surface MSE

yairchn:
This PR changes the references state to use the surface `MSE` value instead of the surface `θ_liq_ice`. This solves the issue in TRMM while consistently using the reference pressure in the initial profile as in the thetali version of the model.


driver/Cases.jl
```diff
@@ -580,7 +580,7 @@ function initialize_profiles(self::CasesBase{TRMM_LBA}, grid::Grid, param_set, s
         qv_star = pv_star * (1 / molmass_ratio) / denom
         aux_gm.q_tot[k] = qv_star * RH / 100
         phase_part = TD.PhasePartition(aux_gm.q_tot[k], FT(0), FT(0)) # initial state is not saturated
-        aux_gm.θ_liq_ice[k] = TD.liquid_ice_pottemp_given_pressure(param_set, aux_gm.T[k], prof_p(z), phase_part)
+        aux_gm.θ_liq_ice[k] = TD.liquid_ice_pottemp_given_pressure(param_set, aux_gm.T[k], p[k], phase_part)
```
yairchn, line 583: back to what we had two days ago, no with a working reference state

src/thermodynamics.jl
```diff
@@ -63,3 +63,15 @@ end
...
+    return mse - e_pot
+end
+
+function moist_static_energy(param_set::APS, z::FT, ts) where {FT}
+    Φ = geopotential(param_set, z)
+    e_int = TD.internal_energy(param_set, ts)
+    Rm = TD.gas_constant_air(param_set, ts)
+    T = TD.air_temperature(param_set, ts)
+    return e_int + Rm * T + Φ
+end
```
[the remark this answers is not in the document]
yairchn: yes! thanks

driver/dycore.jl
```diff
@@ -209,7 +211,10 @@ function compute_ref_state!(
     # determine the reference pressure
     function rhs(logp, u, z)
```
tapios, line 212: Avoid `rhs` as variable/function name whenever possible and replace by something descriptive. Here, what you have is minus the inverse scale height. So if you define this as `minus_inv_scale_height`, it's much clearer what's going on. (Better yet if you can take the minus sign out and put it somewhere else.)

charleskawczynski: Sounds good, this was a legacy name

PR #1130, opened 2022-06-06
use prog var in tendencies

yairchn:
it is better to use the prognostic variable in tendencies as much as possible for accuracy in an implicit solver with an estimated Jacobian. Here I make such changes in the updraft and environment equations


charleskawczynski (approved):
Code changes LGTM, results look good?

src/EDMF_functions.jl
```diff
@@ -926,7 +928,8 @@ function compute_en_tendencies!(
     prog_gm = center_prog_grid_mean(state)
     aux_en_2m = center_aux_environment_2m(state)
     tendencies_en = center_tendencies_environment(state)
-    prog_covar = getproperty(tendencies_en, prog_sym)
+    tend_covar = getproperty(tendencies_en, prog_sym)
```
charleskawczynski, line 931: ❤️ the name change

yairchn:
Bors r+

PR #1131, opened 2022-06-07
remove casename from TC.affect_filter, fix dry bubble init bug

yairchn:


charleskawczynski (approved):
No way! Does this work? If so let's merge 🎉

yairchn:
Bors r+

PR #1133, opened 2022-06-07
Move TC-external types outside of source.

charleskawczynski:
This PR moves several types outside of `src/`, including:

 - `ForcingBase` and related types
 - `RadiationBase` and related types
 - `LESData` and `compute_les_Γᵣ`

These components are really dycore-specific, and not TC specific, so it's better that we push them out to keep them disentangled from TC-specific code.

I haven't done the same with `SurfaceBase`, because it seems much more used in TC `src/` than these other components. We may be able to move it out eventually, but it may be that we can/should just use it as a part of the TC interface.


trontrytel (approved):

charleskawczynski:
bors try

PR #1138, opened 2022-06-07
Move dycore variables into separate file

charleskawczynski:
What do people think of putting all the dycore variables into its own separate file? I like the idea of having two similar files--one for TC and one for the dycore.

This PR separates all of the dycore variables (init/allocate funcs) out of `dycore.jl` into `dycore_variables.jl`.


yairchn (approved):
great idea

PR #1140, opened 2022-06-08
change subdomain mean variable to total energy

yairchn:
This draft attempts to change the subdomain mean variable to total energy. It is not working and it is not clear to me why.


driver/initial_conditions.jl
```diff
@@ -95,15 +95,19 @@ const APL = AtmosphericProfilesLibrary
...
     aux_up_f = TC.face_aux_updrafts(state)
     aux_gm = TC.center_aux_grid_mean(state)
+    aux_tc = TC.center_aux_turbconv(state)
     aux_gm_f = TC.face_aux_grid_mean(state)
     prog_gm = TC.center_prog_grid_mean(state)
     prog_up = TC.center_prog_updrafts(state)
     prog_up_f = TC.face_prog_updrafts(state)
     ρ_0_c = prog_gm.ρ
     ρ_0_f = aux_gm_f.ρ
+    @. w_up_c = aux_tc.w_up_c
```
charleskawczynski, line 110:
```suggestion
    w_up_c = aux_tc.w_up_c
```

driver/initial_conditions.jl
```diff
@@ -133,14 +138,18 @@ function initialize_updrafts_DryBubble(edmf, grid, state)
                 # for now temperature is provided as diagnostics from LES
                 aux_up[i].T[k] = prof_T(z)
                 prog_up[i].ρarea[k] = ρ_0_c[k] * aux_up[i].area[k]
-                prog_up[i].ρaθ_liq_ice[k] = prog_up[i].ρarea[k] * aux_up[i].θ_liq_ice[k]
                 prog_up[i].ρaq_tot[k] = prog_up[i].ρarea[k] * aux_up[i].q_tot[k]
+                aux_up[i].e_kin[k] = kinetic_energy(prog_gm_u[k], prog_gm_v[k], w_up_c[k])
+                ts_up_i = thermo_state_pθq(p_c[k], aux_up[i].θ_liq_ice[k], aux_up[i].q_tot[k])
```
charleskawczynski, line 143: Don’t we need to switch this to the peq constructor now?

src/EDMF_functions.jl
```diff
@@ -255,27 +255,39 @@ end
...
+    w_up_c = aux_tc.w_up_c
     @inbounds for i in 1:N_up
+        @. w_up_c = Ic(aux_up_f[i].w)
         θ_surf = θ_surface_bc(surf, grid, state, edmf, i, param_set)
         q_surf = q_surface_bc(surf, grid, state, edmf, i)
         a_surf = area_surface_bc(surf, edmf, i)
         prog_up[i].ρarea[kc_surf] = ρ_c[kc_surf] * a_surf
-        prog_up[i].ρaθ_liq_ice[kc_surf] = prog_up[i].ρarea[kc_surf] * θ_surf
         prog_up[i].ρaq_tot[kc_surf] = prog_up[i].ρarea[kc_surf] * q_surf
+        ts_up_i = thermo_state_pθq(param_set, p_c[kc_surf], θ_surf, q_surf)
```
charleskawczynski, line 286: Shouldn’t we use peq?

driver/dycore.jl
```diff
@@ -136,10 +136,11 @@ function set_thermo_state_peq!(state, grid, moisture_model, param_set)
             error("Something went wrong. The moisture_model options are equilibrium or nonequilibrium")
         end
         e_pot = TC.geopotential(param_set, grid.zc.z[k])
-        e_int = prog_gm.ρe_tot[k] / ρ_c[k] - aux_gm.e_kin[k] - e_pot
+        aux_gm.q_tot[k] = prog_gm.ρq_tot[k] / ρ_c[k]
+        aux_gm.e_tot[k] = prog_gm.ρe_tot[k] / ρ_c[k]
+        e_int = aux_gm.e_tot[k] - aux_gm.e_kin[k] - e_pot
```
charleskawczynski, line 141: I'd really prefer that we move these definitions into a separate function, maybe `update_grid_mean_aux!`. This function says that it sets the thermo state but it currently does more than that, which is misleading.

src/EDMF_functions.jl
```diff
@@ -255,27 +255,39 @@ end
...
+        e_kin = kinetic_energy(prog_gm_u[kc_surf], prog_gm_v[kc_surf], w_up_c[kc_surf])
```
charleskawczynski, line 287: This function is being deprecated, please see the new use in #1100.

src/update_aux.jl
```diff
@@ -63,21 +63,33 @@ function update_aux!(edmf::EDMFModel, grid::Grid, state::State, surf::SurfaceBas
                     aux_up[i].θ_liq_ice[k] = θ_surf
                     aux_up[i].q_tot[k] = q_surf
                     aux_up[i].area[k] = a_surf
+                    ts_up_i = thermo_state_pθq(param_set, p_c[k], θ_surf, q_surf)
```
charleskawczynski, line 66: `θ_surface_bc` above uses `aux_gm.θ_liq_ice`, should that function be rewritten?

charleskawczynski (commented):
I'm wondering if we should move `ρaθ_liq_ice`/`θ_liq_ice` to the diagnostics array--so that we don't have access and can't accidentally use it in `src/`. Unless, of course, we still need it for some reason.

driver/initial_conditions.jl
yairchn: the initial condition is given in terms of `θ_liq_ice`

src/EDMF_functions.jl
yairchn: the surface value of `θ_liq_ice` is computed from the surface distribution of `θ_liq_ice`.

src/EDMF_functions.jl
charleskawczynski:
```suggestion
        e_kin = aux_up[i].e_kin[kc_surf]
```
We've already computed `aux_up[i].e_kin`--no need for `copy`

yairchn: inside `affect_filter!` we should not aux variables

driver/dycore.jl
yairchn: see #1149

PR #1141, opened 2022-06-08
Clean up unused functions

charleskawczynski:
We're no longer using `total_enthalpy` and `kinetic_energy`. Also, the way we compute `kinetic_energy` has now changed due to the vector types. This was missed in #1100.

### DIFF
--- a/src/thermodynamics.jl
+++ b/src/thermodynamics.jl
@@ -42,6 +42,7 @@ function geopotential(param_set, z::FT) where {FT}
     return grav * z
 end
 
+# TODO: move to Thermodynamics.jl
 function total_enthalpy(param_set::APS, e_tot::FT, ts) where {FT}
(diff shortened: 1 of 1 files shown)


src/thermodynamics.jl
```diff
@@ -42,24 +42,13 @@ function geopotential(param_set, z::FT) where {FT}
...
 end
 
+# TODO: move to Thermodynamics.jl
 function total_enthalpy(param_set::APS, e_tot::FT, ts) where {FT}
     Rm = TD.gas_constant_air(param_set, ts)
     T = TD.air_temperature(param_set, ts)
     return e_tot + Rm * T
 end
 
-function total_enthalpy(param_set::APS, e_tot::FT, p::FT, ρ::FT) where {FT}
```
yairchn, line 51: can we keep these functions until we switch the updraft/env to total energy they are good for experimentation

PR #1142, opened 2022-06-08
remove is surface in update_aux

yairchn:
these calls for surface bc in the updraft were unneeded


charleskawczynski (approved):
Code changes look good to me

yairchn:
Bors r+

PR #1147, opened 2022-06-08
remove the SullivanPatton case

yairchn:
we are not using or testing this case and we don't need another DCBL case


charleskawczynski (approved):
I'm personally okay with this, but I'd prefer we don't drop this case unless _everyone_ is on board.

yairchn:
Bors r+

yairchn:
Bors r-

yairchn:
Bors r+

PR #1151, opened 2022-06-09
apply code changes, need to resolve packages

yairchn:
I made the code changes but it requires upgrading Thermodynamics to have `PhaseEquil_phq`and this requires updating ClimaParameters ...

PR #1152, opened 2022-06-09
use input pressure in TRMM initial state

yairchn:
this is an alternative #1151


trontrytel (approved):

trontrytel:
The TRMM TKE profile doesn't look good after this change. Maybe it's better to stick to the old way of initialising the profile here

yairchn:
> The TRMM TKE profile doesn't look good after this change. Maybe it's better to stick to the old way of initialising the profile here

well the correct fix is #1151 but it requires upgrading the version in thermodynamics, which requires upgrading the ClimaParameters version, which requires upgrading microphysics ....

trontrytel:
> > The TRMM TKE profile doesn't look good after this change. Maybe it's better to stick to the old way of initialising the profile here
> 
> well the correct fix is #1151 but it requires upgrading the version in thermodynamics, which requires upgrading the ClimaParameters version, which requires upgrading microphysics ....

Do you know versions of those packages are needed? I'll give it a try

PR #1154, opened 2022-06-10
Add/use thermo parameter wrapper

charleskawczynski:
This PR:
 - Adds a wrapper `thermodynamics_params(ps) = ps.thermo_params` in TC, and uses this to extract the thermodynamics parameter set from the `param_set`. For now, it just returns a `ThermodynamicsParameters` that we define locally (and overload the appropriate clima parameter methods). When we change to the new parameter system, we can just change `ThermodynamicsParameters` to `TD.ThermodynamicsParameters` with the appropriate parameters.

Arguably, there are some places where we could extract the thermo params and pass them into function calls to limit which variables those functions have access to, however, if we need to (for example) add ice contributions to some term (which need microphysics parameters) then we'd need to change the interface back to include more parameters. So, the approach taken here is: extract the thermo params _just before_ calling `TD.thermo_function`. We can apply the same pattern to SurfaceFluxes and CloudMicryophysics. That said, now that I'm reviewing this, our thermo wrappers may be an exception to this since they're only there for experimenting with the thermo configs..

This is the best way I can think about bracing for impact when we switch to the new parameter system.

### DIFF
--- a/driver/Cases.jl
+++ b/driver/Cases.jl
@@ -228,10 +228,11 @@ initialize_forcing(::AbstractCaseType, forcing, grid::Grid, state, param_set) = 
 ForcingBase(case::Soares, param_set::APS; kwargs...) = ForcingBase(get_forcing_type(case); apply_coriolis = false)
 
 function surface_ref_state(::Soares, param_set::APS, namelist)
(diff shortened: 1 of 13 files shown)


charleskawczynski:
Let's live on the edge. bors r+

PR #1159, opened 2022-06-10
Yc/move coriolis namelist

yairchn:
change Coriolis parameter to always exists, with the value of zero in places it was not applied before, and remove `apply_coriolis` flag


charleskawczynski (approved):
Code changes LGTM

yairchn:
Bors r+

PR #1161, opened 2022-06-10
expand name aliases

costachris:
Add cloud top and cloud base to name aliases.


ilopezgp (approved):
LGTM. FYI, these diagnostics require !calibrate_io, I hope that's ok. I prefer keeping it this way since otherwise we need to traverse the column and compute a few more diagnostics.

charleskawczynski:
We could disable the mse tests in this branch if it's causing too much friction. The way it's currently configured, any given PR may need mse table updates.

PR #1173, opened 2022-06-14
Use correct ClimaCore types and operators for horizontal velocities

charleskawczynski:
This PR
 - Uses the correct types for some variables, namely:
    - `(ug, vu) -> uₕ_g`
    - `(sgs_flux_u,sgs_flux_v) -> sgs_flux_uₕ`
    - `(diffusive_flux_u,diffusive_flux_v) -> diffusive_flux_uₕ`
  - Initializes and mutates them using the constructors that are compatible with the sphere

There are some places where we'll need to revisit so that we're also compatible with topography, but it's probably best that we revisit that when we have something to test.


driver/dycore.jl
```diff
@@ -392,8 +417,7 @@ function compute_gm_tendencies!(
             tendencies_gm.ρq_tot[k] += ρ_c[k] * (gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k])
             tendencies_gm.ρe_tot[k] +=
                 ρ_c[k] * (cv_m * (gm_T_nudge_k + T_fluc) + h_v * (gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k]))
-            tendencies_gm_u[k] += gm_U_nudge_k
-            tendencies_gm_v[k] += gm_V_nudge_k
+            tendencies_gm_uₕ[k] += CCG.Covariant12Vector(CCG.UVVector(gm_U_nudge_k, gm_V_nudge_k), lg[k])
```
charleskawczynski, line 420: @simonbyrne, @yairchn, I have a feeling that this is not correct, what should we be doing we this? I'm not sure how we can / should compute `gm_U_nudge_k = (aux_gm.u_nudge[k] - prog_gm_u[k]) / force.wind_nudge_τᵣ`. Is this okay if `prog_gm_u = TC.physical_grid_mean_u(state)`? (which wraps `uₕ` in `UVVector`)

driver/dycore.jl
```diff
@@ -338,10 +339,37 @@ function compute_gm_tendencies!(
...
+    # f = @. CCG.Contravariant3Vector(
+    #     # CCG.WVector(2 * Ω * sind(lat)),
+    #     CCG.WVector(coriolis_param * ones(axes(ρ_c))),
+    # )  # TODO!: Will need to be changed with topography
+
+    C123 = CCG.Covariant123Vector
+    C12 = CCG.Contravariant12Vector
+    lg = CC.Fields.local_geometry_field(axes(ρ_c))
+    # hcurl = CCO.Curl()
+    # ω³ = @. hcurl(prog_gm_uₕ) # Contravariant3Vector
```
yairchn, line 367: do we need this comment ?

driver/dycore.jl
```diff
@@ -338,10 +339,37 @@ function compute_gm_tendencies!(
...
+
+    # (f + ω³) × # for horizontal advection
```
yairchn, line 369: do we need this comment ?

driver/dycore.jl
yairchn: I don't think we would use it anywhere but in a SCM setting if that helps

src/EDMF_functions.jl
```diff
@@ -741,22 +748,22 @@ function compute_covariance_shear(
...
-    v = prog_gm_v
     area_en = aux_en_c.area
     shear = aux_covar.shear
 
+    C123 = CCG.Covariant123Vector
+    local_geometry = CC.Fields.local_geometry_field(axes(ρ_c))
+    k̂ = @. CCG.Contravariant3Vector(CCG.WVector(FT(1)), local_geometry)
+    Ifuₕ = uₕ_bcs()
+    ∇uvw = CCO.GradientF2C()
+    # TODO: k_eddy should probably be a tensor (Contravariant3 tensor)
```
yairchn, line 759: maybe add 'and Shear² should be a tensor as well so that the product would be a scalar'

yairchn (approved):

driver/dycore.jl
charleskawczynski: Ok, I guess we can leave it if it will never run in a sphere setting.

yairchn: you could compute this the same way the Coriolis tendency is computed as a difference between u and ug, `aux_gm.u_nudge` would be a horizontal vector here and `gm_U_nudge_k` would be the same type as the type as the tendency of that vector, if `1/force.wind_nudge_τᵣ` will be the same type as the Coriolis parameter , here we would take for product instead of a cross product

charleskawczynski: I'm going to leave it as is for now, since the results look fine. We can change/fix it later if needed.

PR #1174, opened 2022-06-15
add stochastic Rico

haakon-e:
test runtime for stochastic rico


charleskawczynski:
@haakon-e, I've added some performance patches, can you rebase and re-push this branch to see how this case does? I suspect that we're suffering from a small timestep, which we may not be able to help with much. This would mean that we're calling the callbacks more, which is where the performance fix was applied (in #1186).

charleskawczynski:
Looks like it's a bit improved: `6:24` -> `4:54`

charleskawczynski:
Hmm, the runtime is allocating quite a bit, we could aim the allocation monitoring at a stochastic job to see what's going on there

PR #1179, opened 2022-06-15
Fix type instabilities

charleskawczynski:
This PR:

 - Removes hard-coded `Float64`, and makes `ForcingBase` and `RadiationBase` type-generic
 - Removes some `ForcingBase` outer constructors, and fallback on the abstract type case, instead. This is possible because of several previous refactoring changes that resulted in most of the `ForcingBase` definitions to be nearly unified.
 - Converts the `EDMFModel` from an inner constructor to an outer constructor. This may be helpful for @dennisYatunin when experimenting with Duals.

I still need to try changing our duals test so that we use Float64 in the EDMF model and apply type promotions where needed, however, this may take some time so let's merge this in first.

### DIFF
--- a/driver/Cases.jl
+++ b/driver/Cases.jl
@@ -110,32 +110,30 @@ LES-driven forcing
 
 $(DocStringExtensions.FIELDS)
 """
-Base.@kwdef struct ForcingBase{T}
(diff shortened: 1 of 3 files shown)


dennisYatunin:
Thank you, Charlie!

PR #1185, opened 2022-06-16
Fully run duals test

charleskawczynski:
This PR changes our duals test to fully run with duals: we will run (and test) the exact same vanilla Bomex case with Float64 and with Dual numbers.

If our duals test pass (our duals simulation output will be compared with the results from the vanilla Bomex), then everything regarding running TC with duals should be fine on the TC side. cc @dennisYatunin.

Next, I'll try initializing EDMFModel with Float64, but I'm not exactly sure how many changes that this will require.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -164,7 +164,7 @@ steps:
     steps:
 
       - label: ":partly_sunny: Bomex duals"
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Looks like it's passing the test 🎉

PR #1186, opened 2022-06-16
Fix type instabilities

charleskawczynski:
This PR fixes a few type instabilities in `dt_max!`

I'm assuming that this should be okay, @trontrytel ? If not, perhaps we could use `sqrt(ε)`?


trontrytel:
Looks good to me.

FYI - those CFL limiters don't always work for precipitation. I.e. you will still be sometimes getting errors for too big dt for the precipitation terminal velocity.

charleskawczynski:
Fixing the type instability wasn't sufficient to fixing the issue. I ended up having to add a function barrier in order to fix some inference issues.

PR #1194, opened 2022-06-19
Add compat entries, update deps

charleskawczynski:
This PR
 - Adds some compat entries for some dependencies
 - ~Pins DiffEqBase in the perf, test and integration_tests environments~
 - updates the dependencies

I opened an issue in DiffEqBase. Fixed in [SciMLBase#778](https://github.com/SciML/DiffEqBase.jl/pull/778).


ilopezgp:
So, do we need to add `ForwardDiff = "0.10"` in CEDMF as well?

PR #1196, opened 2022-06-21
add second entrainment as none

yairchn:
I am adding here a second entrainment model with a default of `None` that can take the `moisture deficit` model. The resulting entrainment is the sum of the entrainment from the two models so that we can use. it to learn a correction for `moisture deficit` using NN, FNO RF etc.


yairchn:
I will make a similar PR to main so that once we move the TC version in CEDMF the ability will remain

charleskawczynski (approved):
Code changes LGTM

src/update_aux.jl
```diff
@@ -300,7 +300,7 @@ function update_aux!(edmf::EDMFModel, grid::Grid, state::State, surf::SurfaceBas
     if edmf.moisture_model isa NonEquilibriumMoisture
         compute_nonequilibrium_moisture_tendencies!(grid, state, edmf, Δt, param_set)
     end
-    compute_entr_detr!(state, grid, edmf, param_set, surf, Δt, edmf.entr_closure)
+    compute_entr_detr!(state, grid, edmf, param_set, surf, Δt, edmf.entr_closure, edmf.added_entr_closure)
```
costachris, line 303: Can we just extract `edmf.added_entr_closure`  within the function, since the whole `edmf` object is passed anyway? The final `edmf.entr_closure` was passed as a separate argument for dispatching.

driver/parameter_set.jl
```diff
@@ -58,11 +58,16 @@ function create_parameter_set(namelist)
     TC = TurbulenceConvection
 
     # this is needed to avoid slowdown when using large parameter vectors
-    use_ran_features = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment") == "RF"
-    use_nn = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment") == "NN"
-    use_nn_nonlocal = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment") == "NN_nonlocal"
-    use_fno = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment") == "FNO"
-    use_linear = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment") == "Linear"
+    use_ran_features = TC.parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "entrainment") == "RF" ||
```
costachris, line 61: I think it suffices to leave this as before,  since the only valid `add_entr_type` is `moisture_deficit` and all the parameters needed for md are already in the parameter_set. Users will set one of these models to be added to the md closure (ie. `entrainment = "NN"`), so the appropriate "use_X_MODEL" will be executed.

costachris (commented):
I think we need equivalent logic for the nonlocal `compute_entr_detr!` function as well. As far as I can tell this wouldn't add anything for ie. the FNO

trontrytel:
Looks good to me.

Do you also want to add it as a command line argument?

PR #1199, opened 2022-06-21
Remove allocs in percentile_bounds_mean_norm

charleskawczynski:
This PR eliminates the allocations in `percentile_bounds_mean_norm`, which accounts for ~40% of allocations in all of the vanilla cases.


charleskawczynski:
@yairchn, @haakon-e any idea why the noisy relaxation process job is failing here? All of the cases seem virtually unchanged, is this job on the brink of failing? Running the new function offline in comparison with the original shows that it's equivalent to machine precision:

```julia
import Distributions as D
import Random
import StatsBase

function percentile_bounds_mean_norm()
    set_src_seed = true
    FT = Float64
    low_percentile = 0.9
    high_percentile = 1.0
    n_samples = 1000
    set_src_seed && Random.seed!(123)
    ∑samples = zero(FT)
    xp_low = D.quantile(D.Normal(), low_percentile)
    xp_high = D.quantile(D.Normal(), high_percentile)
    n_filtered_samples = 0
    @inbounds for i in 1:n_samples
        x = rand(D.Normal())
        if xp_low < x < xp_high
            ∑samples += x
            n_filtered_samples += 1
        end
    end

    set_src_seed && Random.seed!(123)
    x = rand(D.Normal(), n_samples)
    xp_low = D.quantile(D.Normal(), low_percentile)
    xp_high = D.quantile(D.Normal(), high_percentile)
    filter!(y -> xp_low < y < xp_high, x)
    pbmn_old = StatsBase.mean(x)

    pbmn_new = ∑samples / n_filtered_samples
    return pbmn_old, pbmn_new
end

Δmax = 0
Δnext_sampled_num = 0
for i in 1:10000
    pbmn_old, pbmn_new = percentile_bounds_mean_norm()
    Δmax = max(Δmax, abs(pbmn_old-pbmn_new))
end
@show Δmax
```

charleskawczynski:
this is really puzzling..

src/utility_functions.jl
```diff
@@ -3,16 +3,23 @@
...
-    x = rand(D.Normal(), n_samples)
+    ∑samples = zero(FT)
     xp_low = D.quantile(D.Normal(), low_percentile)
     xp_high = D.quantile(D.Normal(), high_percentile)
-    filter!(y -> xp_low < y < xp_high, x)
-    return StatsBase.mean(x)
+    n_filtered_samples = 0
+    @inbounds for i in 1:n_samples
+        x = rand(D.Normal())
+        if xp_low < x < xp_high
```
yairchn, line 17: should there be a `<=` or `=<` here ?

haakon-e:
I'll have a look.

yairchn (approved):
lgtm , we should prob make similar changes in CA, right?

charleskawczynski:
> lgtm , we should prob make similar changes in CA, right?

Yes, absolutely

Issue #133, opened 2021-08-16
Make code type-stable


charleskawczynski:
There are _many_ places in the code that are not floating precision type-stable (we often promote to Float64). When things are well ironed out, we should take a pass at making the code more type stable wrt floating precision.

charleskawczynski:
Basically closed by #1201

PR #1201, opened 2022-06-21
Add Float32 support and test

charleskawczynski:
This PR fixes more type instabilities and adds a Float32 test


charleskawczynski:
Whohoo! I think that we can close #133. We may want / need to continue fixing type instabilities as they arise, but I think that this was the last of many many type-unstable pieces.

PR #1203, opened 2022-06-21
use prog qt in set_thermo_state_peq

yairchn:
We call `set_thermo_state_peq` before we update the aux vars, so we should get `qt` in `set_thermo_state_peq` from the prognostic state, not the aux state


charleskawczynski (approved):
Good catch!

yairchn:
Bors r+

PR #1209, opened 2022-06-23
add a second entrainment model with default to None

yairchn:
similar to #1196 for the main branch


yairchn:
see #1253

yairchn:
see https://github.com/CliMA/TurbulenceConvection.jl/pull/1253

PR #1211, opened 2022-06-25
Improve RF performance

charleskawczynski:
This PR improves the Random Feature model performance by
 - using two smaller `SVectors` for the RF model parameters which, AFAIU, should result in shorter compile times for longer vectors
 - using `NTuple`s for the RF model parameters which, AFAIU, should also result in shorter compile times compared to SVectors (see [this issue](https://github.com/JuliaLang/julia/issues/35547)). Testing out that issue with Tuples results in better compile times, perhaps Base does more clever things than what's possible with SVectors. We can't do the same for matrices, since we still need to call `LinearAlgebtra.dot`.
 - Try to avoid intermediate arrays/allocations in `non_dimensional_function`, which is the major runtime issue at the moment.


charleskawczynski:
Superseded by #1212.

PR #1212, opened 2022-06-25
Try using Arrays in RF model

charleskawczynski:
This is an alternative to #1211. This PR uses `Array`s in the Random Feature model, since large SArrays and Tuples (O(1000)) are _really_ expensive for the compiler. This PR tests out ordinary arrays.

It'll be interesting to compare both the compile time and runtime.

### DIFF
--- a/driver/main.jl
+++ b/driver/main.jl
@@ -145,9 +145,11 @@ function Simulation1d(namelist)
     end
 
     edmf = TC.EDMFModel(FTD, namelist, precip_model, rain_formation_model)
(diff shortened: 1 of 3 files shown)


charleskawczynski:
RF model with Arrays run in 5 min, so this basically fixes the problem. I'm inclined to say we should merge this in the name of getting CI to run faster, but we should think of a long-term stack allocated-friendly way to do this. My other concern is that people will start extending the condition for using Arrays and we'll end up making the code not compatible with gpus.

Issue #319, opened 2021-09-27
Improve IO configuration


charleskawczynski:
My first sketch was very automated, but maybe a bit magical and difficult to understand how things would need to change as it evolves. It also involves several baked in assumptions that I dislike. We're going to follow a simpler `Dict`-based solution for now. Unfortunately this will scale the same as our current solution.

It would be nice if we could somehow specify a single Dict. Maybe, in addition to specifying meta info, dict values could hold references to the fields. This would cut our io lines of code nearly in half.

```julia
function initialize_field(f::CC.Fields.Field, Stats::NetCDFIO_Stats, coord; prefix = "", suffix = "")
    for pn in propertynames(f)
        prop = getproperty(f, pn)
        name = prefix == "" ? "$(pn)_" : "$(prefix)$(pn)_"
        if isempty(propertynames(prop))
            full_name = name * suffix
            group = occursin("ref_state", name) ? "reference" : "profiles"
            t = group == "reference" ? () : ("t",)
            meta = (; dims = (coord, t...), group = group)
            add_field(Stats, full_name; meta...)
        else
            initialize_field(prop, Stats, coord; prefix = name, suffix = suffix)
        end
    end
end

function initialize_io(f::CC.Fields.FieldVector, Stats::NetCDFIO_Stats)
    initialize_field(center_fields(f), Stats, "zc"; suffix = "c")
    initialize_field(face_fields(f), Stats, "zf"; suffix = "f")
end

function export_field(f::CC.Fields.Field, Stats::NetCDFIO_Stats, coord; prefix = "", suffix = "")
    for pn in propertynames(f)
        prop = getproperty(f, pn)
        name = prefix == "" ? "$(pn)_" : "$(prefix)$(pn)_"
        if isempty(propertynames(prop))
            full_name = name * suffix
            group = occursin("ref_state", name) ? "reference" : "profiles"
            write_field(Stats, full_name, prop; group = group)
        else
            export_field(prop, Stats, coord; prefix = name, suffix = suffix)
        end
    end
end

function io(f::CC.Fields.FieldVector, Stats::NetCDFIO_Stats)
    export_field(center_fields(f), Stats, "zc"; suffix = "c")
    export_field(face_fields(f), Stats, "zf"; suffix = "f")
end
```

charleskawczynski:
I think that the io_dict functions that were introduced a while back is relatively elegant. The only piece I'm not super happy about is the dict inside `NetCDFIO_Stats`.

PR #1214, opened 2022-06-27
Refactor IO

### LINKED ISSUE #881: Make `NetCDFIO_Stats` immutable
I think that making `NetCDFIO_Stats` immutable is somewhat in reach. We can do this without changing our dict hack by merging `NetCDFIO_Stats` constructor and the `open_files` function.

charleskawczynski:
This PR
 - Removes the `vars` hack that was added in, ages ago, due to [NCDatasets#135](https://github.com/Alexander-Barth/NCDatasets.jl/issues/135)
 - Makes the `NetCDFIO_Stats` immutable and more portable since it now only depends on the filename and the IO frequency

Closes #881, #319


charleskawczynski:
I've fixed a few things, but the precompile speedup for Bomex has gone from ~10 to ~40 secs. Looking at the flame graph, the IO time is nearly evenly split between `get_var`, `setindex!` (unavoidable) and openeing/closing the NC file. It seems that the perf hit is unavoidable with these changes. Overall runtime is really not that bad, but I'm going to think about this for a bit, since it impacts calibration folks most.

PR #1215, opened 2022-06-28
Change reference state to MSE.

ilopezgp:
# PULL REQUEST

## Purpose and Content

- Use the same reference state as the main version of TC.jl to prevent spurious behavior in deep convective cases.

I followed PR #1151, adding the necessary Thermodynamics functions that were added after version 0.7.1 of Thermodynamics.


yairchn (approved):
LGTM

ilopezgp:
This PR fixes the deep convection behavior in TRMM-LBA. It also leads to changes at the inversion height in the LES_driven_SCM shallow case; the inversion becomes less smooth with these changes. I will make a release and test the deep/shallow behavior of the calibrated model in the LES_driven_SCM cases.

PR #1222, opened 2022-07-11
Fix contour plots

charleskawczynski:
The contour plot axis/title sizes are messed up from the recent Plots.jl update. This PR ~is an attempt to fix them~ reverts and pins an old version of Plots until we hear back about [Plots#4273](https://github.com/JuliaPlots/Plots.jl/issues/4273). @yairchn, @trontrytel.


trontrytel:
This was fixed

PR #1229, opened 2022-07-18
Add integrated vertical SGS flux diagnostics.

ilopezgp:
# PULL REQUEST

## Purpose and Content

Adds diagnostics for two quantities of interest/metrics of the EDMF scheme: the integrated vertical subgrid-scale fluxes of moisture and entropy. 

## Benefits and Risks

- Easy analysis of EDMF response to global warming.
- Enables comprehensible scalar-based loss functions for calibration.

### DIFF
--- a/driver/compute_diagnostics.jl
+++ b/driver/compute_diagnostics.jl
@@ -189,6 +189,10 @@ function compute_diagnostics!(
     # We only need computations above here for calibration io.
     calibrate_io && return nothing
 
+    # Integrated vertical subgrid-scale fluxes (already multiplied by density)
+    diag_svpc.integ_total_flux_qt[cent] = sum(aux_tc_f.massflux_qt .+ aux_tc_f.diffusive_flux_qt)
+    diag_svpc.integ_total_flux_s[cent] = sum(aux_gm_f.massflux_s .+ aux_gm_f.diffusive_flux_s)
+
     #####
     ##### Cloud base, top and cover
     #####
--- a/driver/callbacks.jl
+++ b/driver/callbacks.jl
@@ -53,6 +53,9 @@ function affect_io!(integrator)
         write_ts(Stats, "cloud_cover_mean", diag_svpc.cloud_cover_mean[cent])
         write_ts(Stats, "cloud_base_mean", diag_svpc.cloud_base_mean[cent])
         write_ts(Stats, "cloud_top_mean", diag_svpc.cloud_top_mean[cent])
+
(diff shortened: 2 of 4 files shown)


driver/compute_diagnostics.jl
```diff
@@ -189,6 +190,10 @@ function compute_diagnostics!(
     # We only need computations above here for calibration io.
     calibrate_io && return nothing
 
+    # Integrated vertical subgrid-scale fluxes
+    diag_svpc.integ_total_flux_qt[cent] = sum(ρ_f .* (aux_tc_f.massflux_qt + aux_tc_f.diffusive_flux_qt))
```
yairchn, line 194: you could use `sgs_flux_q_tot` directly - this has the SF.jl BC accounted for

driver/compute_diagnostics.jl
```diff
@@ -189,6 +190,10 @@ function compute_diagnostics!(
...
+    diag_svpc.integ_total_flux_s[cent] = sum(ρ_f .* (aux_gm_f.massflux_s .+ aux_gm_f.diffusive_flux_s))
```
yairchn, line 195: I think that it is importnat to make sure that the surface value of the combined `aux_gm_f.massflux_s .+ aux_gm_f.diffusive_flux_s` is given by SF.jl consistently otherwise it might effect the loss function

driver/compute_diagnostics.jl
ilopezgp: This is a good point @yairchn, and something I have to think about for the entropy. For the moisture flux, this already includes the BC you are referring to. 

We are currently applying the BC to `aux_tc_f.diffusive_flux_qt`, since the mass flux at the lowest level is 0 (`w_up=0`). You can find it [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/theta-var/src/EDMF_functions.jl#L212), where the BC is defined [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/theta-var/src/EDMF_functions.jl#L206).

driver/compute_diagnostics.jl
ilopezgp: We currently do not have an output from SF.jl regarding entropy. There is already a comment in the code pointing to this deficiency, which I agree we should fix. However, the actual implementation of `total_flux_s` is closer to the `s_flux_z` diagnostic from PyCLES, which only includes the resolved LES flux (and hence not the surface flux).

PR #1231, opened 2022-07-19
A bugfix for noneq microphysics to bring it up to date with current main state.

jbphyswx:
# PULL REQUEST

## Purpose and Content
We need to bring our non-equilibrium microphysics up to date with the current main branch now that it is back to using theta as the prognostic variable. This will allow for continuity in development and easier integration later on.

### DIFF
--- a/integration_tests/Manifest.toml
+++ b/integration_tests/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.7.2"
 manifest_format = "2.0"
 
 [[deps.AbstractFFTs]]
@@ -404,7 +404,7 @@ uuid = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 version = "0.8.6"
 
 [[deps.Downloads]]
-deps = ["ArgTools", "FileWatching", "LibCURL", "NetworkOptions"]
+deps = ["ArgTools", "LibCURL", "NetworkOptions"]
 uuid = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 
 [[deps.DualNumbers]]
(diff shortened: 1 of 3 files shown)


src/EDMF_Environment.jl
```diff
@@ -59,6 +59,15 @@ function microphysics(
         if edmf.moisture_model isa NonEquilibriumMoisture
             aux_en.ql_tendency_precip_formation[k] = mph.ql_tendency * aux_en.area[k]
             aux_en.qi_tendency_precip_formation[k] = mph.qi_tendency * aux_en.area[k]
+
+            # keep track of non_eq_microphysics tendency part....
+            T_en = aux_en.T[k]
+            q_en = TD.PhasePartition(aux_en.q_tot[k], aux_en.q_liq[k], aux_en.q_ice[k])
+            ts_en= TD.PhaseNonEquil_pTq(param_set, p_c[k], T_en, q_en)
```
charleskawczynski, line 66:
```suggestion
            ts_en = ts_env[k]
```
I think that this should work too, since we already compute the environmental thermo state in `update_aux.jl` before this is called.

Also, above in this function, we'll need to add `ts_env = center_aux_environment(state).ts`.

integration_tests/Manifest.toml
```diff
@@ -1665,7 +1665,7 @@ version = "1.0.1"
 
 [[deps.Thermodynamics]]
 deps = ["DocStringExtensions", "KernelAbstractions", "Random", "RootSolvers"]
-git-tree-sha1 = "72c84f7322bf3f03ed648b2ca6b491e1d74b087f"
+path = "/home/jbenjami/Research_Schneider/CliMa/Thermodynamics.jl"
 uuid = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
```
trontrytel, line 1669: We should get rid of this and go back to using the default Thermodunamics

jbphyswx: oh i think that's because when I ran the tests i still had  a pkg.develop in my driver --- just running it again  with the develop line removed should remake the manifest and then I can just push again? or is it okay/better just to copy from this line here?

trontrytel: I think either way should work

PR #1237, opened 2022-07-22
Add most missing pieces from theta-var.

ilopezgp:
# PULL REQUEST

## Purpose and Content

- Merges #1217 to main branch,
- Merges #1229 to main branch,
- Adds back the `calibrate_io` clause, which seems to have been deleted. Was this intentional @charleskawczynski ? Are we missing it in other parts of the code?

## Benefits and Risks

- Benefit: Potentially return to the main branch for research purposes.

### DIFF
--- a/driver/callbacks.jl
+++ b/driver/callbacks.jl
@@ -41,6 +41,9 @@ function affect_io!(integrator)
         write_ts(stats, "rwp_mean", diag_svpc.rwp_mean[cent])
         write_ts(stats, "swp_mean", diag_svpc.swp_mean[cent])
 
+        write_ts(stats, "integ_total_flux_qt", diag_svpc.integ_total_flux_qt[cent])
+        write_ts(stats, "integ_total_flux_s", diag_svpc.integ_total_flux_s[cent])
+
         if !calibrate_io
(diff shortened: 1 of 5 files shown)


charleskawczynski (approved):

charleskawczynski:
> Was this intentional

No, I think the IO refactoring (to allow for adding the sphere test case) were pretty large and complicated, so I probably removed that by accident

ilopezgp:
> > Was this intentional
> 
> No, I think the IO refactoring (to allow for adding the sphere test case) were pretty large and complicated, so I probably removed that by accident

Sounds good! Let's see if this works

ilopezgp:
@charleskawczynski it seems like I need to convert the new diagnostics to `Duals` to be able to write them to file, any idea how to do this?

```


│   exception =
--
  | │    MethodError: no method matching write_ts(::Array{NetCDFIO_Stats{Float64}, 3}, ::String, ::Float64)
  | │    Closest candidates are:
  | │      write_ts(::Any, ::Any, ::ForwardDiff.Dual) at /central/scratch/esm/slurm-buildkite/turbulenceconvection-ci/11549/turbulenceconvection-ci/driver/NetCDFIO.jl:160
```

PR #1238, opened 2022-07-22
Correct grid truncation by enforcing z_top.

ilopezgp:
# PULL REQUEST

## Purpose and Content

The current truncated mesh represents a truncation of the GCM mesh that retains enough elements of the GCM mesh to cover the entire requested space, without modifying the mesh points. This means that if a `z_top=4000.0` is requested, the last element of `TC.Grid(mesh).zf.z` is greater or equal than 4000.0, but may be e.g. 4200.0. This causes inconsistencies in the forcing of different benchmarks. How do we force/initialize beyond the original domain? The answer is non-trivial, case-specific and prone to error.

The new truncated mesh retains the number of degrees of freedom or the original truncated mesh, but recomputes the stretch necessary to reach `z_top` exactly. This means that in the end we have  `vec(`TC.Grid(mesh).zf.z`)[end] == 4000.0`.

## Benefits and Risks

Avoids serious problems in the definition of fields outside the original domain. Right now, we are doing a constant extrapolation beyond `z_top`, which means for example that we are relaxing to mixed layers in LES_driven_SCM.

### DIFF
--- a/driver/common_spaces.jl
+++ b/driver/common_spaces.jl
@@ -71,7 +71,6 @@ function construct_mesh(namelist; FT = Float64)
     end
 
     z₀, z₁ = FT(0), FT(nz * Δz)
-    z_stretch = CC.Meshes.Uniform()
 
     z_mesh = if truncated_gcm_mesh
         nzₛ = namelist["grid"]["stretch"]["nz"]
--- a/src/Grid.jl
+++ b/src/Grid.jl
@@ -25,7 +25,26 @@ function TCMeshFromGCMMesh(gcm_mesh; z_max::FT) where {FT <: AbstractFloat}
     faces = map(1:(k_star.i)) do k
         CC.Geometry.ZPoint{FT}(gcm_grid.zf[CCO.PlusHalf(k)].z)
     end
-    return CC.Meshes.IntervalMesh(domain, faces)
+    truncated_mesh = CC.Meshes.IntervalMesh(domain, faces)
+    # Adjust mesh, same degrees of freedom
+    return adjust_mesh(truncated_mesh, z_max)
+end
+
+"Adjusts a stretched mesh to reach an exact domain top, retaining surface Δz and nz"
+function adjust_mesh(mesh, z_top::FT) where {FT <: AbstractFloat}
+    grid = Grid(mesh)
+    zf_vec = vec(grid.zf.z)
+    nz = length(zf_vec)
+    Δz_top = zf_vec[end] - zf_vec[end - 1]
+    Δz_surf = zf_vec[2] - zf_vec[1]
+    z0 = zf_vec[1]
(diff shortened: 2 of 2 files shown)


trontrytel (approved):

charleskawczynski (changes requested):
Can we instead do this somewhat more elegantly by moving this to a PR in ClimaCore? Looking at what we're doing, it seems like a somewhat acceptable interface would be

```julia
     z_mesh = ClimaCore.Meshes.TruncatedIntervalMesh(domain, stretch; nelems = nzₛ, z_max = z₁)
```

We can do this by:
 - Inlining `adjust_mesh` into `TCMeshFromGCMMesh`, renaming `TCMeshFromGCMMesh` to `TruncatedIntervalMesh`

Something like 

```julia
function TruncatedIntervalMesh(::IntervalMesh, stretch; nelems, z_max)
...
end
```

My concern is that this workflow now involves creating, from what I can see, 3 domains:
 - The domain used to create `gcm_mesh`
 - The domain in `TCMeshFromGCMMesh`
 - The domain `new_domain`

We really need to (at least try to) move towards the cleaner pattern of "do things once". And this PR piles on more adjusting things that we didn't do quite correctly the first time, all using ClimaCore and TC code when this really should just be a simple user-facing function in ClimaCore.

ilopezgp:
Sure I can move this to CC, but renaming functions doesn't mean I won't be creating 3 meshes.

yairchn (approved):
LGTM I am fine with it being here , I don't think anyone else will be using this so maybe that is an argument for not moving it to CC. I am fine either way

ilopezgp:
I will open a PR in CC, but I would strongly prefer if this gets merged today wherever it is. I need to run some simulations over the weekend.

charleskawczynski:
> I need to run some simulations over the weekend.

If that's the case, then I'm fine merging this functionality in here, but can we at least make it portable to ClimaCore? The main thing I'd like to avoid is using TC's grid when constructing the space. If you need help inspecting, and becoming more familiar with, ClimaCore data structures, we can look through things interactively together.

ilopezgp:
> > I need to run some simulations over the weekend.
> 
> If that's the case, then I'm fine merging this functionality in here, but can we at least make it portable to ClimaCore? The main thing I'd like to avoid is using TC's grid when constructing the space. If you need help inspecting, and becoming more familiar with, ClimaCore data structures, we can look through things interactively together.

I will give it a try for the next hour, I agree that we should not need the grid or TC functionality. I just have to rewrite the logic.

ilopezgp:
See the alternative ClimaCore [PR](https://github.com/CliMA/ClimaCore.jl/pull/821).

ilopezgp:
See #1240, this is merged as a temporary solution to enable faster integration of TC main with CEDMF.

charleskawczynski (approved):

PR #1235, opened 2022-07-22
testing if this Plots update fixes our contour plots

trontrytel:
I'm trying to see if [this](https://github.com/JuliaPlots/Plots.jl/pull/4275) branch from Plots would fix [this](https://github.com/CliMA/TurbulenceConvection.jl/pull/1222) problem.

PR #1243, opened 2022-07-23
Upgrade deps

charleskawczynski:
Upgrade to the latest deps.

Supersedes #1235 and #1222.


trontrytel:
Nice contour plots are back :)

PR #1244, opened 2022-07-24
Add 2-moment, prescribed Nd autoconversions and accretions

trontrytel:
Co-authored-by: lynnyang <lynnyang@caltech.edu>

# PULL REQUEST

## Purpose and Content

This PR makes use of 2-moment, prescribed cloud droplet number concentration microphysics schemes in the new release of CloudMicrophysics.jl. On the TC.jl side we are creating a rain model that accepts the prescribed droplet number concentration as argument. The default scheme is unchanged. 

The new keyword options are:
`--rain_formation_scheme` (valid options: `clima_1m_default`, `KK2000`, `B1994`, `TC1980`, `LD2004`)
`--prescribed_Nd` (number of droplets per m3)

See the 2-moment CloudMicrophysics.jl [docs](https://clima.github.io/CloudMicrophysics.jl/dev/Microphysics2M/) for details. 

## Benefits and Risks
We will now have more microphysics options. (Uncertain if its a benefit or risk).

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -399,6 +399,8 @@ function Rico(namelist_defaults)
     namelist["forcing"]["coriolis"] = 4.5e-5
 
     namelist["microphysics"]["precipitation_model"] = "clima_1m"
+    namelist["microphysics"]["rain_formation_scheme"] = "clima_1m_default"
+    namelist["microphysics"]["prescribed_Nd"] = 1e8
     namelist["microphysics"]["precip_fraction_model"] = "prescribed" # "prescribed" or "cloud_cover"
     namelist["microphysics"]["prescribed_precip_frac_value"] = 1.0
     namelist["microphysics"]["precip_fraction_limiter"] = 0.3
@@ -440,6 +442,8 @@ function TRMM_LBA(namelist_defaults)
     namelist["time_stepping"]["dt_min"] = 1.0
 
     namelist["microphysics"]["precipitation_model"] = "clima_1m" # "cutoff"
+    namelist["microphysics"]["rain_formation_scheme"] = "clima_1m_default"
+    namelist["microphysics"]["prescribed_Nd"] = 1e8
     namelist["microphysics"]["precip_fraction_model"] = "prescribed" # "prescribed" or "cloud_cover"
     namelist["microphysics"]["prescribed_precip_frac_value"] = 1.0
     namelist["microphysics"]["precip_fraction_limiter"] = 0.3
@@ -537,6 +541,8 @@ function DYCOMS_RF02(namelist_defaults)
     namelist["time_stepping"]["dt_min"] = 1.0
 
     namelist["microphysics"]["precipitation_model"] = "clima_1m" #"cutoff"
+    namelist["microphysics"]["rain_formation_scheme"] = "clima_1m_default"
+    namelist["microphysics"]["prescribed_Nd"] = 1e8
     namelist["microphysics"]["precip_fraction_model"] = "prescribed" # "prescribed" or "cloud_cover"
     namelist["microphysics"]["prescribed_precip_frac_value"] = 1.0
     namelist["microphysics"]["precip_fraction_limiter"] = 0.3
--- a/src/microphysics_coupling.jl
+++ b/src/microphysics_coupling.jl
@@ -47,6 +47,7 @@ Computes the tendencies to qt and θ_liq_ice due to precipitation formation
 function precipitation_formation(
     param_set::APS,
     precip_model::AbstractPrecipitationModel,
+    rain_formation_model::AbstractRainFormationModel,
     qr::FT,
     qs::FT,
     area::FT,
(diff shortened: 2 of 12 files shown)


src/microphysics_coupling.jl
```diff
@@ -108,17 +109,63 @@ function precipitation_formation(
...
+                    CM2.conv_q_liq_to_q_rai_TC1980(microphys_params, q.liq, ρ, N_d = rain_formation_model.prescribed_Nd)
+                )
+            elseif rain_formation_model isa LD2004
+                S_qt_rain = -min(
+                    q.liq / Δt,
+                    CM2.conv_q_liq_to_q_rai_LD2004(microphys_params, q.liq, ρ, N_d = rain_formation_model.prescribed_Nd)
+                )
+            else
+                error("Unrecognized rain formation model")
+            end
```
charleskawczynski, line 136:
```suggestion
            S_qt_rain_0 = q.liq / Δt
            S_qt_rain_args = (microphys_params, q.liq, ρ, N_d = rain_formation_model.prescribed_Nd)
            S_qt_rain_1 = if rain_formation_model isa Clima1M_default
                CM1.conv_q_liq_to_q_rai(microphys_params, q.liq)
            elseif rain_formation_model isa KK2000
                CM2.conv_q_liq_to_q_rai_KK2000(S_qt_rain_args...)
            elseif rain_formation_model isa B1994
                CM2.conv_q_liq_to_q_rai_B1994(S_qt_rain_args...)
            elseif rain_formation_model isa TC1980
                CM2.conv_q_liq_to_q_rai_TC1980(S_qt_rain_args...)
            elseif rain_formation_model isa LD2004
                CM2.conv_q_liq_to_q_rai_LD2004(S_qt_rain_args...)
            else
                error("Unrecognized rain formation model")
            end
            S_qt_rain = -min(S_qt_rain_0, S_qt_rain_1)
```
Not sure about names here, but here's a bit of code golf 🏌🏻⛳. That said, it looks like `conv_q_liq_to_q_rai` could take different `rain_formation_model`s as an argument and just use dispatch. That way this could be written

```julia
            S_qt_rain_0 = q.liq / Δt
            S_qt_rain_args = (microphys_params, q.liq, ρ, N_d = rain_formation_model.prescribed_Nd)
            S_qt_rain_1 = CM1.conv_q_liq_to_q_rai(rain_formation_model, S_qt_rain_args...)
            S_qt_rain = -min(S_qt_rain_0, S_qt_rain_1)
```
And different `rain_formation_model` could hold potentially different arguments if needed

src/microphysics_coupling.jl
```diff
@@ -108,17 +109,63 @@ function precipitation_formation(
...
+            elseif rain_formation_model isa TC1980
+                S_qr = min(
+                    q.liq / Δt,
+                    CM2.accretion_TC1980(microphys_params, q.liq, qr)
+                ) * precip_fraction
+            elseif rain_formation_model isa LD2004
+                S_qr = min(q.liq / Δt, CM1.accretion(microphys_params, liq_type, rain_type, q.liq, qr, ρ)) * precip_fraction
+            else
+                error("Unrecognized rain formation model")
+            end
```
charleskawczynski, line 168: I invite you to try code golfing this 🏌🏻⛳   🙂

driver/generate_namelist.jl
```diff
@@ -399,6 +399,8 @@ function Rico(namelist_defaults)
     namelist["forcing"]["coriolis"] = 4.5e-5
 
     namelist["microphysics"]["precipitation_model"] = "clima_1m"
+    namelist["microphysics"]["rain_formation_scheme"] = "clima_1m_default"
+    namelist["microphysics"]["prescribed_Nd"] = 1e8
```
charleskawczynski, line 403: If we're planning on tuning this, then should it go into clima parameters instead?

src/microphysics_coupling.jl
trontrytel: All those `conv` functions are part of the `CloudMicrophysics.jl` library. I think for the simplicity sake I want to keep them as separate functions. - I don't want to turn CM.jl into Thermodynamics.jl when it comes to code complexity. 

If you want I can write a wrapper for this on the TC.jl side. But I think it will look similar to this code?

driver/generate_namelist.jl
trontrytel: You mean they should also be added to `src/Parameters.jl`?

src/microphysics_coupling.jl
trontrytel: Also, they are part of two different modules - `CM1` (for 1-moment) and `CM2` (for 2 moment).

charleskawczynski: Here's a [thread](https://discourse.julialang.org/t/multiple-dispatch-and-modules/59700/4) I think we may find useful for redesigning things, let me know what you think

driver/generate_namelist.jl
charleskawczynski: The toml file, and be read from the toml dict into a struct, instead of the namelist.

src/microphysics_coupling.jl
trontrytel: I can see how to dispatch over all functions defined in `CM2` module. But I still have a case where another variant of this function is defined in `CM1` module of `CloudMicrophysics.jl`. I think that trying to dispatch over functions from two different modules will not work?

```
julia> module Foo
       function bar(a) 
       return 0
       end
       end

julia> module Bar
       function bar(a, b)
       return 1
       end
       end

julia> using Main.Bar: bar

julia> using Main.Foo: bar
WARNING: ignoring conflicting import of Foo.bar into Main
```

trontrytel: I'll add dispatching over all the functions from `CM2` module. This way we will have only one `if/else` and more dispatch?

driver/generate_namelist.jl
trontrytel: Could we add the toml file to TC.jl repo. So that if I need to add/rename parameters I can do it as a PR to this repository and not CLIMAParmaters?

src/microphysics_coupling.jl
trontrytel: I have opened an issue about refactoring the 2-moment module inside CloudMicrophysics.jl Once that is done, I'll fix things for TC.

driver/generate_namelist.jl
trontrytel: Seems like for now it's actually easier not to have them as part of the toml files. Will change it once we update our calibration pipeline to the new ways

charleskawczynski: Yeah, I guess for now, while calibration is still catching up

src/types.jl
```diff
@@ -257,6 +257,22 @@ struct NoPrecipitation <: AbstractPrecipitationModel end
...
+end
+struct B1994{FT} <: AbstractRainFormationModel
+    prescribed_Nd::FT
+end
+struct TC1980{FT} <: AbstractRainFormationModel
+    prescribed_Nd::FT
+end
+struct LD2004{FT} <: AbstractRainFormationModel
+    prescribed_Nd::FT
+end
```
ilopezgp, line 274: Can you add a doc string for each struct with the paper they refer to, and maybe a brief description if possible?

src/microphysics_coupling.jl
```diff
@@ -108,17 +109,82 @@ function precipitation_formation(
...
+                        q.liq / Δt,
+                        CM2.conv_q_liq_to_q_rai_KK2000(
+                            microphys_params,
+                            q.liq,
+                            ρ,
+                            N_d = rain_formation_model.prescribed_Nd,
+                        ),
+                    )
+            elseif rain_formation_model isa B1994
+                S_qt_rain =
```
ilopezgp, line 126: This is more of a comment on CM. It would be cleaner if the `rain_formation_model` holds a function `conv_q_liq_to_q_rai` so you can just do `rain_formation_model.conv_q_liq_to_q_rai(...)`.

src/microphysics_coupling.jl
```diff
@@ -108,17 +109,82 @@ function precipitation_formation(
...
+                    min(q.liq / Δt, CM1.accretion(microphys_params, liq_type, rain_type, q.liq, qr, ρ)) *
+                    precip_fraction
+            elseif rain_formation_model isa KK2000
+                S_qr = min(q.liq / Δt, CM2.accretion_KK2000(microphys_params, q.liq, qr, ρ)) * precip_fraction
+            elseif rain_formation_model isa B1994
+                S_qr = min(q.liq / Δt, CM2.accretion_B1994(microphys_params, q.liq, qr, ρ)) * precip_fraction
+            elseif rain_formation_model isa TC1980
+                S_qr = min(q.liq / Δt, CM2.accretion_TC1980(microphys_params, q.liq, qr)) * precip_fraction
+            elseif rain_formation_model isa LD2004
+                S_qr =
```
ilopezgp, line 182: Same here, the `rain_formation_model` can have an internal `accretion` function.

ilopezgp:
LGTM as a placeholder until https://github.com/CliMA/CloudMicrophysics.jl/issues/74 is resolved.

ilopezgp (approved):

PR #1245, opened 2022-07-24
Add microphysics scaling parameters for calibrations

trontrytel:
Co-authored-by: kcpham4 <kcpham@caltech.edu>

# PULL REQUEST

## Purpose and Content
This PR adds 5 new parameters to the model. They are responsible for scaling the intensity of different microphysics processes. We need them for sensitivity analysis done using the CEDMF grid search tool. 

## Benefits and Risks
More parameters...


integration_tests/cli_options.jl
```diff
@@ -84,6 +84,21 @@ function parse_commandline()
...
+        help = "Coefficient multiplying the accretion rate (used for calibrations)"
+        arg_type = Float64
+        "--evap_scaling"
+        help = "Coefficient multiplying the evaporation rate (used for calibrations)"
+        arg_type = Float64
+        "--depsub_scaling"
+        help = "Coefficient multiplying the deposition sublimation rate (used for calibrations)"
+        arg_type = Float64
+        "--melt_scaling"
+        help = "Coefficient multiplying the melting rate (used for calibrations)"
```
ilopezgp, line 100: Does the `(used for calibrations)` mean that you do not intend to use this scaling operationally? Only for sensitivity analysis?

src/microphysics_coupling.jl
```diff
@@ -162,12 +165,12 @@ function precipitation_formation(
...
             if T < T_fr
                 S_qs =
-                    min(qr / Δt, CM1.accretion_snow_rain(microphys_params, snow_type, rain_type, qs, qr, ρ)) *
+                    min(qr / Δt, α_accr * CM1.accretion_snow_rain(microphys_params, snow_type, rain_type, qs, qr, ρ)) *
                     precip_fraction *
                     precip_fraction
             else
                 S_qs =
-                    -min(qs / Δt, CM1.accretion_snow_rain(microphys_params, rain_type, snow_type, qr, qs, ρ)) *
+                    min(qs / Δt, α_accr * CM1.accretion_snow_rain(microphys_params, rain_type, snow_type, qr, qs, ρ)) *
```
ilopezgp, line 173: was this meant to change sign?

integration_tests/cli_options.jl
trontrytel: I think it's mostly for sensitivity analysis. I was wondering if we could also start using it for testing or debugging. Setting any of those to zero results in switching off a microphysics process. 

I think it's a good feature to have. I was intending to code it in a way where you could switch between different microphysics schemes, but still have the same "turning knobs" to switch on/off processes or run a sensitivity test to see how the matter.

src/microphysics_coupling.jl
trontrytel: Not at all. Thank you so much for catching this!

integration_tests/cli_options.jl
ilopezgp: I see, then I would probably delete the parenthetical remark. It seems to be something we want to keep permanently, at least in this package.

trontrytel: Sounds good

charleskawczynski (approved):
Code changes look fine to me.

integration_tests/overwrite_namelist.jl
```diff
@@ -22,6 +22,11 @@ function overwrite_namelist!(namelist, parsed_args)
     "precip_fraction_model"        => (nl, pa, key) -> (nl["microphysics"]["precip_fraction_model"] = pa[key]),
     "prescribed_precip_frac_value" => (nl, pa, key) -> (nl["microphysics"]["prescribed_precip_frac_value"] = pa[key]),
     "precip_fraction_limiter"      => (nl, pa, key) -> (nl["microphysics"]["precip_fraction_limiter"] = pa[key]),
+    "acnv_scaling"                 => (nl, pa, key) -> (nl["microphysics"]["microph_scaling_acnv"] = pa[key]),
+    "accr_scaling"                 => (nl, pa, key) -> (nl["microphysics"]["microph_scaling_accr"] = pa[key]),
+    "evap_scaling"                 => (nl, pa, key) -> (nl["microphysics"]["microph_scaling"] = pa[key]),
```
ilopezgp, line 27: why is this not `microph_scaling_evap`?

ilopezgp (commented):
There is a `microph_scaling` parameter that now seems ambiguous, given that there are different scalings per process. I would prefer to name this parameter more precisely.

integration_tests/overwrite_namelist.jl
trontrytel: I wanted to fix that, but it requires changing a name of a parameter inside CLIMAParameters and that means a breaking release for CLIMAParametes. So I was told not to do it.

At least the command line args are correct

trontrytel:
@JuliaRegistrator register

PR #1250, opened 2022-07-25
params overwrite

trontrytel:
# PULL REQUEST

## Purpose and Content
This PR adds back an option to overwrite microphysics parameters with a namelist file. This is needed to integrate with CEDMF.

## Questions
- I was trying to print the value of `E_liq_rai` parameter, which is part of `microphys_params` to test if I'm really overwriting them. But I could not figure out how to access it. Any suggestions here?

- This is a little clunky because now we define a default value in CLIMAParameters and in the overwrite file. 

- Some of those parameters would benefit from being specified via command line as well (thinking about the ones that switch on/off processes). Not sure where it fits here


charleskawczynski (commented):
Would it be easier if we just had an override toml file instead? cc @odunbar, @costachris

trontrytel:
> Would it be easier if we just had an override toml file instead? cc @odunbar, @costachris

I think that would be better in general. But I also think that right now CEDMF interacts with TC via namelists

trontrytel:
The differences in buildkite come from the fact that when the namelist files were still working we were overwriting the default autoconversion timescale to 2500s to match the value used in PyCLES simulations. Bu deleting those overwrites sometime in the past we have mistakenly switched back to using the default value from CLIMAParameters.

costachris (approved):

ilopezgp (approved):

PR #1253, opened 2022-07-27
add entr_ml_closure

yairchn:
Following the darpa meeting, this PR separates the physical and ml entrainment closures. There is a bit of a logic shenanigans in how to combine all the pieces without double counting any component of the entrainment, in order to disentangle it the PR:

1. removes the dMdz entrainment term that was used only in the dry bubble (and is not needed anymore - I suspect that in scampy it was needed due to in consistent timestepping method). 
2. moves the turbulent entrainment to a separate call. 
3. moves the area limiter into the non dimensional MD function.


src/types.jl
```diff
@@ -59,13 +77,32 @@ Base.@kwdef struct εδModelParams{FT, AFT}
...
+    μ_0::FT # dimensional scale logistic function in the dry term in entrainment/detrainment
+    β::FT # sorting power for ad-hoc moisture detrainment function
+    χ::FT # fraction of updraft air for buoyancy mixing in entrainment/detrainment (0≤χ≤1)
+    c_λ::FT # scaling factor for TKE in entrainment scale calculations
+    γ_lim::FT
+    β_lim::FT
+    c_γ::FT # scaling factor for turbulent entrainment rate
+    c_δ::FT # factor multiplier for moist term in entrainment/detrainment
+    Π_norm::AFT
+end
```
ilopezgp, line 93: Is this struct needed? It seems identical to `εδModelParams{FT, AFT}`.

src/types.jl
```diff
@@ -553,28 +600,39 @@ function EDMFModel(::Type{FT}, namelist, precip_model) where {FT}
 
     mean_entr_closure = if entr_type == "moisture_deficit"
         MDEntr(; params = εδ_params)
-    elseif entr_type == "NN"
+    elseif entr_type == "None"
+        c_nn_params = parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "nn_ent_params")
+        nn_arc = parse_namelist(namelist, "turbulence", "EDMF_PrognosticTKE", "nn_arc")
+        NNEntr(; params = εδ_params, biases_bool = nn_biases, c_nn_params, nn_arc)
```
ilopezgp, line 606: This seems to be wrong, you are constructing a Neural network when you want a `EntrNone(; params = εδ_params)` model.

src/closures/entr_detr.jl
```diff
@@ -263,6 +266,7 @@ function compute_entr_detr!(
     surf::SurfaceBase,
     Δt::Real,
     εδ_model::AbstractNonLocalEntrDetrModel,
```
costachris, line 268: `εδ_model` is redundant now and no longer needed for dispatching since only ML models (`εδ_model_ml`) can be nonlocal, so we can remove `εδ_model` from the function signature and just unpack it from the `edmf` object within the function for both instances of `compute_entr_detr!`.

src/closures/entr_detr.jl
```diff
@@ -350,7 +354,10 @@ function compute_entr_detr!(
         Π_groups = parent(aux_up[i].Π_groups)
         ε_nondim = parent(aux_up[i].ε_nondim)
         δ_nondim = parent(aux_up[i].δ_nondim)
+        ε_nondim_ml = copy(ε_nondim)
+        δ_nondim_ml = copy(δ_nondim)
```
ilopezgp, line 358:
```suggestion
        ε_nondim_ml = deepcopy(ε_nondim)
        δ_nondim_ml = deepcopy(δ_nondim)
```

ilopezgp (approved):
Everything else LGTM. Can we have someone else from the DARPA team review this before merging?

src/closures/entr_detr.jl
```diff
@@ -350,7 +354,10 @@ function compute_entr_detr!(
         Π_groups = parent(aux_up[i].Π_groups)
         ε_nondim = parent(aux_up[i].ε_nondim)
         δ_nondim = parent(aux_up[i].δ_nondim)
+        ε_nondim_ml = deepcopy(ε_nondim)
```
costachris, line 357: I'm not sure how `deepcopy` works with climacore fields or if we need to make 2 new `cent_aux_vars_up` fields. Does `parent(aux_up[i].ε_nondim)` just return a normal array?

costachris:
I'm getting method errors for certain combinations involving the nonlocal closures (ie. `"entrainment"= "None"` , `"entrainment_ml" = "NN_nonlocal"`). Looking into why...

src/closures/entr_detr.jl
ilopezgp: Good point, it would be good to get @charleskawczynski to check it out. We are trying to avoid modifying in-place the same object twice later on.

src/closures/entr_detr.jl
```diff
@@ -283,14 +277,117 @@ function compute_entr_detr!(
...
+
+function compute_ml_entr_detr!(
+    state::State,
+    grid::Grid,
+    edmf::EDMFModel,
+    param_set::APS,
+    surf::SurfaceBase,
+    Δt::Real,
+    εδ_closure::AbstractMLEntrDetrModel,
+)
```
charleskawczynski, line 301: This is the only part of this PR where I'd suggest to change. Can we somehow reuse the existing function? It's not a huge deal, but this does duplicate quite a bit of code

src/closures/entr_detr.jl
```diff
@@ -262,17 +253,20 @@ function compute_entr_detr!(
...
-                    er = entr_detr(mean_model, εδ_model_vars, edmf.entr_dim_scale, edmf.detr_dim_scale)
+                    ε_nondim, δ_nondim = non_dimensional_function(mean_model, εδ_model_vars)
+                    ε_dyn, δ_dyn =
+                        εδ_dyn(mean_model, εδ_model_vars, edmf.entr_dim_scale, edmf.detr_dim_scale, ε_nondim, δ_nondim)
                 else
                     # fractional, turbulent & nondimensional entrainment
-                    er = entr_detr(εδ_closure, εδ_model_vars, edmf.entr_dim_scale, edmf.detr_dim_scale)
-                    ε_dyn, δ_dyn = er.ε_dyn, er.δ_dyn
+                    ε_nondim, δ_nondim = non_dimensional_function(εδ_closure, εδ_model_vars)
+                    ε_dyn, δ_dyn =
```
yairchn, line 262: I think here `edmf.entr_dim_scale` should be hard coded to `b_w2` as it's the only scale we have so far for a physical model

src/closures/entr_detr.jl
yairchn: Maybe we can leave this to a future PR and I'll try to do that when I am back from leave ?

src/types.jl
```diff
@@ -574,34 +600,42 @@ function EDMFModel(::Type{FT}, namelist, precip_model, rain_formation_model) whe
...
     mean_entr_closure = if entr_type == "moisture_deficit"
         MDEntr(; params = εδ_params)
-    elseif entr_type == "NN"
+    elseif entr_type == "None"
+        EntrNone(; params = εδ_params)
+    else
+        error("Something went wrong. Invalid entrainment type '$entr_type'")
+    end
+
+    ml_entr_closure = if ml_entr_type == "None"
```
charleskawczynski, line 615: What's happening to `entr_closure`?

yairchn: I think its defined in 642 based on the Noisy process right ?

costachris (approved):

yairchn:
Bors r+

PR #1261, opened 2022-08-09
Upgrade deps

charleskawczynski:
This PR upgrades our deps.


charleskawczynski:
Hmm, github seems to be having some issues..

charleskawczynski:
Oof, this somehow doubles allocations in the init in life cycle (I haven't checked others yet).

PR #1264, opened 2022-08-14
up deps after CMv0.8.0

trontrytel:


charleskawczynski (approved):

trontrytel:
I'm getting `The command exited with status 137` error on a different test case each time I try to merge. This is an out of memory error. Not sure what to do here.

Issue #908, opened 2022-04-04
Flatten namelist


charleskawczynski:
Based on where it seems parameter calibration design is moving, and writing more general-purpose arg parsing for cli options, it seems like flattening the namelist would provide several benefits. Can we discuss this a bit here?

cc @yairchn, @trontrytel, @ilopezgp, @odunbar, @costachris, @haakon-e

trontrytel:
@charleskawczynski we will start with this one but be working on those issues in ClimaAtmos TC not here. I'll open an issue there.  After talking to @dennisYatunin we will try to make it broader and try to improve our handling of TC arguments:
- introduce some hierarchy via arg parsing
- try to define the defaults only in one place

PR #1265, opened 2022-08-17
fix: LES_driven_SCM & DryBubble default namelist

haakon-e:
fixes bugs introduced in #1251 and #34

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -608,11 +608,11 @@ end
 function DryBubble(namelist_defaults)
     namelist = deepcopy(namelist_defaults)
     namelist["meta"]["casename"] = "DryBubble"
-    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["surface_area"] = 0.0
+    namelist["turbulence"]["EDMF_PrognosticTKE"]["surface_area"] = 0.0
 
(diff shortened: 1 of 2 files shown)


charleskawczynski (approved):
Ouch, this looks like it may have taken a while to track down. How is #34 related?

If we fix #908, we'd be able to write a pretty nice / simple log to show users which variables have been changed.

haakon-e:
According to git blame, it was in #34 the changes (bug) in DryBubble was introduced.

haakon-e:
I agree we should get to #908 (alternatively by finishing up CliMA parameters integration). 

From the point of view of calibration, we are in fact assuming that all parameter names are unique, regardless of which part of the name list it is located.

charleskawczynski:
> According to git blame, it was in #34 the changes (bug) in DryBubble was introduced.

Ah, yeah, I missed the dry bubble one. Good catch!

ilopezgp (approved):
LGTM

ilopezgp:
Should we merge @haakon-e ?

haakon-e:
> Should we merge @haakon-e ?

Yes. I'm travelling currently so haven't been able to get around to fix the deps, etc. but can try to get around to this soon

ilopezgp:
@haakon-e just a reminder!

ilopezgp (approved):

PR #1271, opened 2022-08-22
Upgrade to julia 1.8

charleskawczynski:
Upgrade to julia 1.8


charleskawczynski:
Yikes, Soares completely breaks (`DomainError`) when only upgrading to julia 1.8

yairchn:
> Yikes, some cases are having serious issues when only upgrading to julia 1.8:
> 
> * Soares (breaks with `DomainError`)
> * Nieuwstadt (breaks with `DomainError`)
> * GABLS (not breaking but _very_ different looking results)
> 
> @yairchn, any commonalities that you see between these cases? Any hints will help track down what's going off the rails here

are you also upgrading surface fluxes? there is an issue there I'm trying to solve

charleskawczynski:
> are you also upgrading surface fluxes? there is an issue there I'm trying to solve

Ah! good catch!

PR #1273, opened 2022-08-23
limit the SF version in compact in Project

yairchn:
limit the SF.jl version for now until we solve so me issues in SF.jl


charleskawczynski (commented):
I think we need something slightly different. I'm hoping that we can fix compatibility and avoid needing to pin SF

PR #1275, opened 2022-08-24
Pin SciMLBase

charleskawczynski:
This PR pins SciMLBase, which seems to have an issue. cc @kmdeck


kmdeck:
> This PR pins SciMLBase, which seems to have an issue. cc @kmdeck

amazing, thank you! The version of DiffEqBase then doesnt really matter (it goes for 6.95, but that works now).

charleskawczynski:
Hmm, it seems like I need to `(TurbulenceConvection) pkg> add SciMLBase@1.48.1` in order for our release to go through (I guess they try precompiling the package in the default env, too)

PR #1282, opened 2022-09-12
update to julia 1.8.1

trontrytel:
Update to julia 1.8.1


docs/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.0"
+julia_version = "1.7.3"
```
charleskawczynski, line 3: Looks like this was generated with Julia 1.7.3?

trontrytel: oh no!

trontrytel: I thought I got rid of that julia... Where is it coming from @#$%^&!!!

trontrytel: Should be fixed now

charleskawczynski (approved):

ilopezgp:
I think we need to clear the depot

trontrytel:
Some TC simulations were crashing. I'm trying to see if using SurfaceFluxes pinned to v0.4.3 will help

docs/Manifest.toml
```diff
@@ -1606,9 +1617,9 @@ version = "5.10.1+0"
 
 [[deps.SurfaceFluxes]]
 deps = ["DocStringExtensions", "KernelAbstractions", "RootSolvers", "StaticArrays", "Thermodynamics"]
-git-tree-sha1 = "a2271e67a53379a1547ddb4846a47e5e3248ab65"
+git-tree-sha1 = "7a22365358252d9b9c0f423ed91377abce154290"
 uuid = "49b00bb7-8bd4-4f2b-b78c-51cd0450215f"
-version = "0.4.3"
+version = "0.4.5"
```
charleskawczynski, line 1622: Can we just `] resolve` instead of `] up`?

trontrytel: I thought we are supposed to use the `up_deps.jl` for those sorts of things

trontrytel: You mean I should change the Project.toml files and then just do `resolve` instead of `up_deps`?

Project.toml
```diff
@@ -37,7 +37,7 @@ SciMLBase = "~1.48.1"
 StaticArrays = "1.2"
 StatsBase = "0.33"
 StochasticDiffEq = "6.41"
-SurfaceFluxes = "0.4"
+SurfaceFluxes = "= 0.4.3"
```
charleskawczynski, line 40: I'm actually not sure if this is okay off the top of my head. Ref: [this comment](https://github.com/JuliaRegistries/General/pull/67876#issuecomment-1240105329)

trontrytel: Thats great. So it's not compatible with registering things. It is compatible with reality though - v0.4.5 does not work for us right now :)

charleskawczynski: Yeah, agreed. An alternative view is that the patch version was breaking for us, and perhaps should have been a minor (breaking) release

Project.toml
```diff
@@ -37,7 +37,7 @@ SciMLBase = "~1.48.1"
 StaticArrays = "1.2"
 StatsBase = "0.33"
 StochasticDiffEq = "6.41"
-SurfaceFluxes = "= 0.4.3"
+SurfaceFluxes = "0.4.3"
```
charleskawczynski, line 40:
```suggestion
SurfaceFluxes = "~0.4.3"
```

docs/Manifest.toml
```diff
@@ -1697,7 +1708,7 @@ version = "0.3.4"
 
 [[deps.TurbulenceConvection]]
 deps = ["ClimaCore", "CloudMicrophysics", "Dierckx", "Distributions", "DocStringExtensions", "FastGaussQuadrature", "Flux", "LambertW", "LinearAlgebra", "OperatorFlux", "Random", "SciMLBase", "StaticArrays", "StatsBase", "StochasticDiffEq", "SurfaceFluxes", "Thermodynamics", "UnPack"]
-path = ".."
```
charleskawczynski, line 1700: Revert!

PR #1286, opened 2022-09-13
Add pkg manager script

charleskawczynski:
This PR adds a script, similar to `up_deps.jl`, which automatically adds / removes a specified package (and version) to avoid needing to update package versions.


trontrytel:
@charleskawczynski looks like something related to Aqua test is breaking this

PR #1293, opened 2022-09-19
yc/update_sf

yairchn:
after solving issues with tolerance in SF.jl we can update TC


Project.toml
```diff
@@ -37,7 +37,7 @@ SciMLBase = "1"
 StaticArrays = "1.2"
 StatsBase = "0.33"
 StochasticDiffEq = "6.41"
-SurfaceFluxes = "0.4.2 - 0.4.3"
+SurfaceFluxes = "0.4"
```
charleskawczynski, line 40: This needs to be applied to the other environments, we'll also need to `julia .dev/up_deps.jl` after the compats are adjusted to upgrade SF to the latest version

charleskawczynski (approved):

yairchn:
Bors r+

PR #1298, opened 2022-09-30
Nudge to GCM profiles, not equilibrium LES profiles.

ilopezgp:
# PULL REQUEST

## Purpose and Content

Fix the relaxation forcing so that it allows reaching the same steady state dynamical balance as LES.

Right now we are using as relaxation target the equilibrium LES profiles. As can be seen from the second dynamical balance in the picture, this is wrong since the momentum relaxation term is the only term balancing the vertical SGS momentum flux: it would only allow matching LES if the flux is zero.

What we need to use is the GCM profile; we can obtain this as the initial LES profile. I expect this to be an important difference near the surface where GCM velocities are far from the LES ones. For the thermodynamic forcing this is likely an extremely small correction in the free troposphere, so it shouldn't matter much there.

PXL_20220930_175442860 MP_2


charleskawczynski (approved):
Code changes look fine to me

szy21 (approved):

ilopezgp:
bors try

ilopezgp:
bors try

PR #1301, opened 2022-10-04
Update ArtifactWrappers, depependencies

ilopezgp:
# PULL REQUEST

## Purpose and Content

Updates ArtifactWrappers after breaking changes, which allows updating the same package in CalibrateEDMF.jl.


charleskawczynski (approved):
Thanks! I never quite finished propagating those changes up the stack!

PR #1313, opened 2022-10-11
Oh Calculus, my dear old friend.

ilopezgp:
This PR computes the average over two percentiles of a Gaussian analytically, resulting in perfect precision and a 3 order of magnitude reduction in the cost of computation versus the current method. 

This should probably be ported over ClimaAtmos as well.


src/utility_functions.jl
```diff
@@ -22,6 +21,15 @@ function percentile_bounds_mean_norm(
     return ∑samples / n_filtered_samples
 end
 
+"Compute average between two percentiles of a standard Gaussian"
```
charleskawczynski, line 24:
```suggestion
"""Compute average between two percentiles of a standard Gaussian"""
```

src/utility_functions.jl
```diff
@@ -1,5 +1,4 @@
-# compute the mean of the values between two percentiles (0 to 1) for a standard normal distribution
-# this gives the surface scalar coefficients for 1 to n-1 updrafts when using n updrafts
+"Compute the mean of the values between two percentiles (0 to 1) by sampling"
```
charleskawczynski, line 1:
```suggestion
"""Compute the mean of the values between two percentiles (0 to 1) by sampling"""
```

charleskawczynski (approved):
Love it. Can we clean up the parameters that we no longer need? (`n_samples` and `set_src_seed`).

Overall, approve this ∫

ilopezgp:
> Love it. Can we clean up the parameters that we no longer need? (`n_samples` and `set_src_seed`).
> 
> Overall, approve this ∫

I can do that. Should I leave the former method, or remove it as well?

charleskawczynski:
I’m fine with removing it!

PR #1314, opened 2022-10-18
Add back supersat and add in Socrates test case template for RF09

jbphyswx:


jbphyswx:
Still crashes in time though from spurious instabilities even with very small timestep...
Maybe using their vertical grid has something to do with that? Not sure... 
Changing the nudging timescales doesn't fix the issue...

Also the updrafts right now are very meager (small area) and don't match their LWCs so perhaps that's a surface flux thing?

trontrytel:
Looks like things are out of date a bit here. Let's try to bring it to a state where it can be merged. I suggest splitting this PR into two PRs (1) that adds the SOCRATES test cases. We can initially run them with our default microphysics. (2) that adds the different non-equilibrium microphysics options you are testing. This will make it easier to review.

trontrytel:
And a follow up comment after todays meeting. When adding the SOCRATES test case please add a list of different forcings that need to be applied with a short description of what the code is doing about each of them (either in the PR description or on overleaf)

PR #1318, opened 2022-12-15
Fix bottom boundary condition for entropy flux

costachris:
Applies bottom boundary condition for diffusive `s` flux in a manner analogous to `theta_li` and `q_t` ([here](https://github.com/CliMA/TurbulenceConvection.jl/blob/2bfeb1b543acf6271a00348ba50e0f21456d9306/src/EDMF_functions.jl#L207)),  based on divergence at surface. Previously, the divergence for `s` was set to 0 at the surface resulting in ~0 diagnosed `s` flux. Conversion of `theta_li` and `q_t` surface fluxes to `s` adapted from [relationship](https://github.com/CliMA/pycles/blob/edfd57156711ec37e1f47972bfc7ba53277c3487/Csrc/surface.h#L95) in Pycles.


costachris:
`LES_driven_SCM` case:
## Previous 
s_flux_default_final-2
## PR
s_flux_corrected_final

driver/Surface.jl
```diff
@@ -2,6 +2,35 @@ import StaticArrays as SA
...
 
+
+"""
+    surface_entropy_flux_from_thetali_q_tot_flux(
+        thermo_params,
+        ρθ_liq_ice_flux::FT,
+        ρq_tot_flux::FT,
+        ts::TD.PhaseEquil,
+    )
+    Computes entropy flux given fluxes of thetali and q_tot at surface.
```
ilopezgp, line 13: Can you point to the pycles function here for future reference? https://github.com/CliMA/pycles/blob/edfd57156711ec37e1f47972bfc7ba53277c3487/Csrc/surface.h#L95

driver/Surface.jl
```diff
@@ -2,6 +2,35 @@ import StaticArrays as SA
...
+    ρθ_liq_ice_flux::FT,
+    ρq_tot_flux::FT,
+    ts::Union{TD.PhaseEquil{FT}, TD.PhaseNonEquil{FT}},
+) where {FT}
+    cp_m = TD.cp_m(thermo_params, ts)
+    T = TD.air_temperature(thermo_params, ts)
+    q = TD.PhasePartition(thermo_params, ts)
+    p = TD.air_pressure(thermo_params, ts)
+    s_d = TD.specific_entropy_dry(thermo_params, p, T, q)
+    s_v = TD.specific_entropy_vapor(thermo_params, p, T, q)
```
ilopezgp, line 27: These functions accept different pressures in the pycles formulation, but here you are passing the air pressure to both.

driver/Surface.jl
```diff
@@ -2,6 +2,35 @@ import StaticArrays as SA
...
+    surface_entropy_flux_from_thetali_q_tot_flux(
+        thermo_params,
+        ρθ_liq_ice_flux::FT,
+        ρq_tot_flux::FT,
+        ts::TD.PhaseEquil,
+    )
+    Computes entropy flux given fluxes of thetali and q_tot at surface.
+"""
+
+function surface_entropy_flux_from_thetali_q_tot_flux(
```
ilopezgp, line 16: Is there anything specific to the surface in this function definition? If not, we can omit the initial `surface_` and have it as a more general function.

driver/compute_diagnostics.jl
```diff
@@ -169,11 +172,12 @@ function compute_diagnostics!(
...
     end
 
-    # TODO(ilopezgp): Fix bottom gradient
     wvec = CC.Geometry.WVector
     m_bcs = (; bottom = CCO.SetValue(FT(0)), top = CCO.SetValue(FT(0)))
-    # ∇0_bcs = (; bottom = CCO.SetDivergence(wvec(FT(0))), top = CCO.SetDivergence(wvec(FT(0))))
-    ∇0_bcs = (; bottom = CCO.SetDivergence(FT(0)), top = CCO.SetDivergence(FT(0)))
+
+    aeKHs_bc = -surf.ρs_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KH[kf_surf]
+    ∇0_bcs = (; bottom = CCO.SetDivergence(FT(aeKHs_bc)), top = CCO.SetDivergence(FT(0)))
```
ilopezgp, line 179: can you inline the BCs below in the definition of `∇f`? We mat also want to rename `∇f` to `∇s_en` since now it is not just a standard gradient operator, but specific to entropy BCs.

ilopezgp (approved):
LGTM when the comments are addressed!

driver/Surface.jl
costachris: Good point - I double checked and the TD specific entropy functions compute the appropriate dry/vapor partial pressures internally (from the passed p, T, q arguments) whereas the pycles functions accept dry partial pressure to compute dry entropy and vapor partial pressure to compute specific entropy w.r.t vapor. So the calculation is the same, but happens under the hood in the TD `specific_entropy_` functions.

ilopezgp (approved):
LGTM

PR #1321, opened 2023-01-28
add nn helper funcs and linear no-bias option

costachris:
## Purpose 
- [x] Add helper functions for initializing neural networks with random weights + serializing NN parameters
- [x] Add no-bias option for linear entrainment

### DIFF
--- a/driver/generate_namelist.jl
+++ b/driver/generate_namelist.jl
@@ -264,6 +264,8 @@ function default_namelist(
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["linear_ent_params"] =
         SA.SVector{14}(rand(14))
 
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["linear_ent_biases"] = true
+
     # General stochastic entrainment/detrainment parameters
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["general_stochastic_ent_params"] =
     SA.SVector(
--- a/src/closures/nondimensional_exchange_functions.jl
+++ b/src/closures/nondimensional_exchange_functions.jl
@@ -219,6 +219,62 @@ function construct_fully_connected_nn(
 end
 
 """
+serialize_ml_model(
+    ml_model::Flux.Chain,
+    )
+
+    Given Flux NN model, serialize model weights into a vector of parameters.
+    - `ml_model` - A Flux model instance.
+"""
+
+function serialize_ml_model(ml_model::Flux.Chain)
+    parameters = []
+    param_type = eltype.(Flux.params(ml_model))[1]
+    for layer_i in 1:length(ml_model)
+        params_layer_i = Flux.params(ml_model[layer_i]).order.data[1]
+        params_layer_i_flattened = reshape(params_layer_i, length(params_layer_i))
+        push!(parameters, params_layer_i_flattened...)
+    end
+    return convert(Array{param_type}, parameters)
+end
+
+"""
+    construct_fully_connected_nn_default(
+        arc::AbstractArray{Int};
+        biases_bool::Bool = false,
+        activation_function::Flux.Function = Flux.sigmoid,
+        output_layer_activation_function::Flux.Function = Flux.relu,
+    )
+
+Given network architecture, construct NN model with default `Flux.jl` weight initialization (glorot_uniform).
+- `arc` :: vector specifying network architecture
+- `biases_bool` :: bool specifying whether to include biases.
+- `activation_function` :: activation function for hidden layers
+- `output_layer_activation_function` :: activation function for output layer
+"""
+
+function construct_fully_connected_nn_default(
+    arc::AbstractArray{Int};
+    biases_bool::Bool = false,
+    activation_function::Flux.Function = Flux.sigmoid,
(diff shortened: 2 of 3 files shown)


driver/generate_namelist.jl
```diff
@@ -264,6 +264,8 @@ function default_namelist(
     namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["linear_ent_params"] =
         SA.SVector{14}(rand(14))
 
+    namelist_defaults["turbulence"]["EDMF_PrognosticTKE"]["linear_ent_biases"] = true
+
```
trontrytel, line 268: We are setting the new option to true by default here, but to false by default when we parse the namelist in src/types.jl?

src/types.jl
```diff
@@ -567,6 +568,15 @@ function EDMFModel(::Type{FT}, namelist, precip_model, rain_formation_model) whe
         valid_options = [true, false],
     )
 
+    linear_biases = parse_namelist(
+        namelist,
+        "turbulence",
+        "EDMF_PrognosticTKE",
+        "linear_ent_biases";
+        default = false,
+        valid_options = [true, false],
```
trontrytel, line 577: We are setting the new option to false by default here, but to true by default when we create the namelist in generate_namelist?

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -272,13 +328,31 @@ Uses a simple linear model to predict the non-dimensional components of dynamica
...
 
+    c_linear = εδ_model.c_linear
     nondim_groups = collect(non_dimensional_groups(εδ_model, εδ_model_vars))
+
     # Linear closure
     lin_arc = (length(nondim_groups), 1)  # (#inputs, #outputs)
-    lin_model_ε = Flux.Dense(reshape(c_linear[1:6], lin_arc[2], lin_arc[1]), [c_linear[7]], Flux.relu)
-    lin_model_δ = Flux.Dense(reshape(c_linear[8:13], lin_arc[2], lin_arc[1]), [c_linear[14]], Flux.relu)
+    if εδ_model.biases_bool
+        @assert length(c_linear) == 2 * length(nondim_groups) + 2 "Incorrect number of parameters specified for linear closure"
```
trontrytel, line 338: Could this line be kicked outside of if/else statement?

src/types.jl
costachris: good catch,  I'll set it to true everywhere.

trontrytel:
Looks like CI is breaking. So as we discussed, I would suggest splitting the package updates to a separate PR. I'm sure they are responsible for the CI hiccups. 

As for the NN PR, I don't see anything wrong with it. But also I didn't work with edmf NN long enough to be a good reviewer. Two option I see are to a) meet sometime next week and explain to me what is happening and why do we need it, b) tag someone else that can better review that part. Option a) has a benefit of making me a better reviewer for the next time. Option b) is prolly faster

src/closures/nondimensional_exchange_functions.jl
costachris: The number of parameters depends on whether of not `biases_bool` is true (ie. more parameters if `true`), so the number of parameters is conditional on `biases_bool`

costachris:
@trontrytel Option B it is, I know just the person :) cc'ing @haakon-e

haakon-e:
are you requesting a review now, or making more changes first?

costachris:
Requesting a review now

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -218,6 +218,62 @@ function construct_fully_connected_nn(
...
+"""
+serialize_ml_model(
+    ml_model::Flux.Chain,
+    )
+
+    Given Flux NN model, serialize model weights into a vector of parameters.
+    - `ml_model` - A Flux model instance.
+"""
+
+function serialize_ml_model(ml_model::Flux.Chain)
```
haakon-e, line 230: is this used or tested anywhere?

src/closures/nondimensional_exchange_functions.jl
```diff
@@ -218,6 +218,62 @@ function construct_fully_connected_nn(
...
+    )
+
+Given network architecture, construct NN model with default `Flux.jl` weight initialization (glorot_uniform).
+- `arc` :: vector specifying network architecture
+- `biases_bool` :: bool specifying whether to include biases.
+- `activation_function` :: activation function for hidden layers
+- `output_layer_activation_function` :: activation function for output layer
+"""
+
+function construct_fully_connected_nn_default(
```
haakon-e, line 256: is this also used anywhere?

haakon-e (approved):
Can't verify if it works as expected w.r.t. parameter indices, but looks good otherwise.

src/closures/nondimensional_exchange_functions.jl
costachris: It's a helper function for initializing weights in the calibration pipeline. All the other ML/ LES library helper functions are housed in TC, so I was following that pattern. These were tested by ensuring the outputs of `serialize_ml_model` `construct_fully_connected_nn_default` and `construct_fully_connected_nn` with random weights were self-consistent.

PR #1322, opened 2023-01-30
update to julia 1.8.3 and update CompatHelper dependencies

costachris:
## Purpose 
update to julia 1.8.3 and update CompatHelper dependencies


costachris:
Replaced by #1323

PR #1324, opened 2023-02-09
Add inv scale height as entrainment dimensional scale option

costachris:
## Purpose 
Add inverse atmospheric scale height as entrainment dimensional scale option.


haakon-e (approved):
assuming this a reasonable way to define scale height, LGTM

PR #1325, opened 2023-02-10
jb/add socrates test cases

jbphyswx:
## Content

----
- [ ] I have read and checked the items on the review checklist.


jbphyswx:
This won't build because there's no way to specify a registry in the toml and I used my registry at `https://github.com/jbphyswx/MyRegistry` to register `SOCRATES_Single_Column_Forcings`

Ideas to get around that welcome... I dont want to add all the code in that package to this repo.

trontrytel:
Hi @costachris and @charleskawczynski - could you take a look here?

@jbphyswx is adding test cases that are based on SOCRATES papers. The test cases need profiles not only for initial conditions but also for forcings during the simulations. I know that for LES driven SCMs we just store the relevant data online in a box. But alternatively, we also have the [Atmospheric Profiles Library repository](https://github.com/CliMA/AtmosphericProfilesLibrary.jl), where we store the initial conditions from different paper based cases. 

Would you suggest following the LES driven SCM cases setup? Or could we try to extend the APL.jl to include the SOCRATES data? @jbphyswx so far put the data in a separate [repo](https://github.com/jbphyswx/SOCRATES_Single_Column_Forcings) but we would probably like to avoid adding one more repo to the stack.

Project.toml
```diff
@@ -37,7 +37,7 @@ SciMLBase = "1"
 StaticArrays = "1.2"
 StatsBase = "0.33"
 StochasticDiffEq = "6.41"
-SurfaceFluxes = "0.4"
+SurfaceFluxes = "0.4, 0.5, 0.6"
```
trontrytel, line 40: I would rather keep all the changes to dependencies in a separate PR. Lets clean this up when we decide how we want to store the forcing data

trontrytel:
@jbphyswx - could you add in the PR description the list of forcings that are defined for the SOCRATES cases?

driver/Cases.jl
```diff
@@ -78,15 +78,54 @@ struct DryBubble <: AbstractCaseType end
...
 
+abstract type ForcingSOCRATES end 
+abstract type SOCRATES <: AbstractCaseType end # this is the abstract type for all SOCRATES cases, cause you can't subtype non abstract types in julia
+for flight_number in SOCRATES_Single_Column_Forcings.flight_numbers
+    for forcing_type in SOCRATES_Single_Column_Forcings.forcing_types
+        name = Symbol("ForcingSOCRATES_RF$(string(flight_number,pad=2))_$(forcing_type)") # separated this way cause nested dollar signs gave me too much trouble... (invalid type signature) but allows for easier addition of methods
+        @eval struct $name <: ForcingSOCRATES end
+    end
+end
+
```
trontrytel, line 128: Should we be also adding some sort of radiation scheme for SOCRATES cases?

integration_tests/Manifest.toml
```diff
@@ -2,7 +2,7 @@
 
 julia_version = "1.8.5"
 manifest_format = "2.0"
-project_hash = "2e768be259e59f92bc16417b448cc41c0ab4d5b5"
+project_hash = "2a90d6037a0144dcaf273ca04235c82d43edb646"
```
trontrytel, line 5: Lets do toml file related changes separately (after we decide where to store the SOCRATES data)

driver/generate_namelist.jl
```diff
@@ -682,6 +684,48 @@ function LES_driven_SCM(namelist_defaults)
...
+    # The name will be first fixed in CLIMAParameters.
+    namelist["microphysics"]["microph_scaling"] = 1.0
+    namelist["microphysics"]["microph_scaling_dep_sub"] = 1.0
+    namelist["microphysics"]["microph_scaling_melt"] = 1.0
+    namelist["microphysics"]["microph_scaling_acnv"] = 1.0
+    namelist["microphysics"]["microph_scaling_accr"] = 1.0
+
+    # casename for data
+    namelist["meta"]["simname"]  = case_name # "SOCRATES" # switched to using case_name cause it's what NetCDFIO.jl uses to name the output file 
+    namelist["meta"]["casename"] = case_name # record the specific case name which would be a subtype of SOCRATES supertype (e.g. SOCRATES_RF09_obs) (needs to be case_name for Cases.jl) get_case
```
trontrytel, line 722: Would it be cleaner to write something like
```
function SOCRATES(namelist_defaults; research_flight="RF09", forcing="obs")
    ...
    namelist["meta"]["simname"]  = case_name * "_" * research_flight * "_" * forcing
    namelist["meta"]["casename"] = case_name * "_" * research_flight * "_" * forcing
end
```

trontrytel: This way we would be able to add one SOCRATES case to the CI

driver/dycore.jl
```diff
@@ -335,18 +339,20 @@ function compute_gm_tendencies!(
     C123 = CCG.Covariant123Vector
     C12 = CCG.Contravariant12Vector
     lg = CC.Fields.local_geometry_field(axes(ρ_c))
-    @. tendencies_gm_uₕ -= f × (C12(C123(prog_gm_uₕ)) - C12(C123(aux_gm_uₕ_g)))
+    @. tendencies_gm_uₕ -= f × (C12(C123(prog_gm_uₕ)) - C12(C123(aux_gm_uₕ_g))) # geostrophic matters here...
```
trontrytel, line 342: Is this comment needed?

driver/dycore.jl
```diff
@@ -335,18 +339,20 @@ function compute_gm_tendencies!(
...
 
 
     @inbounds for k in TC.real_center_indices(grid)
         Π = TD.exner(thermo_params, ts_gm[k])
-        # LS Subsidence
+        
+        # LS Subsidence (any positive subsidence (ascent) leads to numerical instability) -- why?)
```
trontrytel, line 348: Do you need to specify positive LS subsidence in any of the cases you are adding?

driver/dycore.jl
```diff
@@ -357,7 +363,7 @@ function compute_gm_tendencies!(
             tendencies_gm.ρθ_liq_ice[k] += ρ_c[k] * aux_gm.dTdt_hadv[k] / Π
         end
         if edmf.moisture_model isa TC.NonEquilibriumMoisture
-            tendencies_gm.q_liq[k] += aux_gm.dqldt[k]
+            tendencies_gm.q_liq[k] += aux_gm.dqldt[k] # we never set these, e.g. in socrates... presumably they're zero?
```
trontrytel, line 366: I think thats fine. The papers don't specify LS forcings for the cloud condensate?

driver/dycore.jl
```diff
@@ -385,6 +391,35 @@ function compute_gm_tendencies!(
             end
         end
 
+        if Cases.force_type(force) <: Cases.ForcingSOCRATES
+            # temperature horizontal divergence
+            tendencies_gm.ρθ_liq_ice[k] += ρ_c[k] * aux_gm.dTdt_hadv[k] / Π # as in all cases but dycoms_rf01 above
```
trontrytel, line 396: This looks like it's applied twice? Once in line 396 and then again in line 406 and 408?

driver/dycore.jl
```diff
@@ -385,6 +391,35 @@ function compute_gm_tendencies!(
...
+
+            # geostrophic handled elsewhere (only affects u,v tendency and maybe surface fluxes?)
```
trontrytel, line 398: Would suggest removing the comment

driver/dycore.jl
```diff
@@ -385,6 +391,35 @@ function compute_gm_tendencies!(
...
+            gm_H_nudge_k =  (aux_gm.H_nudge[k] - aux_gm.θ_liq_ice[k]) / force.scalar_nudge_τᵣ # go back to regular tau relaxation (entire column now unlike zhaoyi's func )
+            tendencies_gm.ρθ_liq_ice[k] += ρ_c[k] * (gm_H_nudge_k + H_fluc) # we have no fluc so get that out (but we have the other terms so brought it back maybe itll lfix intsbailitis?)
+
+            # nudge total water back towards our forcing profile
+            gm_q_tot_nudge_k =  (aux_gm.qt_nudge[k] - aux_gm.q_tot[k]) / force.scalar_nudge_τᵣ
+            tendencies_gm.ρq_tot[k] += ρ_c[k] * (gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k])
+
+            # nudge liquid and ice -- but we don't have a ql_nudge qi_nudge from external forcing in socrates? only qt_nudge -- what are ql_nudge and qi_nudge derived from
+            if edmf.moisture_model isa TC.NonEquilibriumMoisture
+                gm_q_liq_nudge_k =  (aux_gm.ql_nudge[k] - prog_gm.q_liq[k]) / force.scalar_nudge_τᵣ
```
trontrytel, line 416: How are the cloud condensate nudging tendencies defined? Do they come from the paper or some sort of saturation adjustment diagnostics done on the flight?

charleskawczynski:
It looks like the things being done in this branch are pretty complicated. Maybe a short zoom meeting would be best to discuss (@jbphyswx), if you'd like?

driver/dycore.jl
```diff
@@ -273,6 +276,7 @@ function compute_gm_tendencies!(
     surf::TC.SurfaceBase,
     radiation::Cases.RadiationBase,
     force::Cases.ForcingBase,
+    t::Real, # added for time varying forcings...
```
trontrytel, line 279: Where is it used?

driver/common_spaces.jl
```diff
@@ -65,6 +65,19 @@ function construct_mesh(namelist; FT = Float64)
         end
         nz = isnothing(nz) ? Int(zmax ÷ Δz) : Int(nz)
         Δz = isnothing(Δz) ? FT(zmax ÷ nz) : FT(Δz)
+    elseif typeof(Cases.get_case(namelist)) <: Cases.SOCRATES # (we dont have a Cases.SOCRATES instance so improvise) # you can't easily create your own mesh in ClimaCore.Meshes so we'll use some representation of the one they have...
```
trontrytel, line 68: Would it be possible to keep the pattern here and only define `Δz` and `nz` here? The `z_mesh` is then created for all cases in lines 102-108

driver/dycore.jl
```diff
@@ -247,8 +251,7 @@ function ∑tendencies!(tendencies::FV, prog::FV, params::NT, t::Real) where {NT
         TC.compute_precipitation_advection_tendencies(precip_model, edmf, grid, state, param_set)
 
         TC.compute_turbconv_tendencies!(edmf, grid, state, param_set, surf, Δt)
-        compute_gm_tendencies!(edmf, grid, state, surf, radiation, forcing, param_set)
-        nothing
+        compute_gm_tendencies!(edmf, grid, state, surf, radiation, forcing, t, param_set)
```
trontrytel, line 254: I couldn't find where `t` is being used. Also, the return for `∑tendencies!` should be kept `nothing`

driver/dycore.jl
```diff
@@ -236,7 +236,11 @@ function ∑tendencies!(tendencies::FV, prog::FV, params::NT, t::Real) where {NT
         # Update aux / pre-tendencies filters. TODO: combine these into a function that minimizes traversals
         # Some of these methods should probably live in `compute_tendencies`, when written, but we'll
         # treat them as auxiliary variables for now, until we disentangle the tendency computations.
-        Cases.update_forcing(case, grid, state, t, param_set)
+        if isa(case, Cases.SOCRATES)
+            Cases.update_forcing(case, grid, state, t, param_set, forcing) # should dispatch to socrates
```
trontrytel, line 240: Why do we have to pass `forcing` here and not just dispatch on the SOCRATES case? Would making the research flight and forcing type be part of parameters specified via `generate_namelist` be able to solve it?

driver/Forcing.jl
```diff
@@ -33,3 +33,31 @@ function initialize(::ForcingBase{ForcingLES}, grid, state, LESDat::LESData)
         aux_gm.v_nudge[k] = nt.v_nudge[k]
     end
 end
+
+function initialize(forcing::ForcingBase{T}, grid, state, param_set) where {T<:ForcingSOCRATES}# added param_set so we can calculate stuff....
+    FT = eltype(param_set)
```
trontrytel, line 38: `param_set` is not used anywhere. You can get the `FT` type from `aux_gm`

driver/Forcing.jl
```diff
@@ -33,3 +33,31 @@ function initialize(::ForcingBase{ForcingLES}, grid, state, LESDat::LESData)
...
+    aux_gm = TC.center_aux_grid_mean(state)
+
+    ug_keys = (:ug_nudge, :vg_nudge)
+    forcing_funcs = forcing.forcing_funcs[]
+
+    # set the geostrophic velocity profiles -- need to check if we actually have a fcn that can take in t=0 and return a profile... ( i think ours is one func for each z so whoops... maybe need to reconstruct?)
```
trontrytel, line 44: Not sure I understand this comment. Maybe you could create a list of issues and questions you still have about the forcing specification and add it to the PR description?

driver/Cases.jl
```diff
@@ -78,15 +78,54 @@ struct DryBubble <: AbstractCaseType end
...
+        fn = QuoteNode(flight_number)
+        ft = QuoteNode(forcing_type) # we have to do this QuoteNode thing so that particulalry forcing_type is not evaluated to a string before Main.eval can analyze it, so that we can actually create a string, e.g. https://stackoverflow.com/a/70018854 
+        eval( quote Base.@kwdef struct $name <: SOCRATES # kwdef to add methods as defaults, see https://discourse.julialang.org/t/default-value-of-some-fields-in-a-mutable-struct/33408/4, this is the only way I could get it to work https://docs.julialang.org/en/v1/manual/metaprogramming/#man-macros (doesn't need to be Main.eval, just the eval for our case here since not in fcn)
+            # create any methods we may come to need for setting up the case
+            flight_number::Int = $fn
+            forcing_type::Symbol = $ft 
+            # move to ForcingSocrates? 
+            get_forcing = () ->  SOCRATES_Single_Column_Forcings.process_case($fn; obs_or_ERA5=$ft,initial_condition=false)
```
trontrytel, line 102: Would it make the code easier to understand if we did hard code all the forcings that define the SICRATES setup?

driver/generate_namelist.jl
jbphyswx: Perhaps, I sort of  mentioned this as a possibility in my comments in `Cases.jl` around lines 87-90. I mostly wanted that so you could eliminate the subtypes of SOCRATES -- we'd need to update a bunch of calls so the flight number and forcing are passed in the object and we don't necessarily dispatch solely on type.  (i.e get_case(::Val{:SOCRATES}) wouldn't return the necessary information when called in `driver/main.jl` so you'd have to update that to potentially include more information from the namelist or something...

driver/dycore.jl
jbphyswx: They are predominantly negative but occasionally following the equation in Appendix C of Atlas paper lead to slightly positive values -- we can double check if my implementation was correct.

driver/dycore.jl
jbphyswx: 396 is horizontal advection tendency, this one is just regular profile relaxation

driver/dycore.jl
jbphyswx: Hmm perhaps this is outdated -- I think this is all in Cases.jl now so I will remove...

driver/Cases.jl
```diff
@@ -78,15 +78,54 @@ struct DryBubble <: AbstractCaseType end
 
 struct LES_driven_SCM <: AbstractCaseType end
 
+using Pkg
+# Use a local CI registry so that SOCRATES_Single_Column_Forcings can be registered and installed on any machine. That package provides machinery to convert our forcings to any vertical grid/time
+Pkg.Registry.add(Pkg.RegistrySpec(url="https://github.com/jbphyswx/MyRegistry")) # see https://discourse.julialang.org/t/more-informative-project-toml-or-partial-manifest-toml/92804/5
+using SOCRATES_Single_Column_Forcings
+abstract type SOCRATES <: AbstractCaseType end # this is the abstract type for all SOCRATES cases, cause you can't subtype non abstract types in julia...
```
trontrytel, line 85: I'm wondering if it wouldn't be simpler to define one case type SOCRATES and then pass the research flight number and forcing type as the case arguments. You could then based on those case input arguments specify the file name from which you are reading the data and the forcing frequency. 

This could potentially simplify this logic?

driver/common_spaces.jl
jbphyswx: I think I did this because the grid is not uniform but an exact copy of Atlas's -- if you want to just switch to a different vertical grid to keep the same format, we could.

driver/Cases.jl
```diff
@@ -731,7 +783,7 @@ function initialize_forcing(::ARM_SGP, forcing, grid::Grid, state, param_set)
     return nothing
 end
 
-function update_forcing(::ARM_SGP, grid, state, t::Real, param_set)
+function update_forcing(::ARM_SGP, grid, state, t::Real, param_set) #- should these be in Forcing.jl?
```
trontrytel, line 786: I think it's ok to keep it in `Cases`. We want to apply minimal number of changes to the code that would get us to SOCRATES setup

driver/dycore.jl
jbphyswx: That's because we don't have the spline functions stored anywhere so they are passed to update so they can be called... maybe we could store them inside the SOCRATES type rather than just the functions to generate the forcings.

having the flight and forcing type in the namelist could help but you'd also need to update the machinery in `main.jl` etc for example so that `get_case` dispatching on Val can actually initialize the SOCRATES type correctly with that information from the namelist.

trontrytel:
Hi @jbphyswx thank you for opening the PR. I left some initial comments and will keep working on this tomorrow. 

The main question I have right now is how to simplify the case specification, so that we don't have to create the sub-types for different research flights and forcing types? How much of that could be hidden by just providing the path to the dataset that specifies the forcing? Similar as it is done with LES driven SCM cases, where the same case is used to run different simulations by just swapping the path to data file?

driver/Cases.jl
jbphyswx: We'd need get_case(casename::String)  from Cases.jl to dispatch differently when the string is SOCRATES I guess right? I think that would be the easiest solution -- we'd need to change it so it so that rather than dispatching on get_case(::Val{:SOCRATES}) it calls to get_case(::Val{:SOCRATES}, namelist) or something like this so that we can use our own constructor automatically? I dont think that would break anywhere else.

trontrytel:
The Socrates dependency got merged https://github.com/CliMA/TurbulenceConvection.jl/pull/1328 🎉 

Could you merge that into your PR and see how things look @jbphyswx ?

trontrytel:
Would it be possible for this PR to not change anything in the `toml` files? Are there any more dependencies that you need updated @jbphyswx?

driver/Cases.jl
```diff
@@ -149,25 +204,33 @@ include("Radiation.jl")
...
-get_case(::Val{:Rico}) = Rico()
-get_case(::Val{:TRMM_LBA}) = TRMM_LBA()
-get_case(::Val{:ARM_SGP}) = ARM_SGP()
-get_case(::Val{:GATE_III}) = GATE_III()
-get_case(::Val{:DYCOMS_RF01}) = DYCOMS_RF01()
-get_case(::Val{:DYCOMS_RF02}) = DYCOMS_RF02()
-get_case(::Val{:GABLS}) = GABLS()
-get_case(::Val{:DryBubble}) = DryBubble()
-get_case(::Val{:LES_driven_SCM}) = LES_driven_SCM()
+get_case(namelist::Dict, param_set::APS)  = get_case(namelist["meta"]["casename"], namelist, param_set)
```
costachris, line 210: I don't think you're using `param_set` anymore (right?), so these function definitions for all the other cases can be reverted (and a new one added for SOCRATES).

driver/Cases.jl
```diff
@@ -1237,4 +1303,110 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+    IC = case.get_initial_conditions(;new_z = vec(grid.zc.z)) # added S so it has a name and then use the method we defined, currently param_set isn't an accepted argument cause it shouldnt be necessary for this part on earth but maybe can edit later...
+    
+    prog_gm_u = copy(aux_gm.q_tot) # copy as template cause u,g go into a uₕ vector so this is easier to work with (copied from other cases...)
+    prog_gm_v = copy(aux_gm.q_tot)
+    @inbounds for k in real_center_indices(grid) # we are goin the TRMM way wit this rather than LES... not sure why they use k.i but maybe it's cause forcing has time component? idk
+        aux_gm.θ_liq_ice[k] = IC.H_nudge[k]
+        aux_gm.q_tot[k]     = IC.qt_nudge[k]
+        prog_gm_u[k]        = IC.u_nudge[k]
+        prog_gm_v[k]        = IC.v_nudge[k]
+        aux_gm.tke[k]       = 0 # what is this supposed to be? unset for interactive? maybe we dont need to set it initially at all idk... didn't check yet but these get updated later interactively so hope it's fine
```
costachris, line 1335: You may have to experiment with this. If results don't match what you expect try initializing with a TKE profile like LES_driven_SCM

driver/dycore.jl
```diff
@@ -385,6 +391,35 @@ function compute_gm_tendencies!(
...
+        if Cases.force_type(force) <: Cases.ForcingSOCRATES
+            # temperature horizontal divergence
+            tendencies_gm.ρθ_liq_ice[k] += ρ_c[k] * aux_gm.dTdt_hadv[k] / Π # as in all cases but dycoms_rf01 above
+
+            # geostrophic handled elsewhere (only affects u,v tendency and maybe surface fluxes?)
+
+            # nudge back to era 5 horizosntal
+            gm_U_nudge_k = (aux_gm.u_nudge[k] - prog_gm_u[k]) / force.wind_nudge_τᵣ 
+            gm_V_nudge_k = (aux_gm.v_nudge[k] - prog_gm_v[k]) / force.wind_nudge_τᵣ
+            tendencies_gm_uₕ[k] += CCG.Covariant12Vector(CCG.UVVector(gm_U_nudge_k, gm_V_nudge_k), lg[k]) # add nudge (again we have dqtdt_hadv so what re dqldt and dqidt derived from?)
```
costachris, line 403: Some of these operations overlap with the LES forcings ([here](https://github.com/CliMA/TurbulenceConvection.jl/blob/9274195055be873f19e9540c6a7a0e777842311c/driver/dycore.jl#L366)). In that if statement you could do something like `Cases.force_type(force) <: Union{Cases.ForcingLES, Cases.ForcingSOCRATES}` there to apply overlapping operations, and do the rest here. Although our `Γᵣ` is different. So you'd have to add a `compute_SOCRATES_Γᵣ` or something similar

driver/Cases.jl
jbphyswx: the one i have does use param_set for socrates -- the constructor uses it to fill out the fields around line 108. The alternative is to keep that call the same and use like a SOCRATESDat file -- but i think then we'd have to pass some more information to surface_ref_state, and add a dispatch for that separately for socrates. There are some other small complications with each so it would be up to yall to say which is better.

costachris:
This is somehow changing the behavior of non-socrates cases in the CI. See ie. life_cycle_tan2018 in the buildkite tests.

integration_tests/Project.toml
```diff
@@ -74,6 +74,7 @@ StaticArrays = "1.2"
 StatsBase = "0.33"
 StochasticDiffEq = "6.41"
 SurfaceFluxes = "0.4"
+SOCRATESSingleColumnForcings = "0.6.0"
```
charleskawczynski, line 77:
```suggestion
SOCRATESSingleColumnForcings = "0.6"
```

I've updated the dependencies in #1333, so we can discard the manifest file changes.

driver/Cases.jl
```diff
@@ -1237,4 +1341,114 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+    FT = TC.float_type(state)
+
+    forcing_funcs = forcing.forcing_funcs[] # access our functions
+    ug_keys = (:ug_nudge, :vg_nudge)
+    # update geostrophic profile
+    g_func = (f) -> f([FT(t)])[1]
+    prof_ug = g_func.(forcing_funcs[:ug_nudge]) # map over the forcing funcs to get the profile at t=0
+    prof_vg = g_func.(forcing_funcs[:vg_nudge])
+    # it wants a fcn out, could edit src/Fields.jl I guess to add another method but maybe it needs to face/center points idk...
+    prof_ug = Dierckx.Spline1D(vec(grid.zc.z), vec(prof_ug); k = 1)
```
costachris, line 1425: I think you can just use the `pyinterp` function from `src/utility_functions.jl`

driver/Cases.jl
```diff
@@ -1237,4 +1341,114 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+
+
+    aux_gm_uₕ_g = TC.grid_mean_uₕ_g(state)
+    TC.set_z!(aux_gm_uₕ_g, prof_ug, prof_vg)
+
+    forcing_funcs = forcing_funcs[setdiff(keys(forcing_funcs), ug_keys)] # remove keys we don't need
+    @inbounds for k in real_center_indices(grid)
+        for (name, funcs) in zip(keys(forcing_funcs), forcing_funcs)
+            func = funcs[k]
+            getproperty(aux_gm, name)[k] = func([t])[1] # turn to vec cause needs to be cast as in https://github.com/CliMA/TurbulenceConvection.jl/blob/a9ebce1f5f15f049fc3719a013ddbc4a9662943a/src/utility_functions.jl#L48, run fcn on vec and index it back outssss
```
costachris, line 1436: Does `getproperty` actually modify objects in place?

driver/Cases.jl
```diff
@@ -1237,4 +1341,114 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+        scalar_nudge_τᵣ = get(namelist["forcing"], "scalar_nudge_τᵣ", 20 * 60) # paper standard
+        (; wind_nudge_τᵣ = wind_nudge_τᵣ, scalar_nudge_τᵣ = scalar_nudge_τᵣ)
+    else # SOCRATES_ERA5
+        wind_nudge_τᵣ = get(namelist["forcing"], "wind_nudge_τᵣ", 60 * 60) # paper standard
+        scalar_nudge_τᵣ = get(namelist["forcing"], "scalar_nudge_τᵣ", (6 * 60) * 60) # paper standard (testing 6 hr relaxation on T/q since we don't have RRTMG to be in line more w/ appendix D)
+        (; wind_nudge_τᵣ = wind_nudge_τᵣ, scalar_nudge_τᵣ = scalar_nudge_τᵣ)
+    end
+end
+
+function update_forcing(case::SOCRATES, grid, state, t::Real, param_set, forcing) # Adapted from ARM_SGP -- should these be in Forcing.jl -- called in dycore.jl
```
costachris, line 1414: Is the idea here that the geostrophic forcing is changing in time?

costachris (changes requested):

driver/Cases.jl
jbphyswx: `src/utility_functions.jl` function evaluates the spline, here we just want to get the spline function itself out (see the comment)

driver/Cases.jl
jbphyswx: yes, because you're just editing the array value aux_gm[name][k]

driver/Cases.jl
```diff
@@ -149,25 +241,34 @@ include("Radiation.jl")
...
+# get_case(::Val{:Rico}) = Rico()
+# get_case(::Val{:TRMM_LBA}) = TRMM_LBA()
+# get_case(::Val{:ARM_SGP}) = ARM_SGP()
+# get_case(::Val{:GATE_III}) = GATE_III()
+# get_case(::Val{:DYCOMS_RF01}) = DYCOMS_RF01()
+# get_case(::Val{:DYCOMS_RF02}) = DYCOMS_RF02()
+# get_case(::Val{:GABLS}) = GABLS()
+# get_case(::Val{:DryBubble}) = DryBubble()
+# get_case(::Val{:LES_driven_SCM}) = LES_driven_SCM()
 get_case_name(case_type::AbstractCaseType) = string(case_type)
```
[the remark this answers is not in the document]
jbphyswx: yes this may have been deprecated in favor of the get_case framework proposed above...

driver/Cases.jl
jbphyswx: yes the forcings change in time

driver/main.jl
```diff
@@ -65,7 +65,7 @@ struct Simulation1d{IONT, P, A, C, F, R, SM, EDMF, PM, RFM, D, TIMESTEPPING, STA
     dt_min::Float64
     truncate_stack_trace::Bool
     surf_ref_state::SRS
-    les_data_kwarg::LESDK
+    aux_data_kwarg::AUXDK
```
charleskawczynski, line 68: This name change would be a good and simple separate PR. Can we break this PR down a bit, and start with this one?

charleskawczynski (commented):
I left a couple comments

charleskawczynski (commented):
I left a couple comments

charleskawczynski (commented):
I left a couple comments

driver/Cases.jl
```diff
@@ -1237,4 +1341,114 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+    IC = case.get_initial_conditions(; new_z = vec(grid.zc.z)) # added S so it has a name and then use the method we defined, currently param_set isn't an accepted argument cause it shouldnt be necessary for this part on earth but maybe can edit later...
+
+    prog_gm_u = copy(aux_gm.q_tot) # copy as template cause u,g go into a uₕ vector so this is easier to work with (copied from other cases...)
+    prog_gm_v = copy(aux_gm.q_tot)
+    @inbounds for k in real_center_indices(grid) # we are goin the TRMM way wit this rather than LES... not sure why they use k.i but maybe it's cause forcing has time component? idk
+        aux_gm.θ_liq_ice[k] = IC.H_nudge[k]
+        aux_gm.q_tot[k] = IC.qt_nudge[k]
+        prog_gm_u[k] = IC.u_nudge[k]
+        prog_gm_v[k] = IC.v_nudge[k]
+        aux_gm.tke[k] = 0 # what is this supposed to be? unset for interactive? maybe we dont need to set it initially at all idk... didn't check yet but these get updated later interactively so hope it's fine
```
jbphyswx, line 1373: something non-0, see what is done for Bomex

driver/Cases.jl
```diff
@@ -1237,4 +1341,114 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+
+
+
+function surface_params(case::SOCRATES, surf_ref_state, param_set; SOCRATESDat, kwargs...) # seems Ri_bulk_crit is never aactually changed...
+    FT = eltype(param_set)
+    sc = case.get_surface()
+    zrough = 0.1 # copied from gabls which is also w/ monin obhukov boundary layer
+    # no ustar w/ monin obukhov i guess, seems to be calculated in https://github.com/CliMA/SurfaceFluxes.jl src/SurfaceFluxes.jl/compute_ustar()
+    kwargs = (; Tsurface = sc.Tg, qsurface = sc.qg, zrough, kwargs...) # taken from gabls cause only other one w/ moninobhukov interactive,
+    return TC.MoninObukhovSurface(FT; kwargs...) # interactive?
```
jbphyswx, line 1388: clarify w/ tapio if this is what we want

jbphyswx: plot' lhf/shf from otput to veritfy

driver/Cases.jl
jbphyswx: Maybe also using the 6 hour relaxation helps with this.

driver/generate_namelist.jl
jbphyswx: we just use `SOCRATES` as the casename so dispatch is easier, but we added other machinery in Cases.jl (including having the flight_number and forcing in the socrates object type), but the case name for the call to TC.jl is left as SOCRATES_RF{flight_number}_{forcing_type}_data

driver/dycore.jl
jbphyswx: we fixed this by not allowing large scale ascent in SOCRATESSingleColumnForcings.jl, can move that back here if necessary

driver/dycore.jl
jbphyswx: this was reorganized

driver/dycore.jl
jbphyswx: can discuss, was left for clarity with the paper text

driver/dycore.jl
jbphyswx: there is no such definition in the paper. I just nudge qt and let things work out from there. I am not sure how the partition would work say if qt is going down with respect to partition between `ql`, `qi`, `qv` but I suspect it's just from `qv` since we didn't add any `ql`, `qi` tendency. I guess that could cause problems if `dqtdt_hadv` was large enough, but the majority of water should be `qv`, so I don't suspect large issues.

driver/dycore.jl
jbphyswx: **removed, there are many other unused things in `driver/dycore.jl | compute_gm_tendencies!()` function though that already existed, idk if that's bad

driver/dycore.jl
jbphyswx: removed the `t`, I don't think that nothing does anything there -- `∑tendencies!` has a return nothing statement at the end regardless.

driver/dycore.jl
jbphyswx: we did move the flight number and forcing type to the namelist, but still the forcing functions need to be passed in so they can be evaluated at time t, and they are not stored in the socrates object as the machinery for passing forcings around is already part of TC.jl's infrastructure. For most other forcings (e.g. LES) these are not functions of time so the socrates `update_forcing` takes the forcing functions in addition to the time in order to evaluate them. For ARM_SGP, these were external functions in APL, which we don't have (partially because we dynamically generate the forcing depending on the vertical grid that is being called.)

driver/Forcing.jl
jbphyswx: these have been updated many times, this is very out of date -- i believe all the forcing specific questions have been resolved one way or another though

driver/Cases.jl
jbphyswx: this is outdated, now we don't dynamically generate the types. we do not hardcode the forcings either but it should be clearer

driver/Cases.jl
jbphyswx: we do do this now, we just needed a custom SOCRATES type infrastructure.

driver/main.jl
```diff
@@ -206,16 +206,16 @@ function Simulation1d(namelist)
         JSON.print(io, namelist, 4)
     end
 
-    case = Cases.get_case(namelist)
-    surf_ref_state = Cases.surface_ref_state(case, param_set, namelist)
+    case = Cases.get_case(namelist, param_set)
```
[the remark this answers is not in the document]
jbphyswx: no, because we didnt want to dynamically create types, we now need the `param_set` to generate the forcings from the input files in `driver/Cases.jl`. see the calls around lines 247 to 257 that over load the old functionality -- should have no effect on prior syntax and setup.

driver/main.jl
```diff
@@ -206,16 +206,16 @@ function Simulation1d(namelist)
...
+    surf_ref_state = Cases.surface_ref_state(case, param_set, namelist) # this doesn't accept aux_data_kwarg which would be useful for socrates... didn't want to change the call for all cases though...
 
     forcing = Cases.ForcingBase(case, FT; Cases.forcing_kwargs(case, namelist)...)
 
     radiation = Cases.RadiationBase(case, FT)
     TS = TimeStepping(FT, namelist)
 
     Ri_bulk_crit::FTD = namelist["turbulence"]["EDMF_PrognosticTKE"]["Ri_crit"]
-    aux_data_kwarg = Cases.aux_data_kwarg(case, namelist)
+    aux_data_kwarg = Cases.aux_data_kwarg(case, namelist, param_set) # change name and pass in all info we may need for this (added param_set) -- I kinda wanted this for surf ref state though...
```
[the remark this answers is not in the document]
jbphyswx: we used to use it to get socratesdata but now we generate the forcing_funcs separately and don't carry the functions around in socratesdat... so removal should work.

Now get_case has the param_set for socrates

driver/dycore.jl
jbphyswx: it was a refernence to how it's hard to tell where the geostrophic wind forcing matters relative to the actual wind nudging tendency and where and how each should be applied since they both operate on winds... it's not clear how they interplay in general so i made a note of that. I've removed the note now

driver/dycore.jl
jbphyswx: we just operate on qt, no partitioning

driver/dycore.jl
jbphyswx: left separate for easier changing/testing and legibility

driver/dycore.jl
```diff
@@ -385,6 +391,35 @@ function compute_gm_tendencies!(
...
+
+            # nudge back to era 5 horizosntal
+            gm_U_nudge_k = (aux_gm.u_nudge[k] - prog_gm_u[k]) / force.wind_nudge_τᵣ
+            gm_V_nudge_k = (aux_gm.v_nudge[k] - prog_gm_v[k]) / force.wind_nudge_τᵣ
+            tendencies_gm_uₕ[k] += CCG.Covariant12Vector(CCG.UVVector(gm_U_nudge_k, gm_V_nudge_k), lg[k]) # add nudge (again we have dqtdt_hadv so what re dqldt and dqidt derived from?)
+
+            # nudge liq-ice pottemp back towards our forcing profile
+            H_fluc = aux_gm.dTdt_fluc[k] / Π # should be 0
+            gm_H_nudge_k = (aux_gm.H_nudge[k] - aux_gm.θ_liq_ice[k]) / force.scalar_nudge_τᵣ # go back to regular tau relaxation (entire column now unlike zhaoyi's func )
+            tendencies_gm.ρθ_liq_ice[k] += ρ_c[k] * (gm_H_nudge_k + H_fluc) # we have no fluc so get that out (but we have the other terms so brought it back maybe itll lfix intsbailitis?)
```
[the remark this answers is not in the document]
jbphyswx: just to be sure if I dont change this it stays 0? 
I didn't know if I should worry about not including it if say the model edits it itself...

driver/dycore.jl
```diff
@@ -385,6 +391,35 @@ function compute_gm_tendencies!(
...
+
+            # nudge total water back towards our forcing profile
+            gm_q_tot_nudge_k = (aux_gm.qt_nudge[k] - aux_gm.q_tot[k]) / force.scalar_nudge_τᵣ
+            tendencies_gm.ρq_tot[k] += ρ_c[k] * (gm_q_tot_nudge_k + aux_gm.dqtdt_fluc[k])
```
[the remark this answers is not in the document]
jbphyswx: same question here as for `dTdt_fluc`

driver/Forcing.jl
jbphyswx: updated

integration_tests/LocalPreferences.toml
```diff
@@ -0,0 +1,6 @@
+[MPIPreferences]
```
costachris, line 1: Remove changes to `integration_tests/LocalPreferences.toml`

.gitignore
```diff
@@ -13,6 +13,7 @@
 *.json
 *.in
 *.gz
+*.out
```
costachris, line 16: Remove changes to `.gitignore`. ie. We should be checking in the integration_tests/Manifest.toml

integration_tests/LocalPreferences.toml
jbphyswx: wait I thought there was a bug and the model doesn't run localpreferences.toml nowadays for mpi, I added that cause of a thread on the slack and some previous discussion

driver/Cases.jl
costachris: I'd suggest initializing gm TKE the same way as `LES_driven_SCM` (a simple profile that decays with height)

driver/Cases.jl
trontrytel: Its a good Tapio question. I'm guessing maybe larger than 0.1 but less than 0.4?

driver/Cases.jl
```diff
@@ -1237,4 +1332,99 @@ initialize_forcing(::LES_driven_SCM, forcing, grid::Grid, state, param_set; LESD
...
+# function initialize(self::RadiationBase{SOCRATES_RF09_obs}, grid, state)
+#     return nothing
+# end
+
+# function update_radiation(self::RadiationBase{SOCRATES_RF09_obs}, grid, state, t::Real, param_set)
+#     return nothing
+# end
+
+# currently still nothing just cause idk what to do w/ RRTMG or if it's long enough to need...
+RadiationBase(case::SOCRATES, FT) = RadiationBase{Cases.get_radiation_type(case), FT}() # i think this should default to none, would deprecate this call for now cause we dont have a use, default is just none... but aux_data_kwarg is in the end of the main.jl initialize_radiation.jl cal so we gotta improvise
```
trontrytel, line 1427: I guess if we are using 20 min relaxation timescales it doesn't matter that we don't have radiation

costachris (approved):

PR #1330, opened 2023-03-12
remove surface area boundary condition

costachris:
- Allow surface area to be prognostically determined in the bottom cell center, instead of prescribed as `surface_area` parameter. In initial conditions, area fraction and updraft w set to nonzero values in bottom cell center and top face, respectively, along with the associated prognostic variables for these variables. This initialization is to break symmetry and prevent trivial a = 0, w = 0 solution. 
- At bottom cell face, `a_up` = 0 & `a_en` = 1. At bottom cell center and above -> `a` prognostically determined. Note: `a_en`  at surface bottom face is needed to compute diffusive fluxes.
- Swap order of `set_edmf_surface_bc` and `filter_updraft_vars`, since updraft variables are used to compute surface `theta` and `qt` fluxes.

### DIFF
--- a/driver/compute_diagnostics.jl
+++ b/driver/compute_diagnostics.jl
@@ -303,7 +303,7 @@ function compute_diagnostics!(
         end
     end
 
-    a_bulk_bcs = TC.a_bulk_boundary_conditions(surf, edmf)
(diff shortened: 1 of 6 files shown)


costachris:
Superseded by #1332

PR #1336, opened 2023-05-03
Try fixing some of GHA CI runs

trontrytel:
This fixes GHA for Linux and OSX. 

On windows I get `ReadOnlyMemoryError()` @charleskawczynski have you seen this before somewhere among our repos?


charleskawczynski (approved):
I think the windows errors have been there a while (maybe since we started inlining lots of stuff in ClimaCore). I'm a bit surprised it's not resolved by now, but I think the windows machines may also have lower memory limits than the other OSs. I think we can look at the last time the bors toml was changed to see when we removed the windows build.

PR #1339, opened 2023-05-25
update SSCF to 0.9

jbphyswx:
## Purpose 
Add dependency to force version 0.9 for SOCRATESSingleColumnForcings

----
- [ ] I have read and checked the items on the review checklist.


charleskawczynski:
I think you need to instantiate so that you incorporate manifest changes too

charleskawczynski (commented):
I forget the environments in TC-- do we need to add this dep to more environments? cc @trontrytel

trontrytel:
> I forget the environments in TC-- do we need to add this dep to more environments? cc @trontrytel

No, I think it's just the integration tests. I am re-trying the Ci now. I think buildkite failed randomly before, not because of this PR.

Btw, can you remind me what the UpDeps test is checking? I wonder why it is failing now, as we have just updated the deps?

charleskawczynski:
> Btw, can you remind me what the UpDeps test is checking? I wonder why it is failing now, as we have just updated the deps?

https://github.com/CliMA/TurbulenceConvection.jl/pull/909#pullrequestreview-930773269

trontrytel (approved):

PR #1344, opened 2023-08-09
Use flux boundary condition for tke

costachris:
Replace Dirichlet boundary condition for tke with flux boundary condition.


src/EDMF_functions.jl
```diff
@@ -962,6 +964,20 @@ function compute_en_tendencies!(
...
+    # compute bottom BC for TKE
+    if is_tke
+        ρ_f_surf = ρ_f[kf_surf]
+        a_e_surf = 1 - edmf.surface_area
+        u_surf = physical_grid_mean_u(state)[kc_surf]
+        v_surf = physical_grid_mean_v(state)[kc_surf]
+        U_surf_norm = sqrt(u_surf^2 + v_surf^2)
+        ustar = surf.ustar
+        surface_tke_turb_flux = get_surface_tke_turb_flux(mix_len_params, ustar, ρ_f_surf, a_e_surf, U_surf_norm)
+        ∇c_turb = CCO.DivergenceF2C(; bottom = CCO.SetValue(CCG.Covariant3Vector(surface_tke_turb_flux)))
```
szy21, line 976: I think it should be a `CCG.WVector` here based on your function that calculate `surface_tke_turb_flux`, but then I am surprised it works...

src/turbulence_functions.jl
```diff
@@ -58,6 +58,14 @@ function get_surface_tke(mixing_length_params, ustar::FT, zLL::FT, oblength::FT)
         return κ_star² * ustar * ustar
     end
 end
+
+function get_surface_tke_turb_flux(mixing_length_params, ustar::FT, ρ_f_surf::FT, a_e_surf::FT, U_surf_norm::FT) where {FT}
+    κ_star² = mixing_length_params.κ_star²
```
szy21, line 63: Is κ_star² von karman constant squared?

src/EDMF_functions.jl
costachris: that was based on analogous operations in ClimaAtmos with DivergenceF2C, but I'll use WVector instead

src/turbulence_functions.jl
costachris: `κ_star²` is ratio of TKE to squared friction velocity. (`κ_star` is ratio of rms turbulent velocity to friction velocity)

szy21: Ah, I might do it wrong in CilmaAtmos then. Is that parameter in https://github.com/CliMA/CLIMAParameters.jl/blob/main/src/parameters.toml?

szy21 (approved):

PR #1351, opened 2023-09-28
Remove MD Entrainment hard coded settings

costachris:
Remove and clean up hard-coded dimensional scales in MD entrainment closure.


src/closures/entr_detr.jl
```diff
@@ -15,11 +15,12 @@ function compute_inverse_timescale(εδ_model, b_up::FT, b_en::FT, w_up::FT, w_e
     Δb = b_up - b_en
     Δw = get_Δw(εδ_model, w_up, w_en)
     c_λ = εδ_params(εδ_model).c_λ
+    smin_rm = εδ_params(εδ_model).smin_rm
 
-    l_1 = c_λ * abs(Δb / sqrt(tke + 1e-8))
+    l_1 = c_λ * abs(Δb / (sqrt(tke) + eps(FT)))
```
trontrytel, line 20: Can tke be zero or negative? If yes, this will break

trontrytel (approved):

src/closures/entr_detr.jl
costachris: It can be 0 (which should he handled by the eps), but not < 0 AFAIK. But I'll add an `abs(tke)` just in case.

PR #1358, opened 2023-10-19
Explicitly declare InterpolateC2F kwargs

costachris:
Explicitly declare InterpolateC2F kwargs for bottom and top boundaries on area.

### DIFF
--- a/driver/compute_diagnostics.jl
+++ b/driver/compute_diagnostics.jl
@@ -303,16 +303,14 @@ function compute_diagnostics!(
         end
     end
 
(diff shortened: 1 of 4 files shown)


szy21 (approved):
You can also just define a `const` operator with the correct boundary conditions and use that, like what we did [here](https://github.com/CliMA/ClimaAtmos.jl/blob/76be7f0c37ac17d3d8229e47c29c6666af4e2ca9/src/utils/abbreviations.jl#L33) in ClimaAtmos.

PR #1363, opened 2023-10-20
Use exponential detrainment limiter and apply to all entr function types

costachris:
Use exponential detrainment limiter instead of the currently-used logistic function and apply to all entr function types. 

## Original 

$$ \left(2 - \frac{1}{{1 + \exp(-\gamma_{lim} \times (a_{max} - a\_{up}))}}\right)^{\beta\_{lim}} - 1$$

## Exponential 

$$ \gamma_{lim}  \exp(-\beta\_{lim} \times (a_{max} - a\_{up}))$$ 

A comparison of the two formulations is provided below (one on a log scale)

combined_max_area_limiter_plot
combined_max_area_limiter_logistic_plot_logscale


szy21 (approved):

szy21:
Oh, a minor thing. If you make this change the description in CLIMAParameters is not accurate anymore: https://github.com/CliMA/CLIMAParameters.jl/blob/945eb72ea84cf5ad8231bcdd7715e8a57dd11e45/src/parameters.toml#L2250-L2260. Not sure if we should change it too?

PR #1389, opened 2024-02-28
Use quadrature to compute environmental ql, qi instead of mean

costachris:
- [x] Use quadratures to compute environmental ql, qi . Previously this was causing instances of nonzero env cloud fraction with zero ql. 

- [x] Remove shrink_mask and use placeholder values for environment sat when no env cloud fraction. This is to address spurious large gradients when there's noisy switching between cloudy/noncloudy env air.


src/EDMF_Environment.jl
```diff
@@ -53,11 +53,18 @@ function microphysics(
...
         else
             aux_en.cloud_fraction[k] = 0
             aux_en_unsat.θ_dry[k] = TD.dry_pottemp(thermo_params, ts)
             aux_en_unsat.θ_virt[k] = TD.virtual_pottemp(thermo_params, ts)
             aux_en_unsat.q_tot[k] = TD.total_specific_humidity(thermo_params, ts)
+            aux_en_sat.T[k] = aux_en.T[k]
+            aux_en_sat.q_vap[k] = 0
+            aux_en_sat.q_tot[k] = aux_en.q_tot[k]
+            aux_en_sat.θ_dry[k] = aux_en.θ_dry[k]
+            aux_en_sat.θ_liq_ice[k] = aux_en.θ_liq_ice[k]
```
trontrytel, line 67: Should we set `aux_en.q_liq[k]` and `aux_en.q_ice[k]` to zero in the non-sat case?

src/EDMF_Environment.jl
```diff
@@ -313,8 +321,28 @@ function microphysics(
...
+                aux_en.cloud_fraction[k] = outer_env.cf
+                aux_en.q_liq[k] = outer_env.ql
+                aux_en.q_ice[k] = outer_env.qi
+
+            else
+                aux_en.cloud_fraction[k] = 0.0
+                aux_en.q_liq[k] = 0.0
+                aux_en.q_ice[k] = 0.0
+
+                aux_en.qt_tendency_precip_formation[k] = 0.0
```
trontrytel, line 335: It would probably be better to do those inside the `precipitation_formation` so that we can use the same `has_condensate` threshold when the function is used on the grid-mean and quadrature input values

src/EDMF_Environment.jl
```diff
@@ -377,11 +407,19 @@ function microphysics(
                 aux_en_sat.T[k] = TD.air_temperature(thermo_params, ts)
                 aux_en_sat.q_tot[k] = TD.total_specific_humidity(thermo_params, ts)
                 aux_en_sat.q_vap[k] = TD.vapor_specific_humidity(thermo_params, ts)
+
```
trontrytel, line 410: Should we also be setting the `unsat` values in the `sat` case? Since we are no longer relying on the mask?

trontrytel (approved):

src/update_aux.jl
```diff
@@ -336,20 +334,11 @@ function update_aux!(edmf::EDMFModel, grid::Grid, state::State, surf::SurfaceBas
     @. ∂θv∂z = ∇c(wvec(If0(θ_virt_en)))
 
     # Second order approximation: Use dry and cloudy environmental fields.
-    cf = aux_en.cloud_fraction
-    shm = copy(cf)
-    pshm = parent(shm)
-    shrink_mask!(pshm, vec(cf))
```
charleskawczynski, line 342: Just a heads up, I think that this is the only place where `shrink_mask!` is used. If it's no longer going to be used, we could delete the function definition, too

src/EDMF_Environment.jl
costachris: Just the precip_formation tendencies: 

```
aux_en.qt_tendency_precip_formation[k] = qt_tendency * aux_en.area[k]
aux_en.θ_liq_ice_tendency_precip_formation[k] = θ_liq_ice_tendency * aux_en.area[k]
```
or the q_rai/q_sno tendencies as well?
```
tendencies_pr.q_rai[k] += qr_tendency * aux_en.area[k]
tendencies_pr.q_sno[k] += qs_tendency * aux_en.area[k]
```

costachris: I think if I set the aux_en precip formation tendencies in `precipitation_formation ` that will result in the tendencies being set twice since  `precipitation_formation` is called both in `microphysics` to deal with the environment and later in `compute_precipitation_formation_tendencies` to deal with the updraft.

trontrytel: How about:
- adding the same `has_condensate` threshold inside `precipitation_formation` around this line https://github.com/CliMA/TurbulenceConvection.jl/blob/9d18d79ada3dcc863f575bf19ffec4daa6c8f446/src/microphysics_coupling.jl#L74. This way you would be returning zero tendencies for when you have virtually no cloud.

- And then you could assign values to the `aux_en.qt_tendency_precip_formation[k]` and others without any if/esle in the EDMFEnvironment and Updrafts?

costachris: To be addressed in later release, tracked by #1396

